PR #141, opened 2026-07-31
fix(daemon): stop the action on timeout, and never roll back over a live one

### LINKED ISSUE #140: Action timeout kills sudo, not the work: orphaned privileged process races an automatic rollback
## Summary

The action timeout stops the daemon **waiting**; it does not stop the **action**. For rollback-eligible actions this then triggers `rpm-ostree rollback` while the original privileged transaction is still running.

## The chain

1. The daemon runs unprivileged (`User=sysknife` in `packaging/sysknife-daemon.service`) and elevates per action with `sudo`. `UpdateSystem` is `sudo rpm-ostree upgrade` (`crates/sysknife-daemon/src/actions/deployment.rs:65`).
2. On timeout, `kill_and_reap` calls `child.start_kill()` (`crates/sysknife-daemon/src/executor.rs:195`), which SIGKILLs only the direct child. That child is `sudo`, which has already forked the real command. There is no `process_group`, `setsid`, or cgroup anywhere in the daemon (`grep -rn "process_group\|setsid\|killpg"` returns nothing).
3. The executor returns `ExecutorError::Io(TimedOut)` and the job becomes `JobState::Failed`.
4. `attempt_rollback_if_needed` (`crates/sysknife-daemon/src/dispatcher.rs:2592`) fires on `Failed` when `rollback_available`, and runs the rollback spec immediately, in the same job.

Net effect: `sudo rpm-ostree rollback` runs against a host where `rpm-ostree upgrade` is still in flight. Best case it fails on a lock and the job reports `RollbackFailed`. Worst case the two interleave and the host lands in a state neither the operator nor the job record describes.

`rpm-ostree` makes it starker: it is a thin client to `rpm-ostreed`, so even killing it directly would not stop the transaction.

## Evidence

Killing the direct child does not stop the work. Matched by exact argv read from `/proc` rather than `pgrep -f`, which also matches the probing shell:

```
--- single command  (what the daemon's timeout test exercises)
    spawned pid=3475918  argv='sh -c sleep 31410'
    RESULT: work SURVIVED -> pid(s) [3475919] still running after the kill
--- two commands, && (snap install, firewall reload)
    RESULT: work SURVIVED -> pid(s) [3475948] still running after the kill
--- two commands, ;  (add/remove user to group)
    RESULT: work SURVIVED -> pid(s) [3475989] still running after the kill
```

## Why the existing test does not catch it

`execute_spec_kills_a_child_that_outruns_the_action_timeout` (`crates/sysknife-daemon/tests/execute_spec.rs:392`) asserts two things: that `execute_spec` returns within 60s, and that the error message contains "timeout". Both are true of a daemon that gives up waiting while the work continues. The test never asserts that the spawned work stopped, so it passes for the wrong reason.

## Affected actions

Every action is orphaned on timeout, because privileged actions all go through `sudo`. The dangerous subset is the rollback-eligible list in `executor::rollback_spec_for`: `UpdateSystem`, `InstallPackages`, `RemovePackages`, `RebaseSystem`, `SetKernelArguments`, `AddLayeredPackage`, `RemoveLayeredPackage`, `ReplaceLayeredPackage`. Those are also the longest-running ones, so they are the most likely to hit the timeout at all.

## Proposed fix

1. Spawn each action in its own process group: `std::os::unix::process::CommandExt::process_group(0)` on the `Command`.
2. On timeout signal the **group**, not the pid: `killpg(-pgid, SIGTERM)`, a short grace period, then `SIGKILL`, then reap. `sudo` propagates signals to its child, and a group kill covers shell-spawned grandchildren.
3. Only treat the action as stopped once the group is gone; if it cannot be stopped, that must be a distinct outcome from "timed out and stopped", because the automatic-rollback decision hangs on it.
4. Do not attempt rollback when the original action could not be confirmed stopped. Racing two privileged package operations is worse than leaving one hung.

`rpm-ostree` still needs care: cancelling its transaction is `rpm-ostree cancel`, not a signal to the client.

## Test that must accompany it

Rewrite the timeout test so it fails today: spawn a marker process from a multi-command script, let it time out, then assert the marker is gone by scanning `/proc` for exact argv. Asserting on the error message alone is what let this through.

## Provenance

Surfaced while answering a taxonomy question in the MCP registry discussions ([#1416](https://github.com/modelcontextprotocol/registry/discussions/1416)) about whether shell expansion, wrappers and delegated subprocesses should be audited as one category or three. Checking our own answer against the code turned this up: expansion and wrappers are well defended here (constant scripts, positional `$1` args in `actions/ssh.rs`, charset allowlists via `validated_safe_arg`), and delegated subprocesses are not.

vladimirrott:
## Summary

Closes #140.

The daemon runs unprivileged (`User=sysknife`) and elevates per action through `sudo`, which forks. The action deadline called `child.start_kill()`, SIGKILLing only the process the daemon spawned, so the privileged work carried on while the daemon logged `killed` and returned a timeout.

The consequence is not a leaked process. A timeout marks the job `Failed`, and `attempt_rollback_if_needed` immediately runs `rpm-ostree rollback` for rollback-eligible actions, so a host could be running an upgrade and a rollback of that upgrade simultaneously, the second started precisely because the daemon believed the first had stopped.

## What changed

- Actions spawn with `process_group(0)`, so the deadline can address the whole tree.
- The deadline signals the **group**: SIGTERM, a five second grace, then SIGKILL. The direct child is reaped first, because a zombie keeps the group alive and would make the liveness check lie.
- If the group still has members after SIGKILL, the daemon returns the new `ExecutorError::ActionNotStopped { program, pgid, timeout_secs }` instead of a plain timeout.
- The dispatcher vetoes automatic rollback for that error. Doing nothing is recoverable; doing both is not.

## Validation

- [x] Tests added or updated
- [x] CI passes

`cargo nextest run --workspace --locked` 1588/1588, `cargo clippy --workspace --all-targets --all-features --locked` clean, `cargo fmt --all --check` clean.

Three tests, each mutation-checked:

- `execute_spec_stops_the_work_not_just_the_direct_child` fails on `main` and passes here. It starts a marker process as a **grandchild** (the shape every `sudo` action has) and asserts it is gone, matched by exact argv read from `/proc`. A substring search over full command lines would also match the test harness, which is how the original probe went wrong.
- `an_action_that_could_not_be_stopped_is_never_rolled_back_over` counts rollback executions and requires zero.
- `an_ordinary_failure_still_rolls_back` is its negative twin, so the veto cannot pass by disabling rollback altogether.

The pre-existing `execute_spec_kills_a_child_that_outruns_the_action_timeout` is left in place but is worth reading as a cautionary example: it asserts only that the call returns a timeout error quickly, which is equally true of a daemon that gives up waiting while the work keeps running. That is why this bug survived it.

## Notes for Reviewers

- `signal_group` treats `ESRCH` as success, since an empty group is the goal, not an error.
- The five second grace is a courtesy window for a package manager to unwind, not a second timeout; the deadline has already passed by then.
- `rpm-ostree` deserves a follow-up: it is a thin client to `rpm-ostreed`, so cancelling its transaction properly means `rpm-ostree cancel`, not signals. The group kill stops the client and now reports honestly when it cannot, which is strictly better than before, but it is not a full cancel.
- Signal handling is `unsafe` by necessity; both call sites carry a SAFETY note explaining why `pgid` can never be `0` (which would signal the daemon's own group).


vladimirrott:
Both review rounds are implemented. The reviews found the first version was materially broken, and I reworked it:

- **EPERM bug (critical):** `group_is_alive` read `killpg`→EPERM as *dead*, so the rollback veto never fired for the sudo/root case it was written for. The group probe now distinguishes ESRCH (gone) / EPERM (alive-but-unreachable) / other errno (unknown→alive); only ESRCH is "stopped".
- **Unbounded wait (critical):** I had removed the pid-targeted `start_kill` floor and left `child.wait()` unbounded, so a signal-ignoring root child made the timeout hang forever. Floor restored, every wait bounded.
- **Honest scope:** an unprivileged daemon cannot force-kill a root child, so containment is real for non-privileged commands and *detection + veto* for sudo actions. Real termination (systemd scope / privileged reaper) is deferred to #142.
- Slot is now held on `ActionNotStopped` so no second mutating action can race the first; stdin nulled on the progress spawn site (SIGTTIN parity).

New tests, all mutation-checked: production spawn path, SIGTERM→SIGKILL escalation, error-variant discrimination, and the veto driven end to end through the real dispatcher (severing the wiring boolean fails it). 1593/1593.

Issue #153, opened 2026-07-31
[known/accepted] daemon does not enforce human presence for approval (same-user)


vladimirrott:
Red-team finding F1.1, filed for the record and closed as accepted. A same-user process that connects to the daemon socket can drive preview→approve→execute without the terminal confirmation the CLI enforces. This is explicitly documented as out of the threat model in SECURITY.md (Layer 4): the approval boundary protects against an untrusted MCP agent, not against malware already running as the same Linux user. No change intended; recorded so the finding is not rediscovered as novel.

Issue #152, opened 2026-07-31
vsock pre-shared token is a plaintext bearer replayable by an observer


vladimirrott:
Red-team finding F6.1 (validated, deferred). The vsock token is sent as the first unencrypted frame and treated as sufficient authentication for the configured role (`SYSKNIFE_TOKEN_ROLE`, default Dev). An adjacent party who can observe one legitimate connection can replay the token and then mint a fresh one-time receipt. Inherent to a bearer token over an unencrypted channel. **Proposed directions:** channel binding, per-connection nonce/HMAC challenge, or requiring TLS/wireguard under vsock. Hard to fully close; documenting the residual risk and threat model is the near-term step. Not yet fixed.

Issue #151, opened 2026-07-31
AptAutoremove approval does not bind the package set that will be deleted


vladimirrott:
Red-team finding F1.2 (validated, deferred). The approved preview for `AptAutoremove` binds only the action name and empty params; `apt-get autoremove -y` resolves the deletion set (which can include kernels/drivers) at execution, so the executed effect is unbound by the approval and the request hash. **Proposed fix:** capture the `--dry-run`/`-s` deletion set at preview time and bind it into the approval, or raise the action to High. Documented in apt.rs:236. Not yet fixed.

Issue #150, opened 2026-07-31
vsock partial-frame DoS: 16 unauthenticated connections exhaust the pool


vladimirrott:
Red-team finding F6.4 (validated, deferred). `framing.rs` pre-allocates the full claimed body (`vec![0u8; len]`, up to 4 MiB) before reading it, and a permit is consumed before token auth. 16 unauth connections each sending a 4 MiB header and withholding the body hold the whole `MAX_CONNECTIONS` pool (~64 MiB) and can sustain the outage across the 10s auth deadline. **Proposed fix:** stream the body / cap the pre-auth allocation, and/or a separate smaller quota for unauthenticated connections. Not yet fixed.

PR #154, opened 2026-07-31
fix(security): close six red-team findings from a defensive audit

vladimirrott:
## Summary

A defensive red-team pass (six attacker profiles at medium reasoning effort, then every finding validated against the code and reproduced where safe) surfaced a set of real paths. This PR fixes six and files the rest as tracked issues.

The strongest was **[#143](https://github.com/lacs-project/sysknife/issues/143), a Dev→root privilege escalation**: the package-install actions validated the package name with the generic `validated_safe_arg`, which permits `/`, so `apt-get install /tmp/x.deb` and `rpm-ostree install /tmp/x.rpm` install a caller-supplied local file and run its maintainer scripts as root — from Medium-risk (Dev) actions.

## Fixed (issue + test each)

| Issue | Finding | Fix |
|---|---|---|
| #143 | Dev installs a local `.deb`/`.rpm` as root | `validated_install_package` rejects paths, `.deb`/`.rpm` suffixes, globs; routed all install/remove/hold package params through it |
| #144 | Dev activates `debug-shell`/`emergency`/`rescue` → root shell | `validated_activatable_unit` denylists root-shell units for start/restart/enable/unmask |
| #145 | AddAuthorizedKey follows an attacker symlink, appends as root anywhere | write runs as the target user via `sudo runuser -u <user> --` |
| #146 | `audit verify` wrong-machine caveat silent for config.toml-forwarded sockets | caveat resolves the socket like the client does, names the source variable |
| #147 | `publish-mcp` dispatch accepts an arbitrary ref under trusted OIDC | reachability guard (refuse commits not reachable from the default branch) + manifest-coherence gate in the workflow |
| #148 | swap helper follows a dangling symlink → root file creation | `O_EXCL|O_NOFOLLOW` create + `dd conv=nocreat` |

## Filed, not fixed (tracked)

Postgres TLS downgrade [#149](https://github.com/lacs-project/sysknife/issues/149), vsock partial-frame DoS [#150](https://github.com/lacs-project/sysknife/issues/150), unbound AptAutoremove approval [#151](https://github.com/lacs-project/sysknife/issues/151), vsock bearer-token replay [#152](https://github.com/lacs-project/sysknife/issues/152). [#153](https://github.com/lacs-project/sysknife/issues/153) (same-user approval) closed as documented and accepted in SECURITY.md Layer 4.

## Validation

- [x] Tests added or updated
- [x] CI passes

`cargo nextest run --workspace --locked` 1599/1599, `cargo clippy --workspace --all-targets --all-features --locked` clean, `cargo fmt --all --check` clean. `docs/action-reference.md` regenerated to show the `runuser` privilege drop.

Method note: the audit was run as a defensive review of our own daemon, not roleplayed exploitation. Two of the six findings had their fix reproduced directly (the mount-edit `O_NOFOLLOW` guard refusing a dangling symlink; the validator rejecting `/tmp/evil.deb`). The privilege-drop and denylist fixes are pinned by structural tests, since the cross-user symlink and console-shell effects are only observable with real accounts.

## Notes for Reviewers

- The `runuser` grant already exists in `packaging/sysknife-sudoers`, so #145 needs no sudoers change; the `sudoers_grants_match_argv` test still passes.
- #147 runs the coherence gate from the checked-out tree, which is safe *because* the reachability guard has already proven the tree is merged — an attacker's PR ref is rejected before the gate runs.
- The deferred items are real but need larger changes (TLS policy, framing rework, approval-binding); each has a concrete proposed direction in its issue.


vladimirrott:
Both review rounds implemented. The reviews found two of the six fixes were incomplete; reworked:

- **C1 (critical):** InstallPackages/RemovePackages still used the loose validator (my rewrite matched `"package"` but they take a `"packages"` list), so `rpm-ostree install /tmp/x.rpm` still ran as root. Fixed, plus new executor-level wiring tests over all 12 package cases — the tests that would have caught this, mutation-verified.
- **C2 (critical):** `runlevel1.target` (a stock alias of rescue.target) bypassed the denylist. Added, and the suffix-strip now covers all unit types.
- **Caveat false-positive:** my LISTEN_URI fix warned on every local verify. Now only SYSKNIFE_SOCKET or a vsock LISTEN_URI warns; local unix stays quiet. The forwarded-unix-via-config case needs daemon machine identity (#146). Added a test that `audit verify` actually prints the caveat (previously unpinned).
- **mount-edit TOCTOU:** dd now writes through the kept fd via /proc/self/fd (pass_fds), reproduced as closing the swap window.
- workflow `::error::` → stdout; sudoers comment corrected; CHANGELOG dedup; filed #155 (op_mount, same class) and noted the snap.rs sibling.

1604/1604, clippy --all-targets --all-features clean.

PR #169, opened 2026-08-03
test(e2e): reach every provider the product supports, and stop exporting an empty model

vladimirrott:
## Problem

The daemon supports **eight** LLM providers (`ProviderConfig` in `crates/sysknife-brain/src/config.rs`: anthropic, ollama, openai, gemini, groq, deepseek, mistral, xai). The E2E harness knew **three** of them.

Consequence, hit for real while setting up a 22.04 validation run: a `GROQ_API_KEY` exported on the host was silently dropped. `ubuntu-vm.sh` never forwarded it into the guest, and both story runners fell through to Ollama, which is not installed in the VM. Fifty stories failed, and the stderr line said `check OPENAI_API_KEY` while the key actually in play was Groq's. One missing variable, disguised as a broken suite.

## Fix

- `ubuntu-vm.sh` forwards `GROQ_API_KEY`, `DEEPSEEK_API_KEY`, `MISTRAL_API_KEY` and `XAI_API_KEY` in **both** the provision and run environments.
- `run-stories.sh` and `dev-stories.sh` detect those four. Existing precedence is untouched, so a host with several keys resolves exactly as before.

### Second, quieter bug

`run-stories.sh` ended with:

```sh
export SYSKNIFE_LLM_MODEL="${SYSKNIFE_LLM_MODEL:-${SYSKNIFE_TEST_MODEL:-}}"
```

which exports an **empty string** when neither is set. `BrainConfig` reads that with `env::var().ok()`, so `""` is `Some("")` and beats the per-provider default: the request goes out naming no model at all. The variable is now left unset in that case.

The per-provider default table went with it. Its three values were byte-identical to `DEFAULT_OPENAI_MODEL`, `DEFAULT_ANTHROPIC_MODEL` and `DEFAULT_GEMINI_MODEL`, so deleting the copy changes no behaviour and removes a second home that could drift from `config.rs`.

## The test is drift-proof by construction

`provider-parity.test.sh` derives the provider list from `config.rs` rather than restating it, so adding a ninth provider fails the test until the harness forwards it. It also refuses to run if the extraction finds fewer than seven keys, because a pattern that silently matches nothing would make every assertion vacuously true. That is precisely the failure mode being fixed next door in #168, and it would be embarrassing to reintroduce it in the test meant to prevent it.

Host-side only: greps scripts, needs no VM, no daemon, no network.

## Verification

- The test reports **21 failures before** this change and **passes after** (7 providers reachable).
- The real block from `run-stories.sh` was extracted and evaluated against fabricated environments:

| Environment | provider | model |
|---|---|---|
| `GROQ_API_KEY` only | `groq` | *unset* |
| `OPENAI_API_KEY` + `GROQ_API_KEY` | `openai` | *unset* (precedence preserved) |
| `GROQ_API_KEY` + `SYSKNIFE_TEST_MODEL` | `groq` | `some-model` |
| `GROQ_API_KEY` + explicit `SYSKNIFE_LLM_MODEL` | `groq` | `chosen` |
| `XAI` / `MISTRAL` / `DEEPSEEK` only | `xai` / `mistral` / `deepseek` | *unset* |
| no keys | `ollama` | *unset* |

- End to end against the live jammy VM with a dummy key: the guest dialled `groq/llama-3.3-70b-versatile` and Groq answered `401 Invalid API Key`, proving the variable now crosses the host boundary. Before this change the same invocation dialled Ollama and failed with a connection error.
- `cargo nextest run --workspace --locked` → 1653 passed.
- `ubuntu-vm-bootstrap.test.sh` still passes.


vladimirrott:
## Review round applied (`076a22d`), plus the rate-limit finding from the first live run

The reviewer caught this PR shipping the exact defect it was written to prevent, twice. Both are fixed.

**1. Nothing ran the test.** This repo has no auto-discovery for `*.test.sh` — each one is an explicit `run_step` in `scripts/ci-local.sh`, and `e2e.yml` only shellchecks them. So the drift guard could never fail, however far the harness drifted. Now wired into the hygiene group next to `ubuntu-vm-bootstrap.test.sh`.

**2. The test claimed repo-wide parity while checking three of five scripts.** The same bug was untouched in `atomic-vm.sh` (only three provider keys, in all three of its env lists) and `tests/e2e/exec/run-exec-stories.sh` (three-provider detection *and* the empty-model export, verbatim). Both fixed; the test now covers five entry points, two forwarding and three selecting.

### Third problem, found by actually running the suite

The first valid 50-story run against a live model exposed something neither reviewer could have seen statically. Stories 55 to 74 planned normally; everything after failed in ~40 ms with `rate limit exceeded`. Cause: the planner's own limiter, `DEFAULT_MAX_RPM = 20`. The suite generates roughly 30 requests/minute, and **exactly 20 stories** got through.

Two consequences, both fixed here:

- `SYSKNIFE_MAX_RPM` is honoured by the product but was forwarded by neither VM script, so no run could raise it. Both forward it now, and the parity test asserts they do.
- `run-stories.sh` reported throttled stories as `FAIL`, indistinguishable from a wrong plan. Worse, it reported them as **`PASS`** for stories 91 to 93, since those accept an empty plan and a throttled run produces exactly that. A throttled suite read as a healthy one. Rate-limited stories are now `RATELIMIT`, counted separately, and that verdict deliberately overrides the story's own verdict.

### Verification, by breaking things on purpose

| Check | Result |
|---|---|
| delete `GROQ_API_KEY` from `atomic-vm.sh` (3 occurrences → 0) | test goes pass → `does not forward GROQ_API_KEY`, exit 1 |
| restore it | passes again |
| `SYSKNIFE_MAX_RPM=1`, run 55 56 57 | `1/3 passed, 0 failed, 2 rate-limited` — not counted as failures |
| `SYSKNIFE_MAX_RPM=1`, run 55 91 92 93 | 91/92/93 all `RATELIMIT`. In the earlier throttled run these three printed `PASS` |
| full suite, `SYSKNIFE_MAX_RPM=200` | `44/50 passed, 0 rate-limited` |

An earlier attempt at the mutation proof used a `sed` that silently matched nothing and "passed" — the same vacuous-green trap in miniature, which is why the proof above quotes occurrence counts rather than just an exit status.

`cargo nextest run --workspace --locked` → 1653 passed. `ci-local: PASS`.

PR #171, opened 2026-08-04
feat(e2e): record the LLM call once, then replay the story suite offline and free

vladimirrott:
## Why

The Ubuntu story suite could only ever run against a live model. Getting the first valid 22.04 number cost a dead OpenAI key, a Mullvad exit IP that Groq blocks, a 20-rpm limiter that silently truncated the run to 20 stories, and about **$0.25** of tokens. Then two consecutive runs **disagreed on three stories**, because `gpt-oss-120b` intermittently emits its reasoning channel instead of a tool call.

So the suite was expensive, credential-bound, throttled, and not reproducible. A recorded number under those conditions is an anecdote, not a baseline.

## What

Record/replay at the `LlmProvider` boundary, ported in concept (not code) from `whystuck`'s `eval/cassette.py`.

One `record` run captures every distinct `complete` call to JSON; every later `replay` answers from the file. The wrapper sits at the provider boundary deliberately, so **prompt assembly, the tool-use loop, the safety fence, the ActionSpec risk substitution and the daemon IPC all still run for real**. Only the network hop is served from disk. Recording story *verdicts* instead would have stubbed out the planner and left assertions that pass without testing anything — the exact defect class #168 just removed.

Keyed on `sha256(surface, system, messages, tools, max_tokens)` where surface is `provider/model`. Order-independent, so stories may run in any order; and a recording made against one model **misses** rather than answering for another, because it is evidence about the model that produced it and nothing else.

## The two guards that matter

Both are taken from `whystuck`, where they caught a real retraction: a swallowed replay miss had turned *"accepted a change that hurt"* into *"the treatment never applied"* — the one distinction that study existed to make.

1. **A miss is an error, never a fallthrough to the network** — and every call also appends its outcome to a ledger. Provider errors propagate today, but a future retry or provider-fallback path would turn a miss into a live call against a different model with nobody the wiser.
2. **A replay that served nothing looks exactly like a replay where everything passed.** `run-stories.sh` requires `hits > 0 ∧ misses = 0`, and truncates the ledger per run so it cannot inherit an older tally.

That second one is not hypothetical here. Earlier today a throttled run reported stories 91/92/93 as `PASS` precisely because they accept an empty plan.

## Two improvements on the original

- **`meta.system_prompt_sha256`** — whystuck's miss message has to guess ("match the configuration the cassette was recorded under… and any lever that changes a prompt"). This one says *the prompt changed* and names the hash.
- **The version is probed before the document is parsed**, so a future entry shape reports a version mismatch instead of "not valid JSON", which would send the reader looking for corruption rather than for the right build.

## No rate limiter under replay

`DEFAULT_MAX_RPM` is consulted per `plan_intent` **before** the provider call, so leaving it installed throttled a 50-story replay at story 20 — on the strength of requests that were never sent. A replay reaches neither spend nor provider load, so the limiter is skipped.

## Proof

22.04, `openai/gpt-oss-120b`, 61 entries, 64 KB:

| Run | Result | Cassette |
|---|---|---|
| record | 46/50 passed, 4 failed, 0 rate-limited | 61 entries written |
| replay | **46/50 passed, 4 failed, 0 rate-limited** | **served 61, 0 misses** |

The replay ran under three adversarial conditions at once:

- `GROQ_API_KEY` set to a **deliberately invalid value**, so any real request would have 401'd → proves zero network calls
- **`SYSKNIFE_MAX_RPM` unset**, so the default 20 would have walled it at story 20 → proves the limiter is skipped
- ledger audit enabled → proves the run actually replayed rather than quietly doing nothing

Same four failures, same messages, per-story time down from **1.5-3 s to 0.12-0.29 s**. The four are real planner defects, now deterministic and reproducible by anyone: **62** `AptUnhold`, **73** `UfwStatus`, **92** a reserved port-0 firewall rule, **101** `DistroboxList`.

`grep` confirms no prompt text is stored verbatim — the ~39 KB system prompt is folded into the hash, which is why 61 entries fit in 64 KB.

## Guard, verified by mutation

`provider-parity.test.sh` now asserts both cassette variables reach the guest and that the runner audits the ledger. Verified the way the rest of this week's work was: deleting the forwarding from `ubuntu-vm.sh` takes the test from pass to two named failures; restoring it passes. A guard nobody has broken on purpose is not a guard.

The ledger is gitignored as per-run output. The cassette **is** committed — it is the fixture.

## Notes for review

- One arm added to the shell's `ProviderError` match, required to keep the workspace compiling after the new variant. A cassette miss gets its own code rather than borrowing `llm_parse_error`, which would send a reader chasing the model for a replay that simply had no recording. Not GUI work.
- A key must still be **present** under replay, because `BrainConfig::from_env` validates one before the planner exists. Its value is never read. I chose not to couple `config.rs` to the cassette for this; the docs say to use a deliberately invalid one, which doubles as the hermeticity proof.
- **Known gap:** the "no limiter under replay" behaviour is proven by the integration run above, not by a unit test — pinning it would mean building a `StateClient` mock inside `planner.rs` for a one-line condition. Worth revisiting if the CI job (phase 2) does not cover it.
- Serde derives were added to `Message`, `Role`, `ContentBlock`, `Completion`, `StopReason`. Additive, but it is public surface on a published crate.

`cargo nextest run --workspace --locked` → **1667 passed** (was 1653). `clippy -D warnings` clean. `cargo fmt --check` clean.

Phase 2 (a CI job that replays the suite as a merge gate) is deliberately a separate PR.


vladimirrott:
## Review round applied (`e1a8502`)

Two reviewers. Both found real defects, and the worst one was this PR **reintroducing the exact failure it exists to prevent**.

### CRITICAL — a capitalized env var disabled the whole audit

`CassetteMode::parse` trims and lowercases. `run-stories.sh` compared the mode **literally**. So `SYSKNIFE_CASSETTE_MODE=Replay` made the planner replay strictly while the harness silently skipped the ledger audit *and* never truncated the ledger.

Combine that with stories 91-93, which accept an empty plan on purpose and swallow the CLI error with `|| true`, and a subset run could **miss every single call and exit 0** with three PASSes and no `Cassette:` line at all.

Fixed on both layers, and reproduced against an empty cassette before and after:

| | before | after |
|---|---|---|
| stories 91/92/93 | `PASS`, `PASS`, `PASS` | 3 × `FAIL: replay had no recorded output for this call (cassette miss), which is not a refusal` |
| audit line | *(absent)* | `Cassette: replay served 0 call(s), 3 miss(es)` |
| exit code | 0 | 1 |

A miss is not a refusal, and the stories now say so themselves — so a subset run stays honest even without the harness.

### HIGH — silent ledger-write failures could produce a false clean audit

Write failures were swallowed with no log anywhere. If writes began failing partway through and the lost lines happened to be the **misses**, the audit read `served N, 0 misses` on a run that had missed. Replay now proves the ledger is writable **at open**, so a read-only checkout fails immediately with the real reason instead of later claiming "served 0 calls"; a later failure is printed loudly. Test added.

### MEDIUM — `jq -s` collapsed both counts on a truncated line

Slurping needs the whole file to parse, so one half-written trailing line from a process killed mid-suite took hits *and* misses to zero and reported "served 0 calls" for a run that had served plenty. Counted line by line now; an unreadable line counts as a miss, matching `read_ledger`'s own policy.

### MEDIUM — a crash path I introduced

A cassette misconfiguration made the Tauri shell fail `from_config`, retry with Ollama defaults, read the same environment, fail identically, and panic on `expect` with a message naming `SYSKNIFE_LLM_PROVIDER`. My change (having `from_config` read the cassette env) created that path. It now says what is actually wrong.

### The second review was right and my excuse was wrong

I claimed the `from_config` wiring test would need a `StateClient` mock built from scratch. `planner.rs` already has an in-crate test module with access to the private `rate_limiter` field. Test added: the limiter is absent under replay and present otherwise.

### Smaller, also from review

- `store()` rolls back if `persist()` fails, so an entry cannot be visible to `lookup` while absent from disk.
- The `parent()` fallback in `persist()` was unreachable — a bare filename yields `Some("")`, not `None`. Rewritten as a filter so it actually fires.
- **`NamedTempFile` creates at 0600 and the rename kept it.** Recording as root inside a VM left a 0600 root-owned fixture in the synced tree, and every later `ubuntu-vm.sh sync` failed on it with "Permission denied". Cassettes are 0644 now, with a test. The directory-ownership case that a file mode cannot fix is documented.

### Guard, verified by mutation

`provider-parity.test.sh` asserts the shell normalisation and the per-story miss guards. Reverting the normalisation takes it from pass to `run-stories.sh does not normalise SYSKNIFE_CASSETTE_MODE the way CassetteMode::parse does`.

### Re-verified end to end

Deliberately with `SYSKNIFE_CASSETTE_MODE=Replay` capitalized, an invalid API key, and no `SYSKNIFE_MAX_RPM`:

```
Summary: 46/50 passed, 4 failed, 0 skipped, 0 rate-limited
Cassette:  replay served 61 call(s), 0 miss(es)
```

Same result as before the fixes, with the audit now armed where it used to be silent.

`cargo nextest run --workspace --locked` → **1670 passed**. `clippy -D warnings` clean. `ci-local: PASS`.

PR #180, opened 2026-08-05
fix(security): close five privileged-path holes, harden the approval prompt, and make audit verify read its anchor

vladimirrott:
Started as an evidence-pipeline fix and grew into a security batch after a
review sweep across the docs and the code. Every fix below is preceded by a
test that fails without it.

## Planner / evidence (original scope)

- Split what the planner may *offer* from what the executor may *run*, so a
  distro-agnostic tool schema stops producing plans the daemon will reject.
- Made the evidence pipeline actually emit evidence, and recorded the first
  artifact-backed Ubuntu 22.04 story run.
- Published figures now derive from `tests/evidence/`; the guard blocked four
  commits in this branch until the docs matched the run.

## Security

| Area | Before | After |
|---|---|---|
| `packages/setup/install-binary.js` | `sudo install` copied from `/tmp/sysknife-install-${pid}-…` at 0644 — a predictable name in a world-writable directory, so a local user could swap the bytes before the copy and land a root-owned 0755 binary | `mkdtemp` (0700 dir, 0600 payload), removed in a `finally` |
| `ConfigureLogRotation` | any absolute path became the stanza header of a root-run logrotate config, so a Medium-risk Dev-tier action could schedule truncation of `/etc/shadow` | confined to `/var/log/`, in the helper *and* `validated_log_path` |
| `RemoveSwap` | charset check then root `os.unlink` — `--file /etc/shadow` deleted it, since swapoff runs `check=False` and the fstab rewrite no-ops | only paths the host reports as swap files (`/proc/swaps` type `file`, or an fstab swap entry), never a symlink |
| Kernel-arg removal | unscreened, on the premise that "removing a dangerous arg is always safe" — which inverts the risk | KSPP-grounded screen on both `SetKernelArguments` and `GrubSetKargs`, with an escape hatch so weakening args stay removable |
| `executor.rs` | a `next_line()` error (InvalidData on non-UTF-8 stdout) or failed `wait()` returned via `?`, orphaning a **root** process group while reporting failure | all three sites `kill_and_reap` first; `StopCause` keeps a read failure from being logged as a timeout |
| Approval prompt | model- and daemon-sourced text printed raw — `\x1b[2K\r` rewrites its own line, `\x1b[1A` walks over the risk badge, length alone scrolls the plan away | `operator_safe` on 8 sites including the prompt string itself; drops C0/DEL/**C1** (U+009B is CSI in 8-bit mode), bidi overrides, zero-width; caps length |
| Audit DB | created 0644 | 0600 at open; the WAL inherits it |
| `audit verify` | reported `audit_anchor: {configured: true}` and never read the anchor — so an operator who set anchoring up was *worse* informed than one who did not, who at least got the caveat | cross-checks against anchored checkpoints; a truncation reaches the exit code |
| RUSTSEC-2026-0097 | `--ignore`d in CI | removed: `rand 0.7.3` left the lockfile and `cargo audit` exits 0 unignored. An ignore for a crate that is not built only arms a silent re-accept |

`client.rs` claimed *all* daemon output was ANSI-stripped. Only command output
is; envelopes deserialize straight through. Doc corrected.

## Not fixed here

The `/usr/bin/sh` and `/usr/sbin/runuser` sudoers grants make the daemon
root-capable. That is a **pre-existing, documented decision** with a TODO in
`packaging/sysknife-sudoers`, unchanged since 0.6.0 — not a regression from this
branch. Removing them needs `ConfigureFirewall` and the container/toolbox/ssh-key
paths moved to root-owned helpers, and sudoers wildcards cannot substitute
because sudo-rs permits `*` only as the final argument. Tracked separately.

Related: the `env … apt-get *` grant's comment cites `-o
APT::Update::Pre-Invoke=` as why a bare `apt-get` grant is a GTFOBins escape —
the env-wrapped grant with a trailing `*` permits the same escape. Not reachable
today (every argument is validated and a leading `-` is rejected), and moot while
`sh` is granted, but the stated reasoning is wrong.

## Verification

- 1,711 tests (from 1,681 at branch point), clippy `--all-features --all-targets`
  clean, `cargo doc -D warnings` clean.
- The orphaned-child test was mutation-checked: reverting the error path leaves
  the group alive and the test names the surviving pgid.
- Three new helper suites wired into CI (`rmswap`, `log-edit`, `grub-kargs-edit`).
- The pre-commit hook now runs all four checks `ci.yml` runs; it was previously
  a strict subset, which is how a broken intra-doc link reached CI.


vladimirrott:
Second commit added: the three blocking findings from the #176 review, all verified against the code first.

**The fence was answering two questions.** `FEDORA_ONLY_ACTIONS` is read by the daemon's `validate_action_platform`, the CLI routing guard *and* the planner's filter. Putting firewalld/toolbox in it fixed the planning defect and made them **unrunnable** on Ubuntu — where both are installable — so an Ubuntu host running firewalld lost firewall management and `UfwStatus` would have answered `Status: inactive`: a confident wrong answer where a refusal was expected. Those five now live in `NON_CANONICAL_ON_DEBIAN`, read only by the planner.

**Eligibility and detection are mutation gates**, per docs/distro-support.md ("the daemon refuses every *mutating* action when it is false"). The gate short-circuited only actions with *no* family requirement, so tagging a read-only action subjected it to `is_supported()` too. That was already live before #176: on Debian, `AptShow` was refused — apt is native there — because Debian is not an eligible *target*. Reads are now exempt from eligibility and detection while the family check still applies to them, and the CLI guard shares the same `min_role_for_action` so it cannot refuse what the daemon would run.

**The class was sampled, not removed.** Ten rpm-ostree/DNF actions sat outside the fence while `AddLayeredPackage` — same `rpm-ostree install` argv as `InstallPackages` — sat inside it. Nine are now fenced, and the classification is **derived from each action's argv**, so an action driving rpm-ostree that is missing from the list fails the build. `GetSystemState` is the tenth, deliberately exempt with the reason pinned by a test, and filed as #181: it is woven through fifteen places in the shared validated prompt blocks and Ubuntu has no equivalent to put there.

Deriving the prompt isolation immediately found what the hand-written list of nine had missed: the Debian prompt documented four `/etc/yum.repos.d` actions' parameters in a cross-distro block, named `GetPendingUpdates` twice as "the Fedora equivalent", and used `UpdateSystem` as Example C's history filter.

Also moved the routing guard before the approval prompt (it ran after, so a Fedora host printed a plan, asked for confirmation, then refused).

Test fallout was the tests being wrong: four integration suites and two unit tests used rpm-ostree actions as generic vehicles while `DaemonState::open` detects the *real* host, so they passed only because the fence was incomplete. Every rollback-capable action is rpm-ostree shaped, so those suites declare a Fedora host; the Ubuntu support-floor test switches to `AptUpgrade`.

1688 tests pass, clippy clean. The test-count guard from the first commit earned itself on the way: it failed, named the three docs still saying 1,681, and refused to pass until the figure matched the run.

Issue #228, opened 2026-08-17
The no-secrets gate's "never echo the credential" assertion cannot fail under pipefail


vladimirrott:
`tests/release/no-secrets.test.sh` guards the secret scanner in both directions, and its fourth check is that the scanner must not print the credential it found (`tests/release/no-secrets.test.sh:88`):

```bash
# --- 4. It must never print the credential it found -------------------------
leak="$(make_key 'gsk_' 52)"
printf 'k = "%s"\n' "$leak" > "$tmp/echo.txt"
if "$CHECK" "$tmp/echo.txt" 2>&1 | grep -qF "$leak"; then
    echo "FAIL: the scanner echoed the credential it found — that is a second leak"
    fail=1
fi
```

The file runs under `set -euo pipefail` (`:18`). `scripts/check_no_secrets.sh` exits 1 when it finds a key. Under `pipefail` the pipeline status is the rightmost non-zero status, so with `grep` matching (0) and the scanner exiting 1, the pipeline is 1 and the `if` is false. The branch cannot execute for any input.

## Reproduction

```bash
$ bash -c 'set -euo pipefail
  fake() { echo "found key: gsk_LEAKED"; return 1; }
  if fake | grep -qF "gsk_LEAKED"; then echo FIRED; else echo "DID NOT FIRE"; fi'
DID NOT FIRE

$ bash -c 'set -eu
  fake() { echo "found key: gsk_LEAKED"; return 1; }
  if fake | grep -qF "gsk_LEAKED"; then echo FIRED; else echo "DID NOT FIRE"; fi'
FIRED
```

## Why it matters

This gate runs in CI (`.github/workflows/ci.yml`) and prints its own pass line. Changing `"${hit:0:7}"` to `"$hit"` in `scripts/check_no_secrets.sh` would spill every discovered credential into CI logs and into pre-commit output on every developer machine, and this gate would still report that findings never echo the secret.

The other three checks in the file are sound. This is the one that cannot fail.

## Scope

Capture the output before matching, so the scanner's exit status stops driving the pipeline:

```bash
out="$("$CHECK" "$tmp/echo.txt" 2>&1 || true)"
if grep -qF "$leak" <<<"$out"; then ...
```

## Tests first

The fix needs a mutation proof, since the bug is that the assertion is vacuous. Point `$CHECK` at a stub that echoes the full credential and exits 1, confirm the repaired guard fails, then restore. That mutation is what proves the guard bites, and it is the step whose absence let this ship.

[...]

Issue #222, opened 2026-08-17
The systemd unit recreates /run/sysknife and /var/lib/sysknife at 0755, discarding the tmpfiles 0750/0700


vladimirrott:
`packaging/sysknife-tmpfiles.conf:11` asks for a private runtime directory:

```
d /run/sysknife     0750 sysknife sysknife -
d /var/lib/sysknife 0700 sysknife sysknife -
```

The unit then re-creates both through systemd (`packaging/sysknife-daemon.service:34`):

```ini
RuntimeDirectory=sysknife
StateDirectory=sysknife
```

Neither `RuntimeDirectoryMode=` nor `StateDirectoryMode=` appears anywhere in the repository, so both take systemd's default of `0755`. `RuntimeDirectory=` is removed at unit stop and recreated at start, so the tmpfiles mode does not survive a single `systemctl restart`.

## Why it matters

`SECURITY.md:196` describes the directory as the first gate:

> `sysknife:sysknife 0750`. A user needs two group memberships to use SysKnife

`docs/vm-daemon-setup.md` makes the same claim, that a non-member of the `sysknife` group is refused before authentication happens because the directory is not traversable. After any restart it is world-traversable, so that layer is gone and the socket's own `0660 sysknife:sysknife` (`crates/sysknife-daemon/src/transport/listen.rs:138`) is the only thing left. Access control still holds, with one layer where the docs promise two, and `/var/lib/sysknife` becomes world-listable. The database file itself is force-chmodded to `0600` (`crates/sysknife-daemon/src/transactions.rs:1279`), and the error text three lines below says "The parent directory should be 0700 — check it."

Worth stating plainly: this is a defence-in-depth regression, not a live hole. Nothing is readable that was not readable before, because the socket and the file modes still gate.

## Scope

- `packaging/sysknife-daemon.service`: add `RuntimeDirectoryMode=0750` and `StateDirectoryMode=0700` so the unit states what the tmpfiles fragment already asks for.
- `tests/e2e/ubuntu-provision.sh:328` carries the same gap and should match.

## Tests first

This is invisible to the Rust suite, which is why it survived three validated live-VM runs. The check belongs beside the other release gates in `tests/release/`: assert that the unit declares both modes, and that they agree with `sysknife-tmpfiles.conf`. Deriving the expected values from the tmpfiles fragment rather than hardcoding them keeps the two sides from drifting again.

## Difficulty

`easy`. Two directives, plus a gate so it cannot regress.

## Getting started

[...]

PR #243, opened 2026-08-18
fix(setup): refuse malformed MCP config

### LINKED ISSUE #225: A malformed .mcp.json is replaced rather than refused, discarding the user's other MCP servers
`mergeMcpServers` refuses to overwrite a config it cannot read, and says why (`packages/setup/mcp-config.js:25`):

```js
} catch (e) {
  // A read failure is NOT the same as malformed JSON. Treating EACCES,
  // EIO, or a symlink loop as "start empty" makes the caller write a
  // fresh file over the top, silently deleting every other MCP server the
  // user had configured — the exact clobbering this function exists to
  // prevent. Refuse instead.
  throw new Error(...)
```

Eight lines later, the parse branch performs exactly that clobbering (`mcp-config.js:42`):

```js
} catch {
  // Malformed JSON — start from an empty config rather than crashing the
  // wizard. There is nothing to preserve in a file we cannot parse.
  existing = {};
}
```

The caller writes the result unconditionally, under a comment that states the promise this branch breaks (`packages/setup/index.js:747`):

```js
// Merge into any existing .mcp.json so the user's other MCP servers survive.
const mcpExisted = fs.existsSync('.mcp.json');
const mcpConfig  = mergeMcpServers('.mcp.json', mcpServers);
```

## Why it matters

"There is nothing to preserve in a file we cannot parse" is the part that does not hold. A trailing comma or a `//` comment is common in editor-managed MCP configs, and such a file still holds the user's other servers. The wizard replaces it with a sysknife-only config, keeps no backup, and reports success. The same path handles `.cursor/mcp.json`.

The two branches disagree about the same question, and the one that already reasoned it through is the read branch.

## Scope

Pick one and apply it to both `.mcp.json` and `.cursor/mcp.json`:

- Refuse, matching the read branch, and tell the user to fix or move the file. Consistent with the existing reasoning.
- Or write `<file>.bak` before replacing, and say so on stdout.

Either way the outcome must not silently contradict the comment on `index.js:747`.

## Tests first

`packages/setup/tests/` covers the merge path but not a malformed existing file, which is why this survived. Write the failing test first: a `.mcp.json` containing a trailing comma plus a second MCP server must either abort or leave a recoverable backup, and must not report a plain success.

## Difficulty

`easy`. One branch, plus the caller's message. Self-contained in the npm package, no Rust required, which makes it a good first issue for anyone more comfortable in JavaScript.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands. `npm test --prefix packages/setup` runs this package's suite. No CLA and no copyright waiver. The project is MIT.

ITSMERNB:
## Summary
- refuse to overwrite an existing malformed MCP config instead of replacing it
- preserve the original file untouched on parse failure
- include the config path and remediation in the error

Fixes #225

## Testing
- `node --test packages/setup/tests/mcp-config.test.mjs` — 8/8 passed
- `node --check packages/setup/mcp-config.js`
- `git diff --check`
- full setup suite was also run; its MCP tests pass, while 17 unrelated Linux-specific tests fail on native Windows

### DIFF
--- a/packages/setup/index.js
+++ b/packages/setup/index.js
@@ -743,18 +743,24 @@ async function main() {
 
   // ── Claude Code ──────────────────────────────────────────────────────────
 
+  // Parse every selected JSON config before writing any integration files. A
+  // refusal must leave the operator fully unconfigured rather than half configured.
+  const claudeMcpPath   = '.mcp.json';
+  const cursorMcpPath   = path.join('.cursor', 'mcp.json');
+  const claudeMcpConfig = doClaude ? mergeMcpServers(claudeMcpPath, mcpServers) : null;
+  const cursorMcpConfig = doCursor ? mergeMcpServers(cursorMcpPath, mcpServers) : null;
+
   if (doClaude) {
     // Merge into any existing .mcp.json so the user's other MCP servers survive.
-    const mcpExisted = fs.existsSync('.mcp.json');
-    const mcpConfig  = mergeMcpServers('.mcp.json', mcpServers);
+    const mcpExisted = fs.existsSync(claudeMcpPath);
 
     // .mcp.json may contain provider API keys in plain text. Restrict to owner
     // read/write so a coworker on a shared workstation (or a stray `cat *` in a
     // build script) cannot recover them. `chmodSync` is idempotent and also
     // tightens permissions on a pre-existing file that was created with the
     // process umask before this change.
-    fs.writeFileSync('.mcp.json', JSON.stringify(mcpConfig, null, 2) + '\n', { mode: 0o600 });
-    fs.chmodSync('.mcp.json', 0o600);
+    fs.writeFileSync(claudeMcpPath, JSON.stringify(claudeMcpConfig, null, 2) + '\n', { mode: 0o600 });
+    fs.chmodSync(claudeMcpPath, 0o600);
     ok(`${mcpExisted ? 'Updated' : 'Created'} .mcp.json  (${targets.length} target${targets.length > 1 ? 's' : ''}: ${targetSummary})`);
 
     if (!fs.existsSync('.claude')) {
@@ -785,11 +791,9 @@ async function main() {
       fs.mkdirSync('.cursor', { recursive: true });
     }
     // Merge into any existing .cursor/mcp.json so other MCP servers survive.
-    const cursorPath    = path.join('.cursor', 'mcp.json');
-    const cursorExisted = fs.existsSync(cursorPath);
-    const cursorMcp     = mergeMcpServers(cursorPath, mcpServers);
-    fs.writeFileSync(cursorPath, JSON.stringify(cursorMcp, null, 2) + '\n', { mode: 0o600 });
-    fs.chmodSync(cursorPath, 0o600);
+    const cursorExisted = fs.existsSync(cursorMcpPath);
+    fs.writeFileSync(cursorMcpPath, JSON.stringify(cursorMcpConfig, null, 2) + '\n', { mode: 0o600 });
+    fs.chmodSync(cursorMcpPath, 0o600);
     ok(`${cursorExisted ? 'Updated' : 'Created'} .cursor/mcp.json  (${targets.length} target${targets.length > 1 ? 's' : ''}: ${targetSummary})`);
 
     const rulesDir = path.join('.cursor', 'rules');
--- a/packages/setup/mcp-config.js
+++ b/packages/setup/mcp-config.js
@@ -9,9 +9,10 @@ const fs = require('fs');
 
 /**
  * Return an MCP config object with `servers` merged under `mcpServers`,
- * preserving any existing servers and unrelated top-level keys. A missing file
- * is treated as an empty config; an unreadable or malformed existing file is
- * refused so the caller cannot overwrite user-managed MCP entries.
+ * preserving any existing servers and unrelated top-level keys. A missing,
+ * empty, or whitespace-only file is treated as an empty config; an unreadable,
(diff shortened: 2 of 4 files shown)


vladimirrott (changes requested):
Thanks for taking this one, and for picking the behaviour that matches the read-failure branch instead of writing a `.bak`. A user who fixes their JSON keeps every server they had.

I ran the suite on Linux: `npm test --prefix packages/setup` gives 80/80, so those 17 failures were the Windows-only paths. I also drove the wizard to see what an operator gets:

```
$ node index.js --no-prompts --no-binary --claude --daemon-mode=skip
  ✗  existing MCP config at .mcp.json is malformed JSON: Expected double-quoted property name in JSON at position 47 (line 3 column 1) ... refusing to overwrite it; fix or move the file and run setup again
$ echo $?
1
```

Nothing written, the file byte-identical, exit 1. Inverting the two tests that asserted the old contract, rather than leaving them to contradict the code, is the part I liked most.

Four changes before I merge.

**1. A run with both integrations writes four files and then refuses.** `--claude --cursor`, a valid `.mcp.json`, a malformed `.cursor/mcp.json`:

```
  ✓  Updated .mcp.json  (1 target: sysknife)
  ✓  Created .claude/hookify.require-sysknife-approval.local.md
  ✓  Created .claude/hookify.sysknife-schema-fetch.local.md
  ✓  Created .claude/hookify.sysknife-bash-guard.local.md
  ✗  existing MCP config at .cursor/mcp.json is malformed JSON: ...
```

The operator lands half configured with four ticks on screen. #241 fixed this same shape in the audit-database migration: run the step that can refuse before the step that commits. Parse both files at the top of the write section (`index.js:749` and `:790` both call `mergeMcpServers`), then write.

**2. An empty file blocks setup and saves nothing.** `: > .mcp.json` now gives `malformed JSON: Unexpected end of JSON input`, and the remedy asks the user to fix a file that holds nothing. The comment you deleted covered that case. Treat empty or whitespace-only content as a missing file.

**3. A UTF-8 BOM hits the same abort.** `﻿` in front of valid JSON throws, and the file looks correct in the editor. Strip the BOM before `JSON.parse` and the user's other servers survive the merge.

**4. `mcp-config.js:39` still overwrites valid JSON that is not an object.** `[]`, `null` and `"x"` fall through to `existing = {}`. You lose nothing worth keeping in those files, though the docstring now promises more than the code delivers.

One optional hardening. `.mcp.json` holds provider API keys, and Node's parse error quotes the first ten characters of the input when the leading token is invalid: `Unexpected token 'A', "ANTHROPIC_"...`. CI and the story harness redirect stderr to files, the same path that put #206's credential on disk. Keep the position clause and drop the rest of `e.message`.

I approved the workflow runs, so checks should appear on both your PRs. A red security-audit is RUSTSEC-2026-0258 in h2 and mine; the fix landed on main in #245.

vladimirrott:
On the red security-audit: your run checked out a merge with main at f9e9907, which predates the h2 bump in #245. This PR leaves `Cargo.lock` alone, so nothing on your side triggers it. Your next push builds a fresh merge against current main and the check turns green.

ITSMERNB:
Addressed all four requested cases in 80d2f1b: both selected MCP configs are preflighted before writes, empty/whitespace files are treated as fresh, UTF-8 BOM is accepted, and non-object JSON is refused. I also took the optional hardening and stopped parse errors from echoing source contents while retaining the position when Node provides it. Focused MCP/setup-contract tests are 18/18; the full Windows setup run is 68/85 with the same 17 existing Unix-specific failures.

vladimirrott (approved):
This is what I hoped a first contribution would look like, and you took the optional hardening on top.

Verified on Linux at 80d2f1b:

- setup suite 85/85
- the wizard refuses before it writes. With `--claude --cursor`, a valid `.mcp.json` and a malformed `.cursor/mcp.json`, the run exits 1 with two files on disk and no `.claude` directory. Before the preflight it wrote four files and printed four ticks.
- empty file and UTF-8 BOM both proceed and keep the user's other servers
- `[]`, `null` and `"x"` are refused with the object message
- a config whose first token is a key no longer echoes it; the error carries the position clause and nothing else

I mutation-tested both new guards rather than trust the green run. Removing the preflight makes `all selected MCP configs are validated before any integration file is written` fail. Echoing raw `e.message` again makes the no-echo test fail.

Your numbers check out as well: 12 mcp-config tests plus 6 setup-contract tests is the 18 you reported, and 85 minus the 17 Unix-specific ones is your 68.

Approving. If you want another in this lane, #228 is a bash assertion that cannot fail because of `pipefail`, the same shape as this one, and #222 is a small Rust fix in the systemd unit. Thanks for carrying this through a review round.

vladimirrott:
This shipped in [v0.9.0](https://github.com/lacs-project/sysknife/releases/tag/v0.9.0), out on crates.io and npm now. `npx sysknife-setup@0.9.0` runs your preflight.

The release notes name the fix and the CHANGELOG entry describes it, since the behaviour change is the kind a user needs to read about: a run that used to succeed against a malformed config now exits 1 and keeps their servers.

Worth knowing for next time: watching releases here is the shortest way to see your own work go out, and new `good first issue` entries show up in the same feed.

ITSMERNB:
Really good to see this already live in 0.9.0. Thanks for taking the time to verify it on Linux, carry it through the release, and explain the release flow. I’ll keep an eye on the release feed from here — and it’s especially satisfying to know the preflight is now what users actually get from `npx sysknife-setup@0.9.0`.

Issue #233, opened 2026-08-17
Story coverage map: 64 Ubuntu and cross-family actions have no story, clustered into takeable pieces


vladimirrott:
64 of the 164 Ubuntu-and-cross-family actions are named by no story at all, and 98 of them are named by no story that runs on Ubuntu. This issue is the map, so the work can be taken one cluster at a time.

## Derivation

```bash
python3 - <<'PY'
import re, glob
from collections import Counter
rows = re.findall(r'^\| `([A-Za-z0-9_]+)` \|.*?\| (All|Ubuntu|Fedora) \|',
                  open('docs/action-reference.md').read(), re.M)
U = {n for n, d in rows if d in ('All', 'Ubuntu')}
named = set()
for f in glob.glob('tests/e2e/stories/*.sh'):
    named |= set(re.findall(r'"([A-Za-z0-9_]+)"', open(f, errors='replace').read()))
ub = [f for f in glob.glob('tests/e2e/stories/*.sh')
      if re.search(r'^# Story \d+ \([^)]*ubuntu[^)]*\)', open(f, errors='replace').read(), re.M)]
named_ub = set()
for f in ub:
    named_ub |= set(re.findall(r'"([A-Za-z0-9_]+)"', open(f, errors='replace').read()))
print('|U| =', len(U), 'uncovered =', len(U - named), 'not on ubuntu =', len(U - named_ub))
PY
|U| = 164 uncovered = 64 not on ubuntu = 98
```

Catalogue split: 189 tabled rows, `All` 97, `Ubuntu` 67, `Fedora` 25. 79 story files carry an `ubuntu` header tag.

Two honest caveats. This counts an action as covered when its name appears as a quoted literal anywhere in a story, which over-counts: `ConfigureWifi`, `GrubSetKargs` and `SetLocale` appear only inside "must not confuse with" comments. And 34 Ubuntu-and-`All` actions are covered only by stories 1 to 54, the atomic family, which no recorded Ubuntu run includes, so `AptListUpgradable`, a Debian-only action, has its sole coverage in a Fedora-family story.

## Clusters, each one a reasonable PR

Ubuntu-family exclusive, highest priority:

- **fail2ban** (4): tracked separately in #219
- **GRUB kargs** (1): `GrubSetKargs`

Cross-family (`All`):

- **sysctl and auditd** (5): `GetSysctl`, `SetSysctl`, `GetAuditRules`, `AddAuditRule`, `RemoveAuditRule`
[...]

Issue #229, opened 2026-08-17
Nothing checks the story-coverage figures in CONTRIBUTING.md, which is how they drifted


vladimirrott:
> **Corrected 2026-09-03 at `60af0ad`.** This issue was filed on 2026-08-17. PR #242 landed the next day and fixed two of the three claims below. One is still wrong, and the guard the issue asks for still does not exist. The original text is kept underneath so the history reads straight.

`CONTRIBUTING.md` publishes story-coverage figures that nothing derives from the story files.

## What is still wrong at `60af0ad`

`CONTRIBUTING.md:50` says:

> The suite is 133 stories: 54 atomic + 79 Ubuntu. **Every Debian-only action now has one.**

`CONTRIBUTING.md:47`, four lines above it in the same table, says the opposite:

> Remaining: story coverage for the cross-family actions, and **five Debian-only ones that still have none**: the four fail2ban actions and `GrubSetKargs`.

Line 47 is the true one. Recounted at `60af0ad`:

```
$ for a in Fail2banStatus Fail2banBanIp Fail2banUnbanIp ConfigureFail2banJail GrubSetKargs; do
    echo "$a: $(grep -rl "\"$a\"" tests/e2e/stories/ | wc -l)"
  done
Fail2banStatus: 0
Fail2banBanIp: 0
Fail2banUnbanIp: 0
ConfigureFail2banJail: 0
GrubSetKargs: 0
```

The story totals and the uncovered counts are correct today. Both recounted at `60af0ad`:

```
$ ls tests/e2e/stories/story-*.sh | wc -l
133
$ # family from each story's own header line
ubuntu=79 atomic=54
```

and the derivation below now returns `{'All': 59, 'Fedora': 10, 'Ubuntu': 5}`, which is what `CONTRIBUTING.md:50` prints.

So the numbers are right and nothing keeps them right. They were wrong for a day in August, somebody fixed them by hand, and the next story added or retired puts them back out.

## The guard, which is the part worth doing

`scripts/check_evidence_claims.py` has no rule that derives an uncovered-action count. Its checks at `60af0ad`:

```
$ grep -nE '^def check_' scripts/check_evidence_claims.py
124:def check_figure
149:def check_bare_test_totals
194:def check_story_claims
261:def check_bare_story_counts
306:def check_action_figures
382:def check_validated_tiers
```

[...]

Issue #221, opened 2026-08-17
audit verify: an inconclusive anchor masks a broken chain, and --json status ignores the anchor


vladimirrott:
`AuditVerification::exit_code` is deliberate about precedence (`crates/sysknife-daemon/src/audit_chain.rs:1372`):

```rust
/// A detected tamper (`1`) outranks an inability to check (`2`): if the
/// chain is provably broken, reporting "could not verify" because some
/// *other* check was inconclusive would understate what is known.
pub fn exit_code(&self) -> i32 {
    let codes = [...];
    if codes.contains(&1) { 1 } else if codes.contains(&2) { 2 } else { 0 }
}
```

The CLI then folds in the anchor verdict with `max` (`apps/sysknife-cli/src/runner.rs:922`):

```rust
let exit_code = outcome.exit_code().max(anchor_exit);
```

`max(1, 2)` is 2. So the one place that combines a chain verdict with an anchor verdict applies the opposite of the rule stated one crate away, and `docs/the-audit-chain.md` states the same rule the comment does.

## Why it matters

A provably tampered chain exits 2, "could not check", whenever the anchor cross-check is inconclusive. `verify_configured_anchor` returns `CannotVerify` when the checkpoint database cannot be reached (`runner.rs:1059` and `runner.rs:1070`), so an attacker who edits a row and then blocks the route to the checkpoint database converts a tamper verdict into an ops-problem verdict. The 1-versus-2 split exists to keep those apart, and `apps/sysknife-cli/src/cli.rs:142` tells CI operators to rely on it.

## A second, opposite bug in the same function

The JSON headline is computed from the local verdict only (`runner.rs:1361`):

```rust
"status": status_word(verification.exit_code()),
```

That is `verification.exit_code()`, not the combined `exit_code` on line 922. So an intact chain with a `Truncated` anchor exits 1 while `--json` reports `"status": "intact"`. A monitor keyed on that field reads a truncated audit trail as healthy.

So the exit code ignores precedence, and the JSON status ignores the anchor. Both belong in one fix.

## Scope

- `runner.rs:922`: combine the anchor through the same `contains(&1)` precedence rather than `max`.
- `runner.rs:1361`: derive `status` from the same combined verdict the exit code uses.

Nothing in `audit_chain.rs` needs to change; its own helper is already right.

## Tests first

[...]

PR #244, opened 2026-08-18
chore: remove verified dead items

### LINKED ISSUE #235: Delete four verified-dead items: an unused dependency, a stale allow, an unread constant, and dead re-exports
Four items are verifiably dead. Each was proved by deleting it on a workspace copy and running the exact CI command clean:

```
cargo clippy --workspace --all-features --all-targets --locked -- -D warnings
```

Each is independent, so this can land as one PR or four.

### 1. Unused dependency `prost-types`

`crates/sysknife-proto/Cargo.toml:15` declares it and nothing uses it:

```bash
$ grep -rn 'prost_types' --include=*.rs crates apps | wc -l
0
$ grep -n 'import\|google.protobuf' crates/sysknife-proto/proto/sysknife/v1/sysknife.proto
$ # no output: the .proto imports no well-known types, so prost-build never emits prost_types paths
```

`sysknife-proto` is its only consumer, so the `[workspace.dependencies]` entry in the root `Cargo.toml` goes with it. Note that `prost` itself is live, 30 references in the generated code; only the `-types` crate is not.

### 2. A stale lint suppression

`crates/sysknife-core/src/config.rs:439` carries `#[allow(unused_unsafe)]` over an `unsafe { std::env::set_var(...) }`. `set_var` is an `unsafe fn` on current rustc, so the block is used and the lint cannot fire. Around 25 sibling `unsafe { std::env::set_var(...) }` sites in this repo carry no such allow and pass CI.

Worth knowing: this is the only stale one. Stripping all eight `#[allow(...)]` attributes in scope produces warnings for seven of them, so the rest are load-bearing.

### 3. `PRODUCTION_LISTEN_URI` and its tautological test

`crates/sysknife-core/src/lib.rs:11` declares the constant and nothing reads it, in Rust, TOML, service files, scripts, the Makefile, `packaging/`, or docs. Its only test asserts:

```rust
PRODUCTION_LISTEN_URI.starts_with("unix:///run/")
```

which re-reads the literal declared four lines above and would pass however wrong the value became. Its sibling `PRODUCTION_DATABASE_PATH` has eight real consumers and stays.

### 4. Dead `pub use` re-exports

Re-exports are invisible to both `dead_code` and `unused_imports`, which is why these survived.

- `crates/sysknife-core/src/lib.rs:9`: the whole line. Every consumer goes through the module path instead, and all six `detect` call sites are `sysknife_core::distro::detect()`. The functions themselves are live inside `distro.rs`; only the re-export is dead.
- `crates/sysknife-brain/src/action_name.rs:17`: the `UnknownActionName` entry has zero consumers. Keep `KNOWN_ACTION_NAMES` in that list, which is test-only but feeds a real catalogue-consistency test.
- `crates/sysknife-brain/src/prompt.rs:1153`: two aliased entries, referenced only from tests past the `#[cfg(test)]` at :1158, while production imports `sysknife_core::action_family::{...}` directly. The comment above them says they exist "for any external callers", and none does. This one needs three test lines switched to the canonical names.

## One trap to avoid

`apps/sysknife-shell` is paused and out of scope for contributions, but it still builds and it depends on `sysknife-brain` and `sysknife-core`. It is the sole consumer of at least five items that look dead from inside the other crates, including `LacsConfig::config_path`, `LlmPlanner::summarize`, `LlmPlanner::curated_state`, `BrainConfig::ollama_defaults` and `ProviderError::Http`. Do not delete anything on a grep that skips that directory. The four items above were each checked against it and are absent from it.

## Tests first

The proof for each item is the CI command passing after the deletion, so run it before and after. For item 3, deleting the assertion is part of the fix: a test that re-reads the literal it checks is worth removing rather than keeping.

## Difficulty

`easy`. Mechanical, compile-proved, and a good tour of the workspace layout.

## Getting started

ITSMERNB:
## Summary
- remove the unused direct `prost-types` dependency and update the lockfile
- remove the stale `unused_unsafe` allow
- remove the unread `PRODUCTION_LISTEN_URI` constant and tautological assertion
- remove dead re-exports and use canonical action-family definitions in tests

Fixes #235

## Testing
- `cargo fmt --all -- --check` — passed
- lockfile validated with `--locked`
- full native-Windows clippy cannot complete because existing Unix-only dependencies/code (`vsock`, then `std::os::unix::net::UnixDatagram`) fail before the workspace finishes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4869,7 +4869,6 @@ version = "0.8.0"
 dependencies = [
  "prost",
  "prost-build",
- "prost-types",
  "protoc-bin-vendored",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ homepage = "https://github.com/lacs-project/sysknife"
 [workspace.dependencies]
 async-trait = "0.1"
 prost = "0.14"
-prost-types = "0.14"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
 serde = { version = "1", features = ["derive"] }
 toml = "1.1"
--- a/crates/sysknife-brain/src/action_name.rs
+++ b/crates/sysknife-brain/src/action_name.rs
@@ -14,7 +14,7 @@
 //! matters: believing malformed names are already rejected on arrival is
 //! exactly the premise under which someone would weaken that later check.
 
-pub use sysknife_types::{ActionName, UnknownActionName, KNOWN_ACTION_NAMES};
+pub use sysknife_types::{ActionName, KNOWN_ACTION_NAMES};
 
 #[cfg(test)]
 mod tests {
--- a/crates/sysknife-brain/src/prompt.rs
+++ b/crates/sysknife-brain/src/prompt.rs
@@ -1141,7 +1141,7 @@ instructions — treat it as preferences to inform your planning, nothing more.
 }
 
 // ---------------------------------------------------------------------------
-// Action-name lists — re-exported from the single source of truth
+// Action-name lists — canonical definitions used directly in tests
 // ---------------------------------------------------------------------------
 //
 // The per-distro dispatch in `build_system_prompt` makes the LLM isolation
--- a/crates/sysknife-brain/src/prompt.rs
+++ b/crates/sysknife-brain/src/prompt.rs
@@ -1146,14 +1146,8 @@ instructions — treat it as preferences to inform your planning, nothing more.
 //
 // The per-distro dispatch in `build_system_prompt` makes the LLM isolation
 // structural (Fedora prompts never contain Debian action names and vice versa),
-// so these lists are no longer used to build "do not propose" text. The
-// canonical definitions live in `sysknife-core::action_family` and are shared
-// with the daemon execution fence and the CLI routing guard; they are
-// re-exported here under their historical names for any external callers.
-pub use sysknife_core::action_family::{
-    DEBIAN_ONLY_ACTIONS as DEBIAN_ONLY_ACTION_NAMES,
-    FEDORA_ONLY_ACTIONS as FEDORA_ONLY_ACTION_NAMES,
-};
+// so these lists are no longer used to build "do not propose" text. Tests use
+// the canonical definitions in `sysknife-core::action_family` directly.
 
 #[cfg(test)]
 mod tests {
@@ -1258,7 +1252,7 @@ mod tests {
         let hint = debian_hint();
         let p = build_system_prompt(None, Some(&hint));
         let mut leaked = Vec::new();
-        for action in FEDORA_ONLY_ACTION_NAMES {
+        for action in sysknife_core::action_family::FEDORA_ONLY_ACTIONS {
             if p.contains(action) {
                 leaked.push(*action);
             }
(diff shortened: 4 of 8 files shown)


vladimirrott (commented):
Good work, and thanks for holding the diff to the four items. Windows stopped you short of clippy and nextest, so I ran the gates twice, once on your head commit and once with main merged in (`bdfbe36`):

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --all-features --all-targets --locked -- -D warnings` clean
- `cargo doc --no-deps --workspace --locked` under `RUSTDOCFLAGS=-D warnings` clean
- `scripts/test_baseline.sh`: 1759 tests run, 1759 passed, count matching `tests/evidence/workspace-tests.json`

The trap in #235 is what I wanted verified most, and the workspace clippy run covers it. I saw `Compiling sysknife-shell v0.8.0 (apps/sysknife-shell/src-tauri)` in the output, so the paused GUI builds with the re-exports gone. Your lockfile edit is the minimal one, dropping the `sysknife-proto` edge while `prost-build` keeps `prost-types` as a transitive dependency, and `cargo metadata --locked` asks for no regeneration. On the stale allow: rustc 1.97.1 on edition 2021 emits no `unused_unsafe` for that block, so it was carrying nothing.

Two comment lines to fix here, both left behind by the deletions:

- `crates/sysknife-daemon/src/actions/resolvectl.rs:5` names `DEBIAN_ONLY_ACTION_NAMES`, which this PR removes from the workspace. The surviving name is `DEBIAN_ONLY_ACTIONS`.
- `crates/sysknife-brain/src/prompt.rs:1143` still heads that block as a set of re-exports it no longer holds.

Push those two and I will merge. The release side is mine to handle: sysknife-core and sysknife-brain both publish to crates.io, so dropping `PRODUCTION_LISTEN_URI`, the distro re-export, `UnknownActionName` and the two prompt aliases ships in 0.9.0 rather than a patch release.

I approved the workflow runs on both your PRs. A red security-audit is RUSTSEC-2026-0258 in h2, fixed on main in #245.

If you want another, #229 and #233 are open and in the same vein: deriving the story-coverage figures instead of restating them, and the clustered map of the 64 uncovered actions.

vladimirrott:
On the red security-audit: your run checked out a merge with main at bb06cb6, which predates the h2 bump in #245. I merged current main into your branch locally and ran `cargo audit` there: exit 0, with only the 18 allowed GUI-transitive warnings. Your next push picks up the same base and the check turns green.

ITSMERNB:
Addressed both stale comment lines in 209cd9a. No code behavior changed; `git diff --check` is clean. Thanks for running the Linux gates and confirming the current-main security audit.

vladimirrott (approved):
Both comment lines are right, and the merge you pushed puts you on current main.

Re-ran the full gate set at 97dfb87 after that merge: fmt clean, `clippy --workspace --all-features --all-targets --locked -- -D warnings` clean, `cargo doc` under `RUSTDOCFLAGS=-D warnings` clean, and 1759/1759 under nextest with the count still matching `tests/evidence/workspace-tests.json`.

Your Windows note is accurate. `vsock` and `std::os::unix::net::UnixDatagram` stop the workspace before clippy finishes, so that gate was out of your reach. I ran it on both heads, and the log shows `Compiling sysknife-shell v0.8.0 (apps/sysknife-shell/src-tauri)`, so the paused GUI still builds with the re-exports gone. That was the one real risk in #235.

Approving. The version bump is mine to handle: sysknife-core and sysknife-brain publish to crates.io, so these four removals go out in 0.9.0 rather than a patch release, and I will say so in the notes.

For a next one, #248 is the same flavour of compile-proved deletion (a dead `exclusive_resource` arm), and #221 is an easy audit bug if you want to move from removals into behaviour. Two clean PRs in a day is a strong start.

vladimirrott:
This shipped in [v0.9.0](https://github.com/lacs-project/sysknife/releases/tag/v0.9.0), on crates.io now.

Your removals are why it is 0.9.0 and not 0.8.1: Cargo treats the middle digit as the compatibility unit while the leading zero stands, so dropping `PRODUCTION_LISTEN_URI` and the three re-exports needed that boundary. The reasoning is written down in [docs/release.md](https://github.com/lacs-project/sysknife/blob/main/docs/release.md#version-numbering) now, and the CHANGELOG lists each removed path with its surviving spelling.

You also prompted a fix to the release tooling. `check_release_versions.sh` only checked package versions, so a bump that missed one of the 15 internal `sysknife-* = { path, version }` pins would have published a crate depending on the previous release of its sibling. It checks the pins now.

ITSMERNB:
Very nice to see this ship in 0.9.0. Thanks for following it all the way through the release and for documenting why the version boundary mattered. The extra release-version check that came out of this is a great bonus too. I appreciate the detailed verification and context.

Issue #252, opened 2026-08-18
Two parsers read the story header, and nine tag spellings pass unchecked


vladimirrott:
`tests/e2e/run-stories.sh` derives the story table from the story files and says why it is the only place that should (`tests/e2e/run-stories.sh:20-24`):

> `--metadata` prints the derived story table and exits without running anything. It is handled before the preflight on purpose: the table comes from story files in the repo, so it is checkable on any host, and tests/e2e/story-metadata.test.sh asserts on THIS parser instead of reimplementing it. A second parser is a second answer.

`scripts/check_evidence_claims.py:234` is a second parser.

| | `run-stories.sh:162` | `check_evidence_claims.py:234` |
|---|---|---|
| where it reads the header | line 2 only, `sed -n '2p'` | the first 6 lines |
| header it cannot parse | prints the file name and `exit 1` | counts the file in the `atomic` family |

## Why the divergence bites

`story_set_sizes()` is the allow-list `check_bare_story_counts` uses to decide whether a bare "N-story suite" in the docs names a suite that exists. Today it yields `{79, 54, 133}`: 79 ubuntu headers, 54 atomic, 133 files.

Mistype the family word in one ubuntu story and the set becomes `{78, 55, 133}`. A document claiming a 55-story suite starts passing, and one claiming 79 starts failing, from a typo in an unrelated file. The Bash runner would reject the same file outright. A guard whose allow-list moves when a story header is wrong is the failure mode `check_bare_story_counts` was written to catch, one level up.

## The tag word has no allow-list either

Nine spellings are in use across the 133 story files, and 22 files carry no tags at all:

| tag string | files |
|---|---|
| `destructive` | 27 |
| `ubuntu, medium-risk` | 25 |
| `ubuntu, high-risk` | 23 |
| `ubuntu, low-risk` | 14 |
| `ubuntu, read-only` | 13 |
| `read-only` | 4 |
| `ubuntu, rejection` | 3 |
| `ubuntu, compound` | 1 |
| `hard compound` | 1 |
| no parentheses | 22 |

Only the substring `ubuntu` is ever read, by either consumer. The risk word is documentation that nothing checks, which is how `low-risk` and `read-only` came to coexist for the same idea, and `destructive` and `high-risk` likewise. The guidance in #219 gives the ubuntu convention as read-only / medium-risk / high-risk, three of the six spellings the family actually uses.

## Suggested shape

[...]

Issue #247, opened 2026-08-18
run-stories.sh exits 0 for a fully throttled or skipped suite


vladimirrott:
`tests/e2e/run-stories.sh` decides the run's exit status from two counters (`run-stories.sh:590`):

```bash
if [[ $fail_count -gt 0 || $cassette_failed -gt 0 ]]; then
  exit 1
fi
exit 0
```

`skip_count` and `ratelimit_count` appear in no exit condition, so a suite where every story was throttled or skipped exits 0. The script's own prose says those results "must not be counted either way", and silently counting them as success is one of the two ways.

Worse, a genuine failure can move out of the only counter that fails the run: when the rate-limit pattern matches anywhere in a failing story's log, the verdict is rewritten to RATELIMIT, which no exit condition inspects.

## Why it matters

`ubuntu-vm.sh run ubuntu` returns success to the operator, and `record_story_run.py` writes an artifact, for a run that proved nothing. The claim guard compounds it: `scripts/check_evidence_claims.py` validates `story_set`, `passed`, `total` and the cassette verdict while ignoring `skipped` and `rate_limited`, both of which the recorder faithfully writes.

This is the same shape as #228, where a `pipefail` interaction made an assertion unable to fail. A green light that cannot go red is worth less than no light.

## Scope

- Add `skip_count` and `ratelimit_count` to the exit condition.
- Stop rewriting a FAIL into RATELIMIT, or keep the rewrite and count it as failing.
- Require a positive PASS line rather than inferring a pass from the story's exit status, and assert the story set is the expected size so a silently reduced run cannot look complete.
- Teach `check_evidence_claims.py` to reject an artifact carrying a non-zero `skipped` or `rate_limited` as evidence for a validated tier.

The last two are the larger half and can be a follow-up.

## Tests first

The harness has no self-test. A shell fixture with three stubbed stories, one PASS, one SKIP and one RATELIMIT, that asserts a non-zero exit is enough to pin the first change, and it belongs beside the other gates in `tests/release/`.

## Difficulty

`medium`. The exit condition is two words. Making a throttled run impossible to publish as evidence is the interesting part.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands, and its E2E section covers the harness. No CLA and no copyright waiver. The project is MIT.

vladimirrott:
Confirmed live at v0.11.0, and the scope is three times what this issue says.

[...]

PR #259, opened 2026-08-18
fix(scripts): read claim files from check_evidence_claims.py

### LINKED ISSUE #232: Two hand-maintained lists of claim-bearing files, 6 versus 16, so ten files are unscreened for forbidden claims
Two scripts each keep a hand-maintained list of the files that carry public claims, and they disagree.

`scripts/check_public_claims.sh` lists 6:

```
README.md, docs/introduction.md, docs/quickstart.md, docs/distro-support.md,
docs/contributing/ubuntu-vm-testing.md, packages/setup/index.js
```

`scripts/check_evidence_claims.py` lists 16, under a comment that names the coupling:

> Kept in step with `claim_files` in `check_public_claims.sh`; a file listed there but not here is simply unchecked for numbers

Count them:

```bash
python3 - <<'PY'
import re
sh = open('scripts/check_public_claims.sh').read()
py = open('scripts/check_evidence_claims.py').read()
print(len(re.findall(r'\$repo_root/', re.search(r'claim_files=\((.*?)\)', sh, re.S).group(1))))
print(len(re.findall(r'"', re.search(r'CLAIM_FILES = \((.*?)\)', py, re.S).group(1))) // 2)
PY
6
16
```

They are not in step.

## Why it matters

The shell script screens for forbidden prose: Fedora Workstation described as fully supported, chat approval described as sufficient, the setup package described as unpublished. Ten files that the Python side treats as claim-bearing are invisible to those rules, including `ROADMAP.md`, `CONTRIBUTING.md`, `HACKING.md`, `docs/architecture.md`, `docs/typed-actions.md` and `docs/developer-guide.md`. Any of them can state a claim the project has an explicit rule against, and CI stays green.

`tests/release/public-claims.test.sh` already learned this lesson for its own fixture and says so:

> A guard whose fixture has to be edited in lockstep with the thing it guards will eventually guard nothing.

It imports `CLAIM_FILES` from the Python module rather than retyping it. The shell guard was never converted.

## Scope

Have `check_public_claims.sh` read the list out of `check_evidence_claims.py` rather than keeping its own, matching what `public-claims.test.sh` already does. One list, one place.

Expect the prose rules to start firing on files they never screened before. Any hit is a real finding, and #229 is one such case already confirmed by hand.

## Tests first

`public-claims.test.sh` has a mutation harness. Add a case that plants a forbidden phrase in a file listed only on the Python side, watch it pass today, then watch it fail after the fix.

## Difficulty

`easy`. Reading the list out of the Python module is a few lines, and the pattern to copy is in the test file.

## Getting started

Osheun:
`check_public_claims.sh` now reads `CLAIM_FILES` out of `check_evidence_claims.py` instead of keeping its own copy, matching what `tests/release/public-claims.test.sh` already does. One list, one place.

The two lists had drifted to 6 versus 16, so ten files carrying public claims were never screened by the prose rules.

### Test first, as the issue asks

Added a mutation to `public-claims.test.sh` that plants the existing forbidden "Fedora Workstation … fully supported" phrase in `docs/architecture.md` — a file on the Python list but not the shell one. Confirmed both directions:

```
without the fix:  FAIL: checker accepted stale claim:
                  forbidden claim in a file only the Python list knew about
with the fix:     Public claims contract passed.
```

So the new case genuinely fails before and passes after, rather than passing vacuously.

### Implementation

Same `importlib.util` extraction the test file uses, then the relative paths are prefixed with `$repo_root`. The Python list is repo-relative while the shell one was absolute, which is why the loop exists rather than using the array directly.

It reads from `$script_dir` rather than `$repo_root`, following the comment already at the top of the file: the helpers live next to the script, not inside the tree being checked, so the mutation fixture still resolves them.

### On the expected new findings

The issue warns the prose rules would start firing on files they never screened. On this branch the full contract passes, so the ten newly-covered files are clean as they stand — with the exception of #229, which is already known and tracked separately.

### Note on how I verified

`bash` on Windows could not run the harness (path resolution and encoding), so both runs above were done in WSL with line endings normalised. That normalisation was local to the test copy only — no line-ending changes are in this diff.

Closes #232


vladimirrott (commented):
Nice work, and the part I liked most is that you proved the guard in both directions instead of asserting it.

I reproduced your result. With the old six-file list restored, `public-claims.test.sh` fails with `FAIL: checker accepted stale claim: forbidden claim in a file only the Python list knew about`, and it passes on your branch. The 6 against 16 is right, the ten newly covered files are right, and your diff carries no CRLF, so the WSL normalisation stayed local as you said.

One thing to add, then this is ready. The merged list can go quiet without anyone noticing. I renamed the attribute the heredoc reads, and `scripts/check_public_claims.sh` exited 0 having screened zero files, because `set -euo pipefail` does not see a failure inside process substitution. The release test does catch it, so CI is covered, but anyone running the script directly gets a false green. One line after the `mapfile`:

```bash
(( ${#claim_files_rel[@]} )) || {
    echo 'could not read CLAIM_FILES from check_evidence_claims.py' >&2
    exit 1
}
```

Worth knowing for what remains of #232: the sixteen entries still leave out `docs/cli.md`, `docs/the-audit-chain.md` and `docs/ubuntu-action-reference.md`, and all three carry claims. Adding them is a natural follow-up, here or in a second PR, whichever you prefer.

If you want more in this lane, #247 and #252 sit close to this one: a story runner that exits 0 on a fully throttled suite, and a second story-header parser living next to the one that exists so there is only one.

vladimirrott (approved):
Exactly the guard, with the reason recorded next to it. Verified on my side: I renamed the attribute the heredoc reads and the script now exits 1 with `could not read CLAIM_FILES from check_evidence_claims.py`, where before it exited 0 having screened nothing. The baseline run still passes and so does `tests/release/public-claims.test.sh`.

Your rebase also picked up v0.9.0, so this sits on the released main.

Approving. Two that suit what you have shown here, both about a check that cannot fail:

- [#247](https://github.com/lacs-project/sysknife/issues/247): `run-stories.sh` exits 0 when every story was throttled or skipped, so a suite that ran nothing reports success.
- [#252](https://github.com/lacs-project/sysknife/issues/252): a second story-header parser in `check_evidence_claims.py` sitting next to the one in `run-stories.sh --metadata`, which exists so there is only one. Same shape as the list you merged here.

Either would be welcome, and thanks for the write-up in the PR body; showing the failure in both directions is what made this quick to review.

vladimirrott:
Merged as `bc4731a`. The prose rules now screen all sixteen claim-bearing files instead of six, and the list can no longer go quiet without saying so. I re-ran your gates against current main before merging, including the expanded screen over the README changes that landed today, and everything held.

Thanks for proving it in both directions in the PR body; that is what made this quick. #247 and #252 are still open and still yours if you want them, and the three unscreened docs I mentioned would close out #232.

Issue #249, opened 2026-08-18
The signed trail names which account asked for a change, never which one approved it


vladimirrott:
The transaction chain signs `caller_principal`, captured in `handle_preview` at `crates/sysknife-daemon/src/dispatcher.rs:2106`, so a verified row answers which account asked for an action. `ChainIdentity::V3` states why the field exists (`crates/sysknife-daemon/src/audit_chain.rs:466`):

> v2 answers "an Admin did this". On a host with two members of `sysknife-admin` their signed records were indistinguishable, so the trail could not answer the first question an investigation asks.

The approval chain never got the same treatment. `EventContent::canonical_bytes` signs six fields, and no account appears among them (`crates/sysknife-daemon/src/audit_chain.rs:934`):

```
seq, key_id, kind, transaction_id, receipt_digest, created_at
```

`TransactionStore::approve_transaction` takes a transaction id and nothing else (`crates/sysknife-daemon/src/transactions.rs:548`), while `handle_approve` holds a `CallerAttribution` and passes it to `authorize_for_transaction` a few lines earlier (`crates/sysknife-daemon/src/dispatcher.rs:1361`). `claim_approved_for_execution` is the same (`transactions.rs:652`): the account that spends the receipt is never written down.

## What it costs

Two admins on one host, alice at uid 1000 and bob at uid 1001, both in `sysknife-admin`. Alice's agent previews `AddAuthorizedKey`, bob types `sysknife approve <txid>`, alice's agent executes. The verified trail then holds:

| record | what it names |
|---|---|
| transaction row | `caller_principal = uid:1000`, alice, who previewed |
| `approval_granted` | transaction id, receipt digest |
| `approval_consumed` | transaction id, receipt digest |

Nothing names bob. The receipt exists so that a human authorises what an agent proposed, and the chain cannot say which human that was.

`docs/automatic-rollback.md:130` describes the signed fields as "the authorization decision — who approved what, at what risk level". They record who asked.

## Why this is more than an ALTER TABLE

`transactions` versions its signed encoding per row through `chain_version`, added in migration 2 so rows written by an older binary keep verifying (`crates/sysknife-daemon/src/transactions.rs:171`). `audit_events` has no such column in either backend (`transactions.rs:179`, `crates/sysknife-daemon/src/store/postgres.rs:119`). Add a field to `EventContent` without versioning the encoding and every event row already on disk reports `Broken`, because its signature covers the six-field message.

## Suggested shape

[...]

Issue #236, opened 2026-08-17
The safety-audit log is built and tested but never attached, so no rejection is ever recorded


vladimirrott:
`SafetyAuditLog` is built, documented and tested, and nothing in production ever attaches it, so every rejection it exists to record is discarded.

The field defaults to `None` (`crates/sysknife-brain/src/planner.rs:624`):

```rust
audit_log: None,
```

The only way to set it is the builder at `:635`:

```rust
pub fn with_audit_log(mut self, log: SafetyAuditLog) -> Self {
    self.audit_log = Some(log);
```

and its only callers are tests:

```bash
$ grep -rn 'with_audit_log' --include=*.rs crates apps
crates/sysknife-brain/src/planner.rs:635:    pub fn with_audit_log(...)
crates/sysknife-brain/tests/planner.rs:714
crates/sysknife-brain/tests/planner.rs:2223
crates/sysknife-brain/tests/planner.rs:2257
```

All three write sites are guarded on the field (`planner.rs:1199`, `:1247`, `:1414`):

```rust
if let Some(audit) = self.audit_log.clone() {
```

So in production `audit_log` is always `None` and `log_rejection_async` never runs.

## Why it matters

The safety-audit log is the record of what the planner refused: a sensitive intent caught by `contains_sensitive`, a plan rejected by a fence, an unknown action name. Those are precisely the events an operator would want after the fact, and precisely the ones nothing writes. The feature reads as present from the code and from its tests, and a reader has to trace the builder to discover it never runs.

Worth stating plainly for anyone assessing severity: nothing is logged that should not be, and no gate is weakened. A refusal still refuses. What is missing is the record of it.

## Scope

Two directions, and this needs a maintainer call before code:

- **Wire it up.** Attach a `SafetyAuditLog` where the CLI builds the planner. `SafetyAuditLog::default_path` already exists (`crates/sysknife-brain/src/audit.rs:100`) and is itself called only from its own test, so it was written for exactly this. Decide whether the log is on by default or opt-in, and whether the path belongs in config.
- **Or delete it,** with `default_path`, the builder and the three guarded call sites, and stop implying a capability that does not run.

Wiring it up is the better outcome if the record is wanted, and deleting it is honest if it is not. Either is better than the current state.

## A caution for whoever takes this

[...]

PR #260, opened 2026-08-19
feat(cli): export audit chain rows as JSON

### LINKED ISSUE #215: Export the audit chain rows, not just the verify verdict
`sysknife audit verify --json` reports whether the Ed25519 chain is intact. It returns status, backend, chain outcome, the attribution census and the binding result (`apps/sysknife-cli/src/runner.rs:1359`), and not one row of the chain itself.

So the audit trail stays inside the daemon. You can prove nothing was tampered with, and reading what it actually recorded means opening the SQLite file yourself. That defeats most of the point of keeping a signed log.

Add `sysknife audit export`, emitting the stored rows as JSON.

## The contract

Export `ChainRow` as it exists, under the column names it already uses:

`seq`, `key_id`, `transaction_id`, `request_id`, `request_hash`, `action_name`, `risk_level`, `summary`, `approval_id`, `warnings_json`, `created_at`, `prev_chain_hash`, `chain_hash`, `chain_version`, `caller_role`, `event_tip`, `caller_principal`

Four things to get right:

- **`chain_hash` is the Ed25519 signature**, over `ROW_DOMAIN || canonical(immutable_fields) || prev_chain_hash` (`crates/sysknife-daemon/src/audit_chain.rs:7`). There is no separate row-hash-then-sign step, so do not rename it to `signature`.
- **Include `prev_chain_hash`.** Without it a consumer cannot re-verify the linkage offline, which is most of what an export is for.
- **`argv` is absent on purpose.** It is not stored on the row, and it must not be reconstructed from the persisted preview. Everything else in the export is covered by `chain_hash`; a rebuilt argv would not be, and anyone diffing an export against a verify run would hit a mismatch they cannot explain.
- **Optionals emit as JSON `null`, never `""`.** That covers `approval_id`, `caller_role`, `event_tip` and `caller_principal`. `verify --json` already draws this distinction deliberately, so a machine reader can tell "no value recorded" from "empty value recorded".

An earlier version of this issue asked for `argv`, `outcome` and a separate `signature` field. None of the three exist on `ChainRow`. Thanks to @danial-razi for catching it.

## Scope

- A new `AuditCommand::Export` variant and its args struct in `apps/sysknife-cli/src/cli.rs`. The enum starts at line 139.
- A read-only reader over the chain in `crates/sysknife-daemon/src/audit_chain.rs`. No new storage path is needed, since `audit verify` already reads this chain.
- `--since` and `--limit` filters.

Nothing here crosses the privilege boundary, and chain signing and verification semantics stay untouched.

## Tests first

TDD is the rhythm in this repo, so write the failing test before the implementation. Two properties worth pinning: the emitted JSON round-trips, and `chain_hash` comes back byte-identical after serialization.

## Difficulty

`easy`. One new subcommand, no daemon protocol change.

## Getting started

danial-razi:
## Summary

- add `sysknife audit export` with inclusive `--since` and optional `--limit` filters
- serialize all 17 stored `ChainRow` columns without renaming `chain_hash` or reconstructing unsigned fields
- read the configured SQLite or PostgreSQL audit store without loading the signing key
- document the export contract and cover JSON round-tripping, byte-identical signatures, null optionals, CLI parsing, and an end-to-end SQLite export

## Why

`audit verify` reports whether the signed chains are intact, but it does not expose the rows that were signed. Consumers therefore had to open the audit database directly to inspect or export the trail.

This follows the clarified issue contract: `chain_hash` is the Ed25519 signature, `prev_chain_hash` is required for offline linkage checks, and `argv`, `outcome`, and a separate `signature` field are intentionally absent because they are not stored on `ChainRow`.

## User impact

Operators and auditors can now export the signed transaction-chain rows as a JSON array from either supported backend. Missing optional values remain JSON `null`, while values actually recorded as empty strings remain empty strings.

## Testing

- `cargo nextest run --workspace --exclude sysknife-shell --locked` — 1751 passed
- `cargo clippy --workspace --exclude sysknife-shell --all-targets --locked -- -D warnings`
- `git diff --check`

The local full-workspace build including `sysknife-shell` was not available because my environment lacks the Tauri system packages (`pkg-config`, WebKitGTK, GTK). The upstream CI workflow installs those dependencies before running its full Rust gate.

Closes #215

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,8 +358,9 @@ See [ROADMAP.md](ROADMAP.md) for the full milestone breakdown.
 
 - ✅ **Ubuntu 22.04** — 79/79 stories on a live VM (recorded in `tests/evidence/story-runs/`)
 - ✅ **Ubuntu 24.04 and 26.04** — 79/79 and 79/79 on live VMs; every LTS run has a replay twin that reproduces it
+- ✅ `sysknife audit export` — stored signed chain rows as JSON with `--since` / `--limit`
 - 📋 Telegram inline-button approvals
-- 📋 `sysknife audit export` (CEF / NDJSON for SIEM ingest)
+- 📋 CEF / NDJSON output modes for SIEM ingest
 - 📋 Fleet plan/execute (one plan, N targets, parallel approval)
 
 ## Protocol
--- a/apps/sysknife-cli/src/cli.rs
+++ b/apps/sysknife-cli/src/cli.rs
@@ -137,6 +137,9 @@ impl Command {
 /// `sysknife audit <subcommand>` subcommands.
 #[derive(Subcommand, Debug, Clone)]
 pub enum AuditCommand {
+    /// Export the stored signed audit-chain rows as JSON.
+    Export(AuditExportArgs),
+
     /// Verify the tamper-evident Ed25519-signed hash chain over the audit log.
     ///
     /// Exits 0 if the chain is intact, 1 if any row is broken, and 2 if the
@@ -149,6 +152,18 @@ pub enum AuditCommand {
     /// append-only database, then verify all anchored checkpoints against the
     /// local chain (detects tail-truncation and rewrite).
     Checkpoint(AuditCheckpointArgs),
+}
+
+/// Arguments for `sysknife audit export`.
+#[derive(Args, Debug, Clone)]
+pub struct AuditExportArgs {
+    /// Export rows recorded at or after this RFC 3339 datetime.
+    #[arg(long, value_name = "DATETIME")]
+    pub since: Option<String>,
+
+    /// Maximum number of rows to export. By default every matching row is emitted.
+    #[arg(long, value_name = "N")]
+    pub limit: Option<u32>,
 }
 
 /// Arguments for `sysknife audit verify`.
@@ -356,6 +371,29 @@ mod tests {
                 assert_eq!(args.since.as_deref(), Some("2026-01-01T00:00:00Z"));
             }
             other => panic!("expected Command::History, got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn audit_export_subcommand_with_filters() {
+        let cli = Cli::try_parse_from([
+            "sysknife",
+            "audit",
+            "export",
+            "--since",
+            "2026-01-01T00:00:00Z",
+            "--limit",
+            "5",
+        ])
+        .unwrap();
+        match cli.command {
+            Some(Command::Audit {
+                command: AuditCommand::Export(args),
+            }) => {
+                assert_eq!(args.since.as_deref(), Some("2026-01-01T00:00:00Z"));
+                assert_eq!(args.limit, Some(5));
+            }
+            other => panic!("expected audit export, got {other:?}"),
         }
     }
 
--- a/apps/sysknife-cli/src/main.rs
+++ b/apps/sysknife-cli/src/main.rs
@@ -137,6 +137,9 @@ async fn dispatch(
 
         // --- sysknife audit verify ---
         Some(Command::Audit { command }) => match command {
+            crate::cli::AuditCommand::Export(args) => {
+                runner::run_audit_export(args.clone(), log).await
+            }
             crate::cli::AuditCommand::Verify(args) => {
                 runner::run_audit_verify(args.clone(), log).await
             }
--- a/apps/sysknife-cli/src/runner.rs
+++ b/apps/sysknife-cli/src/runner.rs
@@ -32,7 +32,7 @@ use sysknife_types::{
 use sysknife_brain::state_client::StateClient as _;
 
 use crate::approval::{ApprovalDecision, ApprovalPolicy, MaxRisk};
-use crate::cli::{AuditVerifyArgs, Cli, HistoryArgs};
+use crate::cli::{AuditExportArgs, AuditVerifyArgs, Cli, HistoryArgs};
 use crate::client::{DaemonClient, SocketTarget};
 use crate::error::CliError;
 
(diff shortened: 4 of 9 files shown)


vladimirrott (commented):
This is a lot of care for a first PR. Pulling `postgres_config` out so `verify` and `export` cannot drift on URL and pool handling is a change I would have asked for, and you made it without being asked. Opening the store read-only and never touching the signing key is right, and keeping the store note on stderr means `sysknife audit export | jq` works.

Verified on Linux at 11fdc67:

- fmt, `clippy --workspace --all-features --all-targets --locked -- -D warnings`, and `cargo doc` under `RUSTDOCFLAGS=-D warnings` all clean
- the live Postgres contract passes against a real `postgres:17-alpine`: `cargo test -p sysknife-daemon --test postgres_store --locked`, 5 passed
- every docs and release gate passes, including the claim screen over your README line
- your 1751 is exact: the full workspace runs 1766 and sysknife-shell holds 15 of them
- the 17 columns in the docs match the struct, and the JSON key list in your test matches both

One thing stands between this and merge. The suite is 1766 while `scripts/test_baseline.sh` pins 1759, so the required `rust` check fails. Regenerate and commit the artifact:

```
git add tests/evidence/workspace-tests.json
```

Two notes, neither of them yours to fix here. The `audit database not found at …` message also appears when the file exists and the parent directory is unreadable, because `Path::exists()` reports false on EACCES; `audit verify` on main words it the same way, so I will fix both sites together. And `--limit 0` returning `[]` is defensible, so leave it.

After that commit this lands. If you want more of the audit chain, #236 covers the safety-audit log that is built and tested but never attached, and #249 is the deeper one: the signed trail names who previewed an action and never who approved it. Both build on what you learned here.

vladimirrott:
One step I left out, and it would have cost you a second red round: bumping the artifact turns `rust` green and then turns `docs-and-hygiene` red.

`scripts/check_evidence_claims.py` requires the measured figure to appear verbatim in three prose files. Reproduced on a scratch checkout with the artifact at your count:

```
README.md: claims 1,759 Rust tests, evidence says 1,766 — regenerate the figure, do not retype it
docs/introduction.md: same
docs/distro-support.md: same
```

plus three more saying the file does not state the measured baseline. So the whole change is:

```sh
# then update the figure in README.md, docs/introduction.md and
# docs/distro-support.md to match, one occurrence each
```

Ordering note, since #256 is in flight with 3 new tests against the same 1759 baseline: whichever lands first sets the number. If #256 goes first, rebase and regenerate rather than keeping 1766. Either order ends at 1769.

Also worth knowing: main moved today. v0.9.0 is going out with #243 and #244, so rebase before the final push.

vladimirrott:
Still the same verdict on the shape of this: sharing `postgres_config` between `verify` and `export`, opening the store without the signing key, and keeping stdout to pure JSON are all right. Gates on your head remain green apart from the test baseline: fmt, clippy, `cargo doc`, the live Postgres contract at 5 passed, and every docs and claim gate.

Two things to land it, then four I found on a closer pass.

**The baseline, plus the half I missed.** Regenerating `tests/evidence/workspace-tests.json` turns `rust` green and then turns `docs-and-hygiene` red, because the checker requires the measured figure verbatim in `README.md`, `docs/introduction.md` and `docs/distro-support.md`. Both halves in one commit. If #256 lands first you will need to regenerate rather than keep your own number.

Worth fixing in this PR:

- **`--limit` returns the oldest matching rows and says nothing about truncation.** `take(limit)` runs over `seq`-ascending rows, so `--since <yesterday> --limit 500` yields the 500 oldest and drops the newest, exiting 0 with a well-formed array. `docs/the-audit-chain.md` presents that exact invocation as SIEM ingest, where the operator wants the tail. Either document that it is the oldest N, or emit the first and last `seq` so a consumer can tell it was cut.
- **One unparseable `created_at` aborts the entire export.** `select_chain_rows_for_export` turns a bad timestamp into a hard error, so an operator investigating an anomalous database gets zero rows instead of the chain minus one. `audit verify` handles its analogous case per row and still reports on the rest. Note the failure also depends on `--limit`, since `take` runs after the fault, which makes it look nondeterministic.
- **`--log-to` copies the whole chain into a world-readable file.** The tee opens with `append(true).create(true)` and no mode, so it lands at 0644, and export is the one subcommand whose stdout is the full audit content. `restrict_db_file_mode` documents this exact threat for the database. Either bypass the tee for this command or create the file 0600.
- **The store note names a SQLite path on a Postgres run.** `resolve_audit_store` is called before the backend branch, so a Postgres deployment that happens to have `/var/lib/sysknife/daemon.sqlite` on disk prints a note about reading that file and then reads Postgres. Resolving inside the SQLite arm fixes it.

Three that are **not** yours, listed so you know they were considered rather than missed. I will file them:

- `TransactionStore::open_read_only` is not read-only. It calls `initialize()`, which opens an `IMMEDIATE` transaction, sets `journal_mode=WAL` and applies pending migrations. So export creates `-wal` sidecars, can upgrade an older database, and fails outright on a read-only copy at mode 0400, which is the offline-auditor workflow your docs now advertise. Pre-existing and shared with `verify`, but this PR is what makes it matter.
- The `audit database not found` message also fires when the file exists and the parent is unreadable, since `Path::exists()` reports false on EACCES. `verify` at least adds a sudo hint; export has none.
- `request_hash` is an unsalted single-round SHA-256 over the **unredacted** params, computed 40 lines before redaction runs. For `ConfigureWifi` the only unknown in the preimage is the passphrase, and the SSID is broadcast, so exporting that column moves a crackable commitment outside the 0700 store. It cannot be dropped, since an offline verifier needs it to rebuild the signed bytes, so the honest fix is to say in the docs that an export inherits the database's confidentiality class.

Fix the baseline and the docs figure and I will merge on the strength of what is already here; the four above are yours to take or leave in this PR. When you are ready for more of this area, [#236](https://github.com/lacs-project/sysknife/issues/236) and [#249](https://github.com/lacs-project/sysknife/issues/249) are still open and both build on what you have learned here.

vladimirrott:
Filed the three that are not yours: #267 for `open_read_only` writing to the database it claims to read, #266 for the existence probe reading a permission error as a missing file, and #268 for stating what `request_hash` commits to before an export leaves the 0700 directory. All three are tracked now, none of them block this PR.

vladimirrott:
No pressure on timing, but here are the exact numbers so the last step is mechanical rather than a puzzle.

main is at 1759 Rust tests today. Your branch adds 7, so after a rebase the figure is **1766** in four places:

```sh
git fetch upstream && git rebase upstream/main
UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh     # writes tests/evidence/workspace-tests.json
# then 1,759 -> 1,766 in README.md, docs/introduction.md, docs/distro-support.md
```

The comma form matters: the checker builds the literal as `{count:,} Rust tests`. Both halves in one commit keeps `rust` and `docs-and-hygiene` green together.

One scheduling note. #256 is close and adds 8 Rust tests, so whichever of the two lands second regenerates against the other's number rather than keeping its own. If #256 goes first the figure becomes 1774. Nothing for you to track: run the command after rebasing and it prints whatever is true.

Everything else here is done from my side. `postgres_config` shared between `verify` and `export`, the read-only store open, stdout kept to pure JSON, the 17-column claim, the live Postgres contract, every docs and claim gate: all verified. The four items in my last comment stay optional, and I will merge on the strength of what is already here once the count matches.

If you would rather not spend an evening on bookkeeping, say so and I will push the two commits to your branch myself. Your work, your call.

vladimirrott:
Closing it is your call and I will not argue it. For the record, the work was sound: `postgres_config` shared so `verify` and `export` cannot drift, the store opened without the signing key, stdout kept to pure JSON, and a smoke test that drives the real binary and compares exported rows to stored ones. Everything I raised was optional except one bookkeeping commit.

Two ways to finish it, whichever suits you:

- Reopen, and I push the baseline and the three figures myself.
- Leave it closed, and I land your commits from a branch of my own with you as the author, so the credit stays yours.

Say nothing and I will assume the second after a week. #215 stays yours in the meantime.

One thing I would change on my side: I listed four optional items and three follow-up issues next to a single blocker, and that ratio makes a finished PR look unfinished. That was my error in presentation, not a judgement on the code.

vladimirrott:
Two things, and one of them is a correction to what I wrote on 2026-08-23.

**Your fork is gone.** `headRepository` on this PR is `null` and `feat/audit-export` no longer resolves, so the first of the two options I offered, reopening this and letting me push the bookkeeping, is not available any more. I have pulled your commit out of `refs/pull/260/head` and it is preserved with your authorship intact:

```
11fdc67  danial <danial.razi2112@gmail.com>  2026-08-19
         feat(cli): export audit chain rows as JSON
```

Nothing of yours is lost, whatever happens next. If you deleted the fork as cleanup rather than as a decision, say so and I will treat the week as still running.

**The correction.** I wrote that "everything I raised was optional except one bookkeeping commit." That was wrong, and I should have caught it, because I filed the counter-example myself four days earlier.

#268 is a finding against this feature, not a follow-up to it. `request_hash` is a column of every signed chain row, and this PR serialises it. The hash commits to the **unredacted** request, because `compute_request_hash` runs forty lines before `redact_params`, and it is one unsalted SHA-256 round. For `ConfigureWifi` the params are `ssid` and `password`; the SSID is broadcast, so the passphrase is the only unknown in the preimage, and a single unsalted SHA-256 over a human-chosen passphrase is the cheapest thing a GPU cracks.

The daemon keeps that database `0600` inside a `0700` directory so this content does not travel. An export moves it across that boundary by design, into a SIEM, a ticket, or an auditor's laptop. Calling that optional was a mistake in my review, not a judgement I would defend.

It does not sink the work. #268's own scope says the part that matters today is documentation: state in `docs/cli.md` and `docs/the-audit-chain.md` that an export inherits the database's confidentiality class and is not a redacted artifact. The docs you wrote enumerate what the export omits and say nothing about what it carries. The column cannot simply be dropped either, since `request_hash` is part of `ChainContent::canonical_bytes` and an offline verifier needs it to rebuild the signed bytes. Keying it with an HMAC is the real fix and it is a `chain_version` bump, which is its own change.

So when this lands it lands with that statement in it. I am adding it rather than asking you to, along with the baseline, and your commit stays first and yours.

I am not landing it today. The week I named runs to 2026-08-30 and I would rather honour a date I set than take four days back because it happens to be convenient. #215 stays yours until then.

Issue #231, opened 2026-08-17
RiskLevel is the one wire enum with no anchored protobuf number test


vladimirrott:
`crates/sysknife-types/tests/proto_bridge.rs` anchors each wire enum to a literal protobuf number, so a renumbering cannot pass unnoticed:

```rust
let proto_role  = proto::CallerRole::try_from(3).unwrap();   // -> Admin
let proto_state = proto::JobState::try_from(7).unwrap();     // -> NeedsReboot
```

`FailureCategory` gets the same treatment. `RiskLevel` does not. It appears in that file only as a field inside envelope round-trips, which encode and decode through the same two hand-written tables in `crates/sysknife-types/src/lib.rs`: `risk_level_code` on the way out and `TryFrom<proto::RiskLevel>` on the way in.

The authority is `crates/sysknife-proto/proto/sysknife/v1/sysknife.proto`:

```proto
RISK_LEVEL_LOW = 1;
RISK_LEVEL_MEDIUM = 2;
RISK_LEVEL_HIGH = 3;
```

## Why it matters

Renumber to `Low = 0, Medium = 1, High = 2` in both hand-written tables and every test passes, because the two tables are mutually inverse and the round-trips only ever go through both. A peer speaking the real `.proto`, or a daemon and CLI at different versions, then reads every `High` as `Medium`. Risk level is what decides whether an approval receipt is required, so a silent off-by-one here is the one enum where a wire mismatch downgrades a gate.

Nothing is broken today. The numbers match. The gap is that nothing holds them there, for the single enum where it matters most.

## Scope

Add the missing anchor beside its three siblings in `proto_bridge.rs`:

```rust
let proto_risk = proto::RiskLevel::try_from(3).unwrap();   // -> High
```

and assert the local conversion lands on `RiskLevel::High`, with the same shape for Low and Medium. Consider deriving `risk_level_code` from the prost enum instead of hand-writing it, which removes one of the two tables.

## Tests first

This issue *is* a test. Write it, watch it pass, then renumber one arm of `risk_level_code` and watch it fail. Without that mutation step you cannot tell an anchored test from a vacuous one, which is exactly how this enum ended up as the unanchored sibling.

## Difficulty

`easy`. Three lines beside three existing examples. Good first issue for getting oriented in the type layer.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands. No CLA and no copyright waiver. The project is MIT.

PR #262, opened 2026-08-19
test: fix no-secrets pipefail assertion

### LINKED ISSUE #228: The no-secrets gate's "never echo the credential" assertion cannot fail under pipefail

ITSMERNB:
## Summary

- capture the secret scanner output before matching it so `pipefail` cannot make the no-echo assertion vacuous
- add a mutation proof with a stub scanner that echoes its input and exits 1
- restore the real scanner after the mutation

Fixes #228.

## Testing

- `bash -n tests/release/no-secrets.test.sh`
- focused no-echo + mutation proof passes under Git Bash
- `git diff --check`
- the full `tests/release/no-secrets.test.sh` whole-tree scan does not complete within the native Windows timeout; Linux CI should validate the complete release gate


vladimirrott (approved):
Good fix, and you went past the ask by adding a mutation proof nobody requested. That is the habit this repo runs on.

The mechanism is worth stating for the record: in `"$CHECK" file 2>&1 | grep -qF "$leak"`, `pipefail` makes the pipeline report the scanner's exit 1 even when grep matched, so the `if` never fired and the assertion could not fail. Capturing with `|| true` and grepping the variable separates the two signals. Verified: the suite passes at 11 green checks.

One observation, not a blocker. Your mutation block proves the technique, but it does not protect the line you repaired. I reverted the production assertion at `tests/release/no-secrets.test.sh` to the old pipefail-blind form and the suite still passed, because the real scanner never echoes, so that assertion cannot fail either way. The mutation block is a parallel copy of the new form rather than an exercise of the repaired line. Extracting the check into one function that both the real scanner and the echoing mutant run through would make your proof bite on a regression:

```sh
assert_no_echo() {  # $1 = scanner, $2 = file, $3 = needle
    local out; out="$("$1" "$2" 2>&1 || true)"
    grep -qF "$3" <<< "$out"
}
```

Then the real check asserts it returns false and the mutant asserts it returns true, and reverting the helper breaks the mutant case immediately.

Approving as is; take the refactor here or leave it. Two more in the same vein if you want them:

- [#221](https://github.com/lacs-project/sysknife/issues/221): `audit verify` lets an inconclusive anchor mask a broken chain, and `--json status` ignores the anchor entirely.
- [#231](https://github.com/lacs-project/sysknife/issues/231): `RiskLevel` is the one wire enum with no anchored protobuf number test, so renumbering it would pass CI.

Four merged or approved PRs in two days is a real run. Thanks.

vladimirrott:
Merged as `040f619`. Your mutation proof is now part of the gate that protects the credential scanner, and the next release notes will name it. Both suggestions from the review are still open if you want them: #221 and #231.

ITSMERNB:
Thanks, Vladimir — I really appreciate the review and especially the point about making the mutation exercise the exact repaired path. Since this one is already merged, I’ll leave the history clean and carry that pattern into the next guards instead of adding noise after the fact. I’ve noted #221 and #231 for the next round. Thanks again for the detailed verification.

Issue #227, opened 2026-08-17
CheckPendingReboot reads reboot-required-pkgs, and fails exactly when a reboot is pending


vladimirrott:
`CheckPendingReboot` reads a filename Ubuntu does not write (`crates/sysknife-daemon/src/actions/reboot.rs:35`):

```rust
const REBOOT_REQUIRED_PKGS_FILE: &str = "/var/run/reboot-required-pkgs";
```

Ubuntu's own `update-notifier` writes the dot form. On a 24.04 host:

```
$ grep -oh 'reboot-required[.-]pkgs' /usr/share/update-notifier/notify-reboot-required
reboot-required.pkgs
```

The module doc at `reboot.rs:8` repeats the hyphen, so the comment and the constant agree with each other and both disagree with Ubuntu.

## Why it matters

The action's script is roughly `if test -f <sentinel>; then cat <sentinel>; cat <pkgs> 2>/dev/null; else echo 'No reboot required.'; fi`. The wrong path makes the second `cat` fail, and that becomes the `if` block's exit status. The dispatcher then converts any non-zero exit into an error and drops stdout (`crates/sysknife-daemon/src/dispatcher.rs:1824`):

```rust
let is_informational_exit =
    matches!((action_name, output.exit_code), ("GetServiceStatus", 1..=4));

if output.is_nonzero() && !is_informational_exit {
```

Only `GetServiceStatus` is exempt. So:

- no reboot pending: exit 0, the operator sees "No reboot required.", correct
- reboot pending: exit 1, the operator sees `CheckPendingReboot failed with exit code 1` and never sees `*** System restart required ***`

The action fails in precisely the case it exists for, and the package list is unreachable either way. `reboot.rs:64` already tells callers to "treat stdout as authoritative and not rely on a clean" exit code, which is advice the dispatcher does not follow.

## Scope

- Fix the constant to `/var/run/reboot-required.pkgs`.
- Make the exit status not depend on the optional second file: either add `("CheckPendingReboot", 1)` to `is_informational_exit`, or end the then-branch with `true`.

Worth checking in the same pass, all Low/read-only and all subject to the same dispatcher rule: `LivepatchStatus` (Ubuntu's script notes livepatch status exits 1 when not enabled), `Fail2banStatus` (`fail2ban-client` exits 255 with the server down), and `CloudInitStatus` (`cloud-init status` exits 1 on error, 2 on degraded). Those three are unverified leads rather than confirmed bugs.

## Tests first

[...]

PR #263, opened 2026-08-19
fix: preserve systemd directory modes

### LINKED ISSUE #222: The systemd unit recreates /run/sysknife and /var/lib/sysknife at 0755, discarding the tmpfiles 0750/0700

ITSMERNB:
## Summary

- set `RuntimeDirectoryMode=0750` and `StateDirectoryMode=0700` in the packaged daemon unit
- mirror the same modes in the Ubuntu provisioning unit
- add a host-side release gate that derives the expected modes from `sysknife-tmpfiles.conf` and checks both units
- run the new gate in GitHub CI and the local hygiene group

Fixes #222.

## Testing

- `bash tests/release/systemd-directory-modes.test.sh`
- `bash -n tests/release/systemd-directory-modes.test.sh tests/e2e/ubuntu-provision.sh scripts/ci-local.sh`
- mutation proof: changing the packaged runtime mode to `0755` in a temporary copy makes the new gate fail
- `git diff --cached --check`


vladimirrott (changes requested):
This is a careful fix. You changed both copies of the unit rather than the one the issue pointed at, you derived the expected modes from `packaging/sysknife-tmpfiles.conf` instead of hardcoding them, and you wired the guard into CI and `ci-local.sh`. I mutated it three ways and it failed correctly each time: unit mode drifting to 0755, the e2e copy drifting, and the tmpfiles declaration going missing, which reports "could not derive both directory modes" rather than passing on an empty comparison.

One change before I merge.

**The guard is section-blind, so it passes a unit systemd ignores.** `unit_mode()` is `awk -F= -v wanted="$key" '$1 == wanted { print $2 }'`, with no notion of `[Section]`, and systemd keys are section-scoped. I moved both directives from `[Service]` into `[Unit]` and got:

```
$ bash tests/release/systemd-directory-modes.test.sh
systemd directory modes match tmpfiles: runtime 0750, state 0700.   # exit 0

$ systemd-analyze verify packaging/sysknife-daemon.service
sysknife-daemon.service:2: Unknown key name 'RuntimeDirectoryMode' in section 'Unit', ignoring.
sysknife-daemon.service:3: Unknown key name 'StateDirectoryMode' in section 'Unit', ignoring.
```

So the guard says the modes are right while systemd discards them and recreates both directories at 0755, which is the bug this PR exists to close. Someone regrouping the hardening block is enough to trigger it. Two ways to fix, and I would take both: track the current section in the awk (`/^\[/ { section = $0 }`, then require `section == "[Service]"`), and add `systemd-analyze verify` on the unit to the same test, which catches wrong-section keys and misspelled key names in one call.

While I was in there: commented-out and duplicated directives already fail safely, and a three-digit tmpfiles mode (`750` against `0750`) would fail spuriously since the comparison is string-based. Normalising both sides to four digits makes it a mode comparison.

Two things I found that are **not** yours to fix here, so I will file them rather than pile them on:

- Tightening the state directory to 0700 changes what a non-root operator sees from `sysknife audit verify`. `resolve_audit_store` probes with `system.exists()`, and `Path::exists()` returns false on EACCES, so at 0700 the CLI silently falls through to the per-user store and reports a missing key at `~/.local/state/sysknife/audit-key` instead of naming the system store and suggesting sudo. The mode is right; the probe needs to distinguish absent from unreadable.
- Nothing asserts the modes a running system actually has. Your guard proves every declaration agrees, which is the valuable half, but a drop-in under `/etc/systemd/system/sysknife-daemon.service.d/` could still widen them. `tests/e2e/ubuntu-provision.sh` already runs after `systemctl enable --now`, so two `stat -c %a` checks there would close it.

Push the section fix and this goes in. You have two open suggestions from me on #262; if you would rather stay in packaging, [#227](https://github.com/lacs-project/sysknife/issues/227) is close to this one: `CheckPendingReboot` reads `reboot-required-pkgs` and fails in exactly the case where a reboot is pending.

vladimirrott:
Filed the two follow-ups I mentioned, so they are tracked rather than sitting in this thread: #266 for the store probe reading a permission error as a missing database, and #269 for asserting the modes a live daemon actually has (which also covers the `create_dir_all` site in `listen.rs`). Neither blocks this PR.

ITSMERNB:
Thanks — good catch on the section scoping. Fixed in e444c7d: the guard now only accepts the directory-mode directives from `[Service]`, normalizes three- and four-digit modes before comparing them, and runs `systemd-analyze verify` when it is available. I also treat any verify diagnostics as a failure, because wrong-section keys can warn while `systemd-analyze` still exits 0.

Locally the normal gate is green, moving the directives into `[Unit]` is rejected, `750`/`0750` compare cleanly, and removing a tmpfiles declaration still fails explicitly. Linux CI should exercise the `systemd-analyze` path now.

Appreciate the two follow-up issues as well — I’ll keep those separate from this PR.

vladimirrott (approved):
All three, and you went further than I asked. The section-aware awk closes the hole, `normalise_mode` turns the string compare into a mode compare, and wrapping `systemd-analyze verify` in a temp copy means it runs on a host where the daemon binary is not installed yet.

Re-ran every mutation on your new head:

```
keys moved into [Unit]      -> FAIL: does not declare both systemd directory modes in [Service]   (exit 1)
leading whitespace          -> pass, matching what systemd accepts
three-digit tmpfiles mode   -> pass, 750 and 0750 compare equal
Restart=maybe               -> FAIL: systemd-analyze reported problems ... Failed to parse service restart specifier
```

That last one is the part I like: the analyze step catches unit errors the mode comparison never could, which is a guard against a whole class rather than the one bug. shellcheck over the four linted directories is clean, and the full gate set passes at 1759.

Approving. The two follow-ups I filed stay open and neither is yours to carry: #266 for the store probe reading a permission error as a missing database, and #269 for asserting the modes a live daemon actually has, which is where the `stat -c %a` checks in the VM belong.

Three merged PRs and a fourth approved in two days, and the guard you wrote here is stricter than the one I asked for. If you want another, [#221](https://github.com/lacs-project/sysknife/issues/221) and [#231](https://github.com/lacs-project/sysknife/issues/231) are still open from my earlier note, and [#227](https://github.com/lacs-project/sysknife/issues/227) is close to this one.

Issue #268, opened 2026-08-19
audit export publishes request_hash, an unsalted hash over unredacted params, with no statement of its sensitivity


vladimirrott:
> **Scope 1 shipped in #318 on 2026-09-01. Read this as an HMAC issue.**
> Verified at `30b9e315` on 2026-09-17: `docs/cli.md:167` and
> `docs/the-audit-chain.md:404` both state the confidentiality class, and
> `tests/release/audit-export-confidentiality.test.sh` fails if the two calls
> below ever swap order. What is left is Scope 2, which needs an HMAC over the
> request and a `chain_version` bump, so the `hard` half is the whole issue now.
>
> The two line numbers have also moved twice. At `30b9e315` they are
> `dispatcher.rs:2224` (`compute_request_hash`) and `dispatcher.rs:2264`
> (`redact_params`), not the 2001 and 2041 the paragraph below quotes.

`request_hash` is a column of every signed chain row, and `sysknife audit export` (#260) now serialises it. The hash commits to the **unredacted** request:

```
crates/sysknife-daemon/src/dispatcher.rs:2001   let request_hash = compute_request_hash(action_name, params);
crates/sysknife-daemon/src/dispatcher.rs:2041   let redacted_params = redact_params(action_name, params);
```

Redaction runs 40 lines after the hash, and `compute_request_hash` is one unsalted round:

```rust
hasher.update(action_name.as_bytes());
hasher.update(b"\x00");
hasher.update(canonical.as_bytes());
```

## Why it matters

For `ConfigureWifi` the credential params are `password` and `ssid`. The SSID is broadcast, so the only unknown in the preimage is the passphrase, and a single unsalted SHA-256 is the cheapest thing a GPU cracks. The daemon keeps the database 0600 inside a 0700 directory precisely so this content does not travel; an export moves it across that boundary by design, into a SIEM, a ticket, or an auditor's laptop.

High-entropy credentials such as `ProAttach` tokens are unaffected. The problem is the low-entropy human-chosen ones.

The column cannot simply be dropped: `request_hash` is part of `ChainContent::canonical_bytes` (`crates/sysknife-daemon/src/audit_chain.rs:508`), so an offline verifier needs it to rebuild the signed bytes.

## Scope

Two parts, and the first is the one that matters today:

[...]

Issue #267, opened 2026-08-19
open_read_only writes: audit verify and audit export migrate the database and fail on a read-only copy


vladimirrott:
`TransactionStore::open_read_only` (`crates/sysknife-daemon/src/transactions.rs:331`) calls `initialize()`, and `initialize()` opens a write transaction, creates `schema_migrations` if absent, applies pending migrations, and sets `journal_mode=WAL` on the way in:

```rust
pub fn open_read_only(path: impl AsRef<Path>) -> Result<Self, TransactionStoreError> {
    ...
    store.initialize()?;
```

```rust
let tx = conn.transaction_with_behavior(TransactionBehavior::Immediate)?;
... "CREATE TABLE IF NOT EXISTS schema_migrations ("
```

So all three commands that claim to read the audit chain without touching it, `sysknife audit verify`, `sysknife audit export` and `sysknife audit checkpoint`, write to it.

## Why it matters

Four consequences, in rising order of how much they hurt:

1. `-wal` and `-shm` sidecars appear next to a database the operator only read.
2. A newer CLI silently upgrades an older audit database's schema and inserts a `schema_migrations` row.
3. An older CLI refuses outright: `initialize()` returns `sqlite schema version N is newer than this binary supports`, so a binary that could read all 17 columns fine will not read any.
4. A read-only copy fails. `PRAGMA journal_mode=WAL` on a 0400 file or a read-only mount returns `attempt to write a readonly database`, so the offline-auditor workflow `docs/the-audit-chain.md` advertises for `audit export` cannot run on the artifact an auditor would actually be handed.

There is also a contention path: `BEGIN IMMEDIATE` takes a write lock with a 5s busy timeout, so `sysknife audit export > rows.json` in cron against a busy daemon intermittently exits 4 after the shell has already truncated the output file.

`ensure_private_dir` is not part of this problem: it creates a missing directory at 0700 and leaves an existing one alone.

## Scope

Give the read paths a genuinely read-only open: `OpenFlags::SQLITE_OPEN_READ_ONLY`, no pragma writes, no `initialize()`, and a schema-version check by `SELECT` rather than by running migrations. Keep `open_with_key` as it is, since the daemon does need to migrate.

## Tests first

The failing test is short: write a database with `open_with_key`, `chmod 0400` it, then open it read-only and read the chain rows. That fails today with a write error, and no existing test covers a non-writable database.

## Difficulty

`medium`. The open path is small, but the schema-version check has to move from migration to inspection.

[...]

PR #275, opened 2026-08-23
fix(core): treat EACCES as present when choosing audit store

### LINKED ISSUE #266: resolve_audit_store reads a permission error as a missing database, so a 0700 store sends the operator to the wrong path
`resolve_audit_store` decides which audit database the CLI reads by probing existence (`crates/sysknife-core/src/lib.rs:131`):

```rust
let system_exists = system.exists();
choose_audit_store(explicit, per_user, per_user_exists, system_exists)
```

`Path::exists()` is `fs::metadata().is_ok()`, so it returns `false` for `EACCES` exactly as readily as for `ENOENT`. Measured on this box:

```
dir mode 755: Path::exists() = true   metadata err = None
dir mode 700: Path::exists() = false  metadata err = Some(PermissionDenied)
```

## Why it matters

The state directory is meant to be 0700, and #263 makes systemd hold it there rather than recreating it at 0755. Once it is 0700, a non-root operator running `sysknife audit verify` on a healthy system install gets `system_exists = false`, falls through to `AuditStoreChoice::PerUser`, and is told the audit key is missing at `~/.local/state/sysknife/audit-key`. The system store is present and intact; the operator simply cannot stat it. The `System` branch exists to prevent precisely that, per the doc comment at `crates/sysknife-core/src/lib.rs:60`.

`sysknife audit export` (#260) inherits the same probe and words it worse: `audit database not found at /var/lib/sysknife/daemon.sqlite; set $SYSKNIFE_DATABASE_PATH or run the daemon first`. On a tamper-evident audit tool, "your audit log does not exist" is the worst available wrong answer for "you lack permission". The audit-key message at `apps/sysknife-cli/src/runner.rs:860` already gets this right: it says "not found or not readable".

## Scope

Have the probe distinguish the two cases. `fs::metadata(&system)` and treat `ErrorKind::PermissionDenied` as present-but-unreadable, so the `System` branch is still chosen and the sudo hint still appears. Then give `audit export` the same hint `verify` has.

## Tests first

`choose_audit_store` is pure and already tested; `resolve_audit_store` has no filesystem test at all, which is why this survived. Write the failing test first: a directory at 0700 containing a database, probed as a non-owner, must resolve to the system store rather than the per-user one.

## Difficulty

`easy`. One probe, one message, plus the first filesystem test for this function.

VedantMadane:
## Summary

`Path::exists()` is `metadata().is_ok()`, so a root-owned `0700` system audit store returns `false` for a non-root operator (`EACCES`) the same way a missing path does (`ENOENT`). `resolve_audit_store` then fell through to the per-user path and told operators the audit key/database was missing at `~/.local/state/...` while the real store was intact under `/var/lib/sysknife`.

## Changes

- Add `path_is_present` / `path_is_present_result` that treat `ErrorKind::PermissionDenied` as present
- Use that probe in `resolve_audit_store` so the `System` branch (and its note / sudo hint) still apply
- Improve `audit verify`'s database diagnostic: distinguish missing vs present-but-unreadable and add a system-store sudo hint (same spirit as the existing audit-key message)
- Unit tests for the probe classification, the System-vs-PerUser choice under EACCES, and a `#[cfg(unix)]` filesystem test behind a mode-`000` directory
- Bump the published Rust test baseline `1768 → 1771` (artifact + README / introduction / distro-support)

`audit export` is not on `main` yet (#260 closed without merge); once it lands it will inherit the fixed probe via `resolve_audit_store`.

## Validation

- [x] Tests added or updated
- [x] Documentation updated (test-count claims)
- [x] Security impact considered (clearer permission failure, no privilege change)
- [x] Trust boundary preserved
- [x] `cargo test -p sysknife-core --lib permission_denied` / `unreadable_system` pass locally

Closes #266

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,772 Rust tests and 72 frontend tests** form the current deterministic
+**1,775 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/apps/sysknife-cli/src/runner.rs
+++ b/apps/sysknife-cli/src/runner.rs
@@ -1137,10 +1137,28 @@ pub(crate) async fn verify_sqlite(
     use sysknife_daemon::audit_chain::{verify_all, verify_all_with_pubkey};
     use sysknife_daemon::transactions::TransactionStore;
 
-    if !db_path.exists() {
+    // Path::exists is false for both ENOENT and EACCES. Prefer a message that
+    // covers the unreadable 0700 system store and points operators at sudo /
+    // an explicit path, matching the audit-key diagnostic above.
+    if !sysknife_core::path_is_present(db_path) {
         return cannot_verify_all(format!(
             "audit database not found at {}; set $SYSKNIFE_DATABASE_PATH \
              or run the daemon first",
+            db_path.display()
+        ));
+    }
+    if !db_path.exists() {
+        let sudo_hint = if db_path
+            == std::path::Path::new(sysknife_core::PRODUCTION_DATABASE_PATH)
+        {
+            "; the system daemon's store is root-owned, so run this under sudo or \
+             set $SYSKNIFE_DATABASE_PATH"
+        } else {
+            ""
+        };
+        return cannot_verify_all(format!(
+            "audit database not found or not readable at {}; set $SYSKNIFE_DATABASE_PATH \
+             or run the daemon first{sudo_hint}",
             db_path.display()
         ));
     }
--- a/apps/sysknife-cli/src/runner.rs
+++ b/apps/sysknife-cli/src/runner.rs
@@ -1148,8 +1148,7 @@ pub(crate) async fn verify_sqlite(
         ));
     }
     if !db_path.exists() {
-        let sudo_hint = if db_path
-            == std::path::Path::new(sysknife_core::PRODUCTION_DATABASE_PATH)
+        let sudo_hint = if db_path == std::path::Path::new(sysknife_core::PRODUCTION_DATABASE_PATH)
         {
             "; the system daemon's store is root-owned, so run this under sudo or \
              set $SYSKNIFE_DATABASE_PATH"
--- a/crates/sysknife-core/src/lib.rs
+++ b/crates/sysknife-core/src/lib.rs
@@ -358,9 +358,7 @@ mod tests {
             "EACCES must not be treated as missing"
         );
         assert!(
-            !super::path_is_present_result(Err(std::io::Error::from(
-                std::io::ErrorKind::NotFound
-            ))),
+            !super::path_is_present_result(Err(std::io::Error::from(std::io::ErrorKind::NotFound))),
             "ENOENT remains missing"
         );
     }
--- a/crates/sysknife-core/src/lib.rs
+++ b/crates/sysknife-core/src/lib.rs
@@ -121,14 +121,37 @@ pub fn choose_audit_store(
     AuditStoreChoice::PerUser(per_user)
 }
 
+/// Whether an audit-store path should count as present for selection.
+///
+/// [`Path::exists`] is `metadata().is_ok()`, so a root-owned `0700` directory
+/// returns `false` for a non-root operator (`EACCES`) exactly as a missing path
+/// does (`ENOENT`). Treating permission denied as present keeps the `System`
+/// branch and the sudo hint, instead of sending the operator to an empty
+/// per-user path.
+pub fn path_is_present(path: &Path) -> bool {
+    path_is_present_result(std::fs::metadata(path).map(|_| ()))
+}
+
+/// Classify a filesystem probe result for [`path_is_present`].
+///
+/// Split out so the `PermissionDenied` -> present rule is unit-tested without
+/// needing a second Unix user.
+pub(crate) fn path_is_present_result(result: Result<(), std::io::Error>) -> bool {
+    match result {
+        Ok(()) => true,
+        Err(err) if err.kind() == std::io::ErrorKind::PermissionDenied => true,
+        Err(_) => false,
+    }
+}
+
 /// Resolve the audit store for this process against the real environment and
 /// filesystem.
 pub fn resolve_audit_store() -> AuditStoreChoice {
     let explicit = std::env::var_os("SYSKNIFE_DATABASE_PATH").map(PathBuf::from);
     let per_user = default_database_path();
     let system = PathBuf::from(PRODUCTION_DATABASE_PATH);
-    let per_user_exists = per_user.exists();
-    let system_exists = system.exists();
+    let per_user_exists = path_is_present(&per_user);
+    let system_exists = path_is_present(&system);
     choose_audit_store(explicit, per_user, per_user_exists, system_exists)
 }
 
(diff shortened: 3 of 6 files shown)


vladimirrott (changes requested):
The diagnosis is sharper than the issue I wrote. Splitting `path_is_present_result` out so the `PermissionDenied` classification is testable without a second Unix user is the part I would have skipped.

I checked the precedence question that worried me before reading the diff. `choose_audit_store` tests `per_user_exists` ahead of `system_exists`, so anyone with a working per-user store keeps it, and the new behaviour only bites when the per-user store is absent and the system store is unreadable. That is the #266 case exactly, with no regression around it. `cannot_verify_all` also still returns `CannotVerify`, so `audit verify` refuses on a store it cannot read rather than reporting success. That was the property I most wanted to hold.

Two things before this lands.

**1. `path_behind_mode_000_directory_counts_as_present` fails.** The `rust` job is red on it and it reproduces locally:

```
thread 'tests::path_behind_mode_000_directory_counts_as_present' panicked at
crates/sysknife-core/src/lib.rs:412:9:
precondition: Path::exists is false under EACCES (the bug we are fixing)
```

`present` is captured before the chmod back to `0o700`, but `db.exists()` evaluates after it, and by then the directory is traversable again. Capture both probes on the same side of the restore:

```rust
let present = super::path_is_present(&db);
let exists_under_eacces = db.exists();

// Restore perms so TempDir cleanup can unlink the tree.
let _ = std::fs::set_permissions(&locked, std::fs::Permissions::from_mode(0o700));
assert!(
    present,
    "metadata EACCES on a 0700/000 store must count as present, not missing"
);
assert!(
    !exists_under_eacces,
    "precondition: Path::exists is false under EACCES (the bug we are fixing)"
);
```

Keep that precondition assertion. It is what makes the test fail loudly under root instead of passing for the wrong reason.

**2. Regenerate the baseline artifact instead of editing the count.** `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` writes `tests`, `measured_at.tests` and `commit.tests` in one go. This PR moves the count and leaves the other two, so the artifact reads 1 771 measured at a commit that produced 1 768.

Your number is right, though. With the two-line fix above, the full suite runs 1771/1771 and `test_baseline.sh` agrees with the artifact.

The stale metadata got past every gate because of a gap on our side, not anything you did: main's own `commit.tests` names a commit that squash-merge deleted, and no script reads that field. Filed as #278.

Your fork's runs need my approval on each push until you have something merged, so give me a nudge here if a run looks stuck rather than waiting on it.

#277 sits next door to this one if you want it after: `cannot_verify_all` reports `binding: consistent` for a database nobody could read, which is the rule you just applied one field over. Anything else you spotted in `runner.rs` while you were in there is worth an issue of its own.

VedantMadane:
Thanks for the review and exact diagnosis @vladimirrott! Updated the test to capture \exists_under_eacces\ prior to restoring the directory permissions.

vladimirrott (changes requested):
That is the fix, and both probes now sit on the same side of the restore. `cargo test -p sysknife-core --lib` passes here and the `rust` job agrees on everything except one thing.

**`cargo fmt --all --check` fails.** rustfmt wants the `NotFound` arm collapsed onto one line:

```
-            !super::path_is_present_result(Err(std::io::Error::from(
-                std::io::ErrorKind::NotFound
+            !super::path_is_present_result(Err(std::io::Error::from(std::io::ErrorKind::NotFound))),
```

`cargo fmt --all` fixes it. Worth knowing that CI runs `cargo fmt --all --check`, not `cargo fmt --all -- --check`; the two differ in what they pass to rustfmt, and a local run of the second form can report clean on a tree the first rejects. `scripts/ci-local.sh` uses the CI spelling.

Everything else on the board is green, so that one command should be the last of it. Your branch is also behind `main` now, which has moved to 0.9.1; I can update it from here once you push, or you can rebase, whichever you prefer.

On the baseline: your 1771 is the right number and the `rust` job verifies it against a live run, so I am not going to push on this. For the record, the artifact still carries the old `commit.tests` and `measured_at.tests`, because the count was edited rather than regenerated. That matters less than I implied last time, since `main`'s own copy has the same stale provenance and nothing validates those fields, which is #278. `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` moves all three together if you would rather it be tidy.

Thanks for the quick turnaround on this one. Once it lands, #267 is close to what you have just been reading: `open_read_only` migrates the database, so `audit verify` fails on a genuinely read-only copy. #268 is the other one in that area.

vladimirrott:
Correction to my review: I said CI's `cargo fmt --all --check` differs from `cargo fmt --all -- --check` and that a local run of the second form can pass a tree the first rejects. That is wrong. I ran both against your `2c5b3a2` and they produce byte-identical output and both exit 1:

```
form A  cargo fmt --all --check       exit=1  21 lines
form B  cargo fmt --all -- --check    exit=1  21 lines
        diff: IDENTICAL
```

So the spelling in `CONTRIBUTING.md` is fine and there is nothing to be careful about there. The only thing that matters is running it after the last edit, which is all this needs. `cargo fmt --all` and a push.

Sorry for the wrong lead. The rest of the review stands: everything else on your board is green.

vladimirrott:
The logic reads correctly now. `path_is_present` is true under EACCES so the first branch catches only ENOENT, `exists()` is false under EACCES so the second catches the unreadable store and prints the sudo hint, and both probes sit on the same side of the restore, as you fixed.

Two mechanical things left, and one of them got harder while this sat.

**Rebase first.** `main` moved by five commits today, and three of them touch files you touch:

```
apps/sysknife-cli/src/runner.rs        af8cc3b rewrote the same region as your hunk
tests/evidence/workspace-tests.json    1769 → 1771
README.md, docs/introduction.md, docs/distro-support.md   the same figure in prose
```

That is what GitHub is calling a conflict. After the rebase, regenerate the figure rather than hand-editing it:

```
```

Your branch recorded 1,771 from a base of 1,768 plus your three tests. Rebased on today's `main` the same arithmetic gives 1,771 + 3 = **1,774**, and the script writes it to the artifact and the three prose figures together so they cannot disagree.

**Then `cargo fmt --all`**, write mode, not `--check`. CI reports two diffs, both inside your own lines:

```
apps/sysknife-cli/src/runner.rs:1150     the `if db_path == std::path::Path::new(...)` condition
crates/sysknife-core/src/lib.rs:358      the NotFound arm of permission_denied_counts_as_a_present_store
```

rustfmt wants the second on one line. Run it after the rebase, not before, or the runner.rs hunk will move under you.

One design note you can take or leave: the two branches now print near-identical prose one after the other, and a reader of the source has to hold both messages in their head to see which fires when. A single branch with the hint appended conditionally would say the same thing once. Not a condition of merging.

vladimirrott:
Pushed the mechanical half to your branch rather than asking again. Your two commits stay first and unchanged; mine sits on top.

**Rebased onto main**, which resolved the conflict on its own: main had moved past the point your `runner.rs` hunk collided with.

**`cargo fmt --all`**, the two diffs CI reported, both inside your own lines. The `db_path == Path::new(PRODUCTION_DATABASE_PATH)` condition and the `NotFound` arm of `permission_denied_counts_as_a_present_store`; rustfmt wanted each on one line. Running it before the rebase would have been wasted, since the `runner.rs` hunk moved.

**The baseline at 1,774.** Worth knowing what happened there, because it is the reason this sat: the rebase silently dropped your own baseline commit as already-applied. You recorded 1768 to 1771, and main reached 1771 independently through #285 and #295, so git saw nothing left to apply and your three added tests went uncounted. Regenerated at 1774 with the three prose figures, and `check_evidence_claims.py` agrees.

That the target moved twice while this waited is #306, and #308 makes it the maintainer's job rather than yours.

I reviewed the logic again on the rebased tree and it reads correctly: `path_is_present` is true under EACCES so the first branch catches only ENOENT, `exists()` is false under EACCES so the second catches the unreadable store and prints the sudo hint. `scripts/ci-local.sh` passes the full board on the pushed head.

One design note you can take or leave, not a condition: the two branches now print near-identical prose one after the other, and a reader has to hold both messages in mind to see which fires when. A single branch with the hint appended conditionally would say it once.

Merging once CI agrees. #266 closes with this.

vladimirrott (approved):
Green on `8f37b50` after two rebases: 11 checks pass, `container-smoke` path-skipped. Logic re-read on the rebased tree and it holds. The baseline landed at 1,775, measured each time rather than computed, and `docs/distro-support.md` re-conflicted on every rebase, which is #306 in practice. Thanks for the EACCES fix and for the fast turnaround on the test ordering; #266 closes with this.

Issue #278, opened 2026-08-24
Nothing checks that workspace-tests.json was measured at the commit it names


vladimirrott:
`scripts/record_test_baseline.py` stamps two provenance fields beside every count it records:

```python
doc.setdefault("measured_at", {})[args.field] = ...          # line 57
doc.setdefault("commit", {})[args.field] = args.commit or "unknown"   # line 63
```

Nothing reads them back. `scripts/test_baseline.sh` compares a live run's count against the `tests` field and stops. `scripts/check_public_claims.sh` compares the three published figures against the `tests` field and stops. So editing one integer by hand satisfies the entire board: the suite reports N, the artifact says N, README, `docs/introduction.md` and `docs/distro-support.md` all say N, and the artifact still names the commit and timestamp of a measurement that produced a different number.

Current main shows the drift. `commit.tests` is `c1cadf045edb7e118a6512d98ac4692a9c370a81`:

```
$ git merge-base --is-ancestor c1cadf045edb7e118a6512d98ac4692a9c370a81 origin/main; echo $?
1
$ git branch --contains c1cadf045edb7e118a6512d98ac4692a9c370a81
  p256
```

A merge commit from a branch that squash-merge collapsed. The field points at history that no longer exists.

## Why it matters

The artifact exists so a published number derives from a measurement instead of somebody's recollection. A provenance field nobody validates gets you back to recollection with extra steps. The count itself is still guarded on every PR, so no figure on main is wrong today.

## Scope

The squash-merge interaction is the interesting part, and it is why a naive fix will not hold. A contributor who regenerates the artifact records their own branch head. That commit is reachable from the PR head, so an "ancestor of HEAD" check passes in the PR's own CI run, then stops being true the moment the PR squash-merges and the recorded commit disappears from main.

Pick one:

- Validate ancestry only where it can hold, meaning in the PR run, and accept that main's copy carries a dead reference.
- Record something that survives a squash. The hash of the tree that holds the tests would still identify what was measured.
- Drop the `commit` field and stop implying a provenance nobody checks. `measured_at` alone is honest about being informational.

Whichever way it goes, say so in `CONTRIBUTING.md` next to the existing test-count instructions, because the failure mode this creates is a contributor hand-editing the integer and watching every gate go green.

## Tests first

`tests/release/` holds shell tests for guards like this. Write one that edits a count, leaves the provenance fields alone, and expects a non-zero exit.

## Difficulty

[...]

Issue #277, opened 2026-08-24
binding_status reports consistent after a database nobody could read


vladimirrott:
`cannot_verify_all` builds the result for "the store or the key would not open" (`apps/sysknife-cli/src/runner.rs:1118`). Chain and events both become `CannotVerify`, and `attribution` is `None` for a reason the code states:

```rust
// The store or the key could not be opened, so no census was taken.
// `None`, not a census of zero rows: a database nobody could read and an
// empty one that read fine must not serialize the same way.
attribution: None,
```

`binding` breaks that rule three lines earlier:

```rust
binding: BindingOutcome::Consistent { bindings_checked: 0 },
```

Both readers then discard the count. `apps/sysknife-cli/src/mcp_server.rs:1059`:

```rust
BindingOutcome::Consistent { .. } => "consistent",
```

and the JSON writer at `runner.rs:1381` matches the same way. So `audit verify` against a database it could not open reports `binding: consistent` beside a chain that reports `cannot_verify`.

## Why it matters

The verdict itself stays correct. `AuditVerification::exit_code` takes the worst of the three checks, so an unopenable store still exits 2 and no caller is misled about the outcome. What misleads is the field a reader greps or a dashboard charts, and on a tamper-evidence tool, "consistent" for a check that never ran is the wrong word by the standard quoted above.

## Scope

- Give `BindingOutcome` a third variant for "not checked".
- Return it from `cannot_verify_all` (`runner.rs:1125`), which is the only production site that builds this state. The other `Consistent { bindings_checked: 0 }` literals live under `#[cfg(test)]` and are fixtures.
- Map it to `2` in `binding_outcome_to_exit_code` (`crates/sysknife-daemon/src/audit_chain.rs:1421`) and to its own label in both readers.
- `verify_event_binding` keeps returning `Consistent` with a real count.

## Tests first

`binding_outcome_to_exit_code` and both label mappings are pure functions. Assert that an unreadable store produces a binding label other than `consistent`, and that `exit_code()` still returns 2 rather than 1.

## Difficulty

`easy` as a change, one variant threaded through four match sites. It touches audit reporting, so read the audit-chain section of `docs/architecture.md` first.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands. MIT, no CLA.

k4its1t:
[...]

PR #280, opened 2026-08-24
fix(cli): operator_safe drops U+2028 and U+2029

### LINKED ISSUE #276: operator_safe keeps U+2028 and U+2029, so the invisible-character set it calls complete is not
`operator_safe` drops every invisible formatting character before a plan summary reaches the terminal, and says so (`apps/sysknife-cli/src/operator_text.rs:83`):

```
///   The set is meant to be complete; a partial one reads as neutralised
///   without being so.
```

Two characters fall through. The drop arms cover `U+200B..=U+200F`, `U+202A..=U+202E`, `U+2066..=U+2069`, `U+FEFF`, soft hyphen, CGJ, the Mongolian selectors, `U+2060..=U+2064`, `U+206A..=U+206F`, interlinear annotation and the tag block. U+2028 LINE SEPARATOR and U+2029 PARAGRAPH SEPARATOR sit in the gap between `U+200F` and `U+202A`, so `_ => true` keeps them:

```rust
// operator_text.rs:107
'\u{202a}'..='\u{202e}' | '\u{2066}'..='\u{2069}' => false,
```

#274 closed exactly this two-codepoint gap in `crates/sysknife-brain/src/sanitize.rs`, on the path to the model. This is the operator-facing half of the same hole.

## Why it matters

Terminals do not break a line on U+2028. They draw a placeholder or nothing, which puts both codepoints in the class this file already drops by name: a character that renders as nothing and "can therefore split a word the operator reads as whole". `runner.rs:515` prints a preview summary directly above an approval prompt, so `remove\u{2028}nothing` is a plan the operator reads one way and approves for another.

The blast radius stays small. `render_multiline` counts lines with `str::lines`, which ignores both codepoints, so neither one pushes the rendered line count past `MAX_RENDERED_LINES`.

## Scope

- Add `'\u{2028}' | '\u{2029}'` to the drop arms, or widen the bidi arm to `'\u{2028}'..='\u{202e}'` the way #274 did.
- Add both codepoints to the loop in `bidi_overrides_and_zero_width_characters_are_dropped`.
- Confirm `operator_safe_block` inherits the fix, since it calls `operator_safe` per line.

## Tests first

The existing loop in `bidi_overrides_and_zero_width_characters_are_dropped` passes straight through the gap. Add the two codepoints, watch it fail, then touch the match.

## Difficulty

`easy`. One match arm and one test loop.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands. MIT, no CLA, no copyright waiver.

Georgefifth:
## Summary

`operator_safe` documents its invisible-character drop set as complete, but the arms covered `U+200B..=U+200F` and `U+202A..=U+202E`, leaving U+2028 LINE SEPARATOR and U+2029 PARAGRAPH SEPARATOR in the gap between them. Terminals render both as nothing — the same class the function already drops by name — so a plan summary like `remove\u2028nothing` can read one way to the operator approving it and mean another. This is the operator-facing half of the hole #274 closed on the model path.

The fix widens the bidi arm to `'\u{2028}'..='\u{202e}'`, the same widening #274 applied. `operator_safe_block` inherits it (it calls `operator_safe` per line); the issue notes `str::lines` ignores both codepoints, so neither can inflate the rendered line count past `MAX_RENDERED_LINES`.

## Related Issue

Closes #276.

## Validation

- [x] Tests added or updated — `bidi_overrides_and_zero_width_characters_are_dropped` now covers both codepoints; confirmed failing before the match change, passing after (TDD)
- [x] Documentation updated if behavior changed — CHANGELOG entry added in the project's style
- [x] Security impact considered — this change only extends an existing drop set
- [x] Trust boundary preserved (daemon remains the only privileged executor) — CLI-only change, no daemon/planner/shell code touched
- [x] CI passes — locally: `cargo fmt --check`, `cargo doc` (`-D warnings`), `cargo clippy --workspace --all-features --all-targets -- -D warnings` all clean; `cargo nextest run --workspace --locked` — 1765/1768 pass; the 3 failures are in `sysknife-daemon::execute_spec` process-stop tests and reproduce identically on untouched `main` in this container environment (process-group stop confirmation doesn't work here), unrelated to this change

## Notes for Reviewers

Single match arm + two test loop entries + changelog. No behavior change for any character outside U+2028/U+2029.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,15 @@ Releases before `0.2.5` predate the public launch; their notes live in the
   The shell script now reads `CLAIM_FILES` from the Python module, and refuses to
   report success when that read comes back empty, which `set -euo pipefail` cannot
   catch on its own inside process substitution. Thanks to @Osheun.
+- **`operator_safe` kept U+2028 and U+2029, so the invisible-character set it
+  calls complete was not.** ([#276](https://github.com/lacs-project/sysknife/issues/276))
+  The drop arms covered `U+200B..=U+200F` and `U+202A..=U+202E`, leaving the two
+  line/paragraph separators in the gap between them. Terminals render both as
+  nothing, which is exactly the class the function drops by name: a plan summary
+  reading `remove nothing` could hide a different target from the operator
+  approving it. The bidi arm is now `U+2028..=U+202E`, the same widening
+  [#274](https://github.com/lacs-project/sysknife/pull/274) applied on the
+  model-facing path.
 
 ### Documentation
 
--- a/apps/sysknife-cli/src/operator_text.rs
+++ b/apps/sysknife-cli/src/operator_text.rs
@@ -104,7 +104,10 @@ pub(crate) fn operator_safe(s: &str) -> String {
             // C1 controls: U+009B is CSI to a terminal in 8-bit mode.
             c if ('\u{80}'..='\u{9f}').contains(&c) => false,
             // Bidi overrides and isolates — reorder what is displayed.
-            '\u{202a}'..='\u{202e}' | '\u{2066}'..='\u{2069}' => false,
+            // U+2028/U+2029 sit in the gap below the bidi arm; terminals
+            // render them as nothing, so they belong in the same drop set
+            // (#274 widened the brain-side range the same way).
+            '\u{2028}'..='\u{202e}' | '\u{2066}'..='\u{2069}' => false,
             // Zero-width and byte-order marks — hide text in plain sight.
             '\u{200b}'..='\u{200f}' | '\u{feff}' => false,
             // U+061C ARABIC LETTER MARK is a bidi control in exactly the way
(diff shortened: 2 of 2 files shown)


vladimirrott (approved):
Right shape, and you extended the existing loop rather than adding a parallel test beside it. That is the difference between a guard and a second opinion.

Three things I checked past CI.

**The widened arm cannot shadow anything.** `operator_safe` matches in order, but every arm above yours also maps to `false`, and U+2028 is not in the `'\t' | '\n' | '\r' | ' '` arm, so widening only adds to the drop set. U+202F NARROW NO-BREAK SPACE stays outside the range, which is the right call: it renders as a space, not as nothing.

**Mutation.** Reverting to `'\u{202a}'..='\u{202e}'` fails your extended loop and names the survivor:

```
BASELINE:  12 passed; 0 failed
MUTATION:  bidi_overrides_and_zero_width_characters_are_dropped ... FAILED
           U+2028 survived: "remove\u{2028}nothing"
```

So the loop bites rather than passing on its own arithmetic.

**One subtlety in your favour.** `str::trim` uses `char::is_whitespace`, which is true for U+2028, so before your change a leading or trailing separator was already removed and only an interior one survived. Your test puts it in the middle, which is the case that was actually broken. Had it been at either end the test would have passed without the fix.

I resolved a conflict on your branch rather than sending you round the loop for it: #282 landed a CHANGELOG entry at the same anchor as yours. Both entries are kept, and I added the `Thanks to @Georgefifth.` line that its neighbours carry. Nothing else touched, and `scripts/ci-local.sh` passes on the result.

Merging once the fresh run is green. It goes out in the next release.

If you want more in this area, #277 is small and unclaimed: `audit verify` reports `binding: consistent` for a database it could not open, which is the wrong word for a check that never ran. #278 has more in it. And anything else you noticed in `operator_text.rs` while you were there is worth an issue of its own.

PR #282, opened 2026-08-24
fix(types): route outbound envelope encoding through the generated enums

### LINKED ISSUE #281: Outbound envelope encoding hand-types the wire numbers while inbound trusts the generated enum
The two directions of the envelope bridge disagree about which spelling of a wire number is authoritative.

**Outbound** writes the integer from a hand-maintained table. `crates/sysknife-types/src/lib.rs:758`, `:792`, `:839`, `:886`:

```rust
caller_role: caller_role_code(value.caller_role),
risk_level: risk_level_code(value.risk_level),
status: job_state_code(value.status),
```

Each table is a literal match (`:589`, `:598`, `:606`):

```rust
fn risk_level_code(value: RiskLevel) -> i32 {
    match value {
        RiskLevel::Low => 1,
        RiskLevel::Medium => 2,
        RiskLevel::High => 3,
    }
}
```

**Inbound** does not work that way. It routes through the prost-generated enum, which is derived from the `.proto`, so the authority is consulted on every decode (`:806`):

```rust
risk_level: RiskLevel::try_from(
    proto::RiskLevel::try_from(value.risk_level).map_err(|_| BridgeError::InvalidEnum {
        field: "risk_level",
        value: value.risk_level,
    })?,
)?,
```

A variant-to-variant conversion that would remove the hand-typed number already exists (`:650`), and nothing calls it. Its sibling at `:618` says why it was written: "this bridge is not on the live request path today".

```rust
impl From<RiskLevel> for proto::RiskLevel {
    fn from(value: RiskLevel) -> Self {
        match value {
            RiskLevel::Low => proto::RiskLevel::Low,
            RiskLevel::Medium => proto::RiskLevel::Medium,
            RiskLevel::High => proto::RiskLevel::High,
        }
    }
}
```

## Why it matters

Renumber `sysknife.proto` and the two directions part company. Inbound moves with the file, because prost regenerates. Outbound keeps writing the old integer, because the table is prose. A peer speaking the real `.proto`, or a daemon and CLI built from different commits, then reads `High` as `Medium`, and the risk level is what decides whether an approval receipt is required.

The round-trip tests cannot see it. `preview_envelope_round_trips_through_proto` encodes through `risk_level_code` and decodes through the generated enum, so it fails only if the two disagree today, never if the `.proto` moves out from under the outbound table.

Nothing is wrong right now. The numbers match. What is missing is anything holding them there on the encode side.

## Scope

- Replace each `*_code(...)` call with `i32::from(proto::X::from(value))`, which puts the outbound path through the same generated enum the inbound path already trusts, then delete the three tables.
- `FailureCategory` needs nothing: `proto::ResultEnvelope` carries no `failure_category` field, so it has no envelope encode path and no table. It is the shape the other three should end up in.
- Related but separate: #231 adds the missing `RiskLevel` anchor test. Worth having either way. That one is three lines in a test file; this one is the encode path, and a first-timer should be able to take #231 without inheriting this.

## Tests first

Renumber one arm of `risk_level_code` to `RiskLevel::High => 2` and run `cargo test -p sysknife-types`. Everything passes. That is the gap, stated as a mutation.

After the change the same edit has nowhere to land, because no integer is left in the encode path to edit.

## Difficulty

`easy`. Five call sites and three deletions, all in one file.

## Getting started

Georgefifth:
## Summary

The two directions of the envelope bridge disagreed about which spelling of a wire number is authoritative. Inbound routes through the prost-generated enum, so a renumbered `.proto` moves it. Outbound wrote integers from three hand-maintained match tables (`caller_role_code`, `risk_level_code`, `job_state_code`), so it would have kept writing the old numbers — a peer reading `High` as `Medium`, and the risk level is what decides whether an approval receipt is required.

The variant-to-variant conversions (`impl From<X> for proto::X`) already existed and were unused. This routes the five outbound call sites through them — `i32::from(proto::X::from(value))` — so the encode path consults the same generated authority the decode path already trusts, and deletes the three tables. The mutation the issue describes (renumbering `risk_level_code`) now has nowhere to land: no integer remains in the encode path.

## Related Issue

Closes #281. (#231 deliberately left untouched, as the issue asks.)

## Validation

- [x] Tests added or updated — no new tests needed; the existing round-trip suite covers the encode path and passes unchanged
- [x] Documentation updated if behavior changed — CHANGELOG entry added
- [x] Security impact considered — this removes a drift surface in the approval-deciding risk level encoding; no behavior change while the numbers match
- [x] Trust boundary preserved (daemon remains the only privileged executor) — types-crate conversion only
- [x] CI passes — locally in container: `cargo test -p sysknife-types` (3+13+8 pass), `cargo fmt --check`, `cargo doc` (`-D warnings`), `cargo clippy --workspace --all-features --all-targets -- -D warnings` all clean; `cargo nextest run --workspace --locked` — 1764/1768 pass; the 4 failures are the same `sysknife-daemon` process-stop tests that fail on untouched `main` in this container environment, unrelated to this change

## Notes for Reviewers

Net deletion (−38/+18). Five call sites, three table deletions, one stale comment refreshed, all in `crates/sysknife-types/src/lib.rs`. No references to the deleted functions remain anywhere in the tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,15 @@ Releases before `0.2.5` predate the public launch; their notes live in the
   The shell script now reads `CLAIM_FILES` from the Python module, and refuses to
   report success when that read comes back empty, which `set -euo pipefail` cannot
   catch on its own inside process substitution. Thanks to @Osheun.
+- **Outbound envelope encoding hand-typed the wire numbers while inbound
+  trusted the generated enum.** ([#281](https://github.com/lacs-project/sysknife/issues/281))
+  The three `From<…> for proto::…` conversions existed but were unused, and the
+  encode path wrote integers from three literal match tables instead. A
+  renumbered `.proto` would have moved the decode side with the file while the
+  encode side kept writing the old numbers — a peer reading `High` as `Medium`,
+  which decides whether an approval receipt is required. The encode path now
+  goes through the same prost-generated enums the decode path already trusts
+  (`i32::from(proto::X::from(value))`), and the three tables are deleted.
 
 ### Documentation
 
--- a/crates/sysknife-types/src/lib.rs
+++ b/crates/sysknife-types/src/lib.rs
@@ -586,40 +586,11 @@ pub struct TransactionRecord {
     pub warnings: Vec<String>,
 }
 
-fn caller_role_code(value: CallerRole) -> i32 {
-    match value {
-        CallerRole::Observer => 1,
-        CallerRole::Dev => 2,
-        CallerRole::Admin => 3,
-        CallerRole::Boot => 4,
-    }
-}
-
-fn risk_level_code(value: RiskLevel) -> i32 {
-    match value {
-        RiskLevel::Low => 1,
-        RiskLevel::Medium => 2,
-        RiskLevel::High => 3,
-    }
-}
-
-fn job_state_code(value: JobState) -> i32 {
-    match value {
-        JobState::Queued => 1,
-        JobState::Running => 2,
-        JobState::Succeeded => 3,
-        JobState::Failed => 4,
-        JobState::Canceled => 5,
-        JobState::RolledBack => 6,
-        JobState::NeedsReboot => 7,
-    }
-}
-
 impl From<CallerRole> for proto::CallerRole {
-    // Matched directly rather than round-tripping through the i32 code table
-    // and unwrapping. The conversion is total, so it should not be able to
-    // panic: this bridge is not on the live request path today, but once the
-    // daemon speaks proto it would be a panic-on-drift inside a root process.
+    // Matched directly against the variants rather than round-tripping through
+    // an i32 code table, so a renumbered `.proto` moves this conversion along
+    // with the prost-generated enum instead of drifting behind it. The
+    // conversion is total, so it cannot panic.
     fn from(value: CallerRole) -> Self {
         match value {
             CallerRole::Observer => proto::CallerRole::Observer,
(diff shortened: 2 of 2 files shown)


vladimirrott (approved):
Exactly the fix, and you did one thing I had not asked for. The comment on `From<CallerRole>` said the bridge "is not on the live request path today", which stopped being true the moment you routed the envelopes through it. Updating it beats leaving a comment that now misleads the next reader.

Checked on `d112234`: all five call sites moved, the three tables are gone, and nothing else in the workspace referenced them.

One thing my mutation pass turned up. It does not block this and it is not yours to fix, but it changes what the fix is worth knowing about.

I changed `RiskLevel::High => proto::RiskLevel::Medium` in the outbound map on your branch and the whole type-layer suite stayed green, 13 of 13. The reason is that every envelope test uses `RiskLevel::Medium`, and `proto_bridge.rs:130` and `:237` hard-code `risk_level: 2` on the decode side, so `High` and `Low` never traverse the map in either direction.

So your change removes the hand-typed integer, which was the ask and is a real improvement. What it leaves is the variant mapping itself, and nothing pins that. The same was true before your PR, so nothing regressed here.

It does mean #284 as written will not close that hole either: its assertion runs `proto::RiskLevel::try_from(3)` into the local type, which is the decode direction only. One extra line anchors the encode side:

```rust
assert_eq!(i32::from(proto::RiskLevel::from(RiskLevel::High)), 3);
```

I have left the same note on #284 so whoever gets there first has it.

Merging this. It goes out in the next release, and watching Releases is the way to catch that without me remembering to tell you.

If you want more in this area, #277 is next door: `audit verify` reports `binding: consistent` for a database it could not open, which is the wrong word for a check that never ran. And anything else you spotted in `sysknife-types` while you were in there is worth an issue of its own.

PR #284, opened 2026-08-24
sysknife-types: add the missing RiskLevel wire anchor test

### LINKED ISSUE #231: RiskLevel is the one wire enum with no anchored protobuf number test

Georgefifth:
Problem:
- `proto_bridge.rs` anchors `CallerRole`, `JobState` and `FailureCategory` each to a literal protobuf wire number so a .proto renumbering cannot pass unnoticed, but `RiskLevel` has no anchor — it only appears inside envelope round-trips, which encode and decode through mutually inverse hand-written tables and therefore cancel out. Renumbering both tables (e.g. Low=0, Medium=1, High=2) passes every test while a peer speaking the real .proto reads every High as Medium.

Solution:
- Add the anchor in the same shape as the three sibling tests: `proto::RiskLevel::try_from(3)` decoded to the local type, asserted equal to `RiskLevel::High`. The .proto authority is `RISK_LEVEL_HIGH = 3`.

Result:
- New test `risk_level_round_trips_through_proto` passes; the whole proto_bridge suite is 14/14. `cargo fmt --check` clean.

Fixes #231

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,771 Rust tests and 72 frontend tests** form the current deterministic
+**1,772 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/crates/sysknife-types/tests/proto_bridge.rs
+++ b/crates/sysknife-types/tests/proto_bridge.rs
@@ -34,6 +34,12 @@ fn risk_level_round_trips_through_proto() {
     let local_level = RiskLevel::try_from(proto_level).unwrap();
 
     assert_eq!(local_level, RiskLevel::High);
+
+    // Anchor the encode direction too. Decode alone leaves the outbound map
+    // unpinned: every envelope test uses RiskLevel::Medium, and proto_bridge.rs
+    // hard-codes `risk_level: 2` in both fixtures, so High never traverses the
+    // map and swapping its arm keeps the whole type-layer suite green.
+    assert_eq!(i32::from(proto::RiskLevel::from(RiskLevel::High)), 3);
 }
 
 #[test]
--- a/crates/sysknife-types/tests/proto_bridge.rs
+++ b/crates/sysknife-types/tests/proto_bridge.rs
@@ -29,6 +29,14 @@ fn failure_category_round_trips_through_proto() {
 }
 
 #[test]
+fn risk_level_round_trips_through_proto() {
+    let proto_level = proto::RiskLevel::try_from(3).unwrap();
+    let local_level = RiskLevel::try_from(proto_level).unwrap();
+
+    assert_eq!(local_level, RiskLevel::High);
+}
+
+#[test]
 fn request_envelope_round_trips_through_proto() {
     let value = RequestEnvelope {
         action_name: "InstallFlatpak".to_string(),
--- a/docs/distro-support.md
+++ b/docs/distro-support.md
@@ -53,7 +53,7 @@ apt and a read-only root, so the Debian-family action set cannot apply.
 | **Other Fedora Atomic 41+ variants** | rpm-ostree family | Detection and shared action tests | **Experimental** (eligible, awaiting a fresh VM run) |
 | **Fedora Workstation / Server** | `dnf` family incomplete | Detection tests only | **Experimental** |
 
-The deterministic workspace baseline is 1,771 Rust tests plus 72 frontend
+The deterministic workspace baseline is 1,772 Rust tests plus 72 frontend
 tests. Those tests verify action construction, policy, approval, storage, and
 UI behavior, but they do not replace a real distribution VM run.
 
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -141,7 +141,7 @@ flow.
 
 ## Status
 
-190 typed actions · 1,771 Rust tests + 72 frontend tests · MIT
+190 typed actions · 1,772 Rust tests + 72 frontend tests · MIT
 
 SysKnife is the reference implementation of the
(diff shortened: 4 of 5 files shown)


vladimirrott (changes requested):
The anchor is the right shape and sits beside its three siblings, which is what #231 asked for.

Two things before it lands.

**1. The test count moved.** Adding a `#[test]` takes the workspace to 1769, and the `rust` job compares that against the recorded 1768:

```
Starting 1769 tests across 39 binaries
```

`UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` rewrites `tests/evidence/workspace-tests.json` with the count, the commit and the timestamp together, which is better than editing the integer by hand. Three docs then carry the figure and have to move with it: `README.md`, `docs/introduction.md` and `docs/distro-support.md`. CONTRIBUTING documents the command under "Branch, code, test".

**2. Anchor the encode direction too, while you are here.** Your assertion covers decode: `try_from(3)` into `proto::RiskLevel`, then into the local type. I mutation-tested the other direction and it is not covered. Changing `RiskLevel::High => proto::RiskLevel::Medium` in the outbound map leaves the whole type-layer suite green at 13 of 13, because every envelope test uses `RiskLevel::Medium` and `proto_bridge.rs:130` and `:237` hard-code `risk_level: 2`, so `High` never traverses the map in either direction.

One more line closes it:

```rust
assert_eq!(i32::from(proto::RiskLevel::from(RiskLevel::High)), 3);
```

#282 is merged now, so `From<RiskLevel> for proto::RiskLevel` is on the live encode path and that assertion reads the generated enum on both sides. You will want to rebase on `main` for the same reason.

One coordination note, and no harm done. I wrote in #281 that #231 was worth leaving as an on-ramp for someone arriving cold, and you said twice you would leave it, so I had it held for that. The work here is good and I would rather have it than not. Going forward, tell me what size you want and I will point you at it directly: there is plenty in the medium range, and I will keep the three-line ones for first-timers.

#277 and #278 are both unclaimed if you want something with more in it.

vladimirrott:
Refreshing one number from my review. It expired; nothing is wrong on your side.

I wrote 1769. `main` has taken three commits with Rust changes since, so the recorded baseline is 1771 and your one test makes the target **1772**:

```sh
git fetch upstream && git rebase upstream/main
# then 1,771 -> 1,772 in README.md, docs/introduction.md, docs/distro-support.md
```

The comma form matters, since the checker builds the literal as `{count:,} Rust tests`.

Nothing else moved. The encode-direction assertion is still the other item, and #282 is on `main`, so `From<RiskLevel> for proto::RiskLevel` is there to assert against.

A figure that expires while your PR waits is my problem rather than yours: I am the one merging the commits that move it. Written up as #306, where the fix I am leaning towards is that the maintainer regenerates it at merge instead of asking you to chase it. Say the word and I will push the baseline commit to your branch now so only the assertion is left.

vladimirrott:
Pushed both remaining items to your branch rather than asking a third time. Your commit stays first and the test is yours.

**The encode assertion**, the line from my review, with a comment saying why it is there:

```rust
assert_eq!(i32::from(proto::RiskLevel::from(RiskLevel::High)), 3);
```

I checked it earns its place before adding it. Mutating the outbound arm to `RiskLevel::High => proto::RiskLevel::Medium` leaves your original decode assertion green and fails the new one at `proto_bridge.rs:42`. Restored and re-run clean afterwards.

**The baseline**, at 1,772, with the three prose figures and `check_evidence_claims.py` agreeing. That is the convention going into CONTRIBUTING in #308: whoever merges is what moves the figure, so whoever merges carries it. It was never yours to chase.

`scripts/ci-local.sh` passes the full board on the pushed head. Merging once CI agrees.

If you would rather have written the assertion yourself, say so and I will drop my commit and wait; I would rather over-ask than take work off you. #277 and #278 are still unclaimed and both have more in them than this one did.

vladimirrott (approved):
Both items in, encode direction mutation-proved, baseline at 1,772 with the prose and the claim screen agreeing. Board green locally and 11 checks green here. Thanks for the anchor and for the rebase; #231 closes with this.

PR #289, opened 2026-08-24
fix(audit): report unrun binding checks

### LINKED ISSUE #277: binding_status reports consistent after a database nobody could read

k4its1t:
## Summary

Represent an unrun transaction-to-approval-event binding check separately from a successful zero-row check. When the audit store or key cannot be opened, CLI JSON and the MCP report now publish `not_checked` instead of `consistent`, while the aggregate result keeps exit code 2.

## Related Issue

Closes #277.

## Validation

- [x] Tests updated without changing the published test count
- [x] `cargo +stable test -p sysknife-daemon audit_chain --locked` — 83 passed
- [x] `cargo +stable test -p sysknife-cli binding --locked` — 2 passed
- [x] `cargo +stable fmt --all -- --check`
- [x] Clippy passed for `sysknife-daemon` and `sysknife-cli`; only the existing macOS-only vsock dead-code/unused-variable warnings were allowed
- [x] `npm ci`, production dependency audit, TypeScript check, and frontend baseline — 72 passed
- [x] Changed-file pre-commit hooks, Markdown lint, and repository secret scan
- [x] Documentation updated for the new machine-readable status
- [x] Security impact considered; no signing, verification, trust-boundary, or `verify_event_binding` behavior changed

`scripts/ci-local.sh --no-postgres` was also run on macOS. Its remaining failures reproduce platform assumptions already present on `main` (Linux-only vsock references, Ubuntu-only apt/distro tests, and Bash 4/GNU utility requirements); the Ubuntu PR checks are authoritative for those gates.

## Notes for Reviewers

`BindingOutcome::NotChecked` is a new public enum variant, so exhaustive downstream matches will need an arm. Under the project's 0.y versioning policy, that is a next-minor release consideration even though the runtime fix is small.

AI assistance disclosure: OpenAI Codex helped inspect the issue, implement the patch, and run the validation commands. I reviewed the resulting diff and test output before submission.

### DIFF
--- a/apps/sysknife-cli/src/mcp_server.rs
+++ b/apps/sysknife-cli/src/mcp_server.rs
@@ -310,8 +310,9 @@ pub struct AuditVerifyReport {
     /// `"cannot_verify"`. Reported separately from `status` so a clean
     /// authorisation trail can never paper over a tampered approval trail.
     pub approval_events_status: String,
-    /// `"consistent"` or `"missing_event"`: whether every event tip committed
-    /// by a transaction row is still present in the event chain.
+    /// `"consistent"`, `"not_checked"`, or `"missing_event"`: whether every
+    /// event tip committed by a transaction row is still present in the event
+    /// chain.
     pub binding_status: String,
     /// Backend label: a filesystem path for SQLite, the literal `"postgres"`
     /// for Postgres deployments.
@@ -1035,11 +1036,20 @@ fn outcome_label(outcome: &sysknife_daemon::audit_chain::VerifyOutcome) -> &'sta
     }
 }
 
+fn binding_outcome_label(outcome: &sysknife_daemon::audit_chain::BindingOutcome) -> &'static str {
+    use sysknife_daemon::audit_chain::BindingOutcome;
+    match outcome {
+        BindingOutcome::Consistent { .. } => "consistent",
+        BindingOutcome::NotChecked => "not_checked",
+        BindingOutcome::MissingEvent { .. } => "missing_event",
+    }
+}
+
 fn outcome_to_report(
     verification: sysknife_daemon::audit_chain::AuditVerification,
     backend: String,
 ) -> AuditVerifyReport {
-    use sysknife_daemon::audit_chain::{BindingOutcome, VerifyOutcome};
+    use sysknife_daemon::audit_chain::VerifyOutcome;
 
     // One helper for both report arms and for the `CannotVerify` arm below, so
     // the census can only reach the report one way. The first version of this
@@ -1055,11 +1065,7 @@ fn outcome_to_report(
         VerifyOutcome::CannotVerify { .. } => 0,
     };
     let approval_events_status = outcome_label(&verification.events).to_string();
-    let binding_status = match &verification.binding {
-        BindingOutcome::Consistent { .. } => "consistent",
-        BindingOutcome::MissingEvent { .. } => "missing_event",
-    }
-    .to_string();
+    let binding_status = binding_outcome_label(&verification.binding).to_string();
 
     // The detail fields describe the first *break*, wherever it was found. A
     // broken transaction chain is reported ahead of a broken event chain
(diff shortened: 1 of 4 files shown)


vladimirrott (approved):
This is the shape I was hoping for and then some. `NotChecked` as its own variant, `binding_outcome_to_exit_code` mapping it to 2, both readers going through one helper each instead of two open matches, a human-readable line for the non-JSON path, and `docs/the-audit-chain.md` updated so the machine-readable status is documented rather than discovered.

The part I want to point at is what you did with the tests. You renamed three existing ones and widened them instead of adding a fourth, so the published test count never moved and the artifact coupling stayed out of it. Two other PRs open right now are stuck on exactly that. Eleven checks green on `194f20e8`.

I also checked the thing that worried me before reading: `exit_code()` is unchanged for every existing case. `cannot_verify_all` used to give `binding` code 0 with `chain` at 2, so the worst-of-three was 2; now `binding` is 2 as well and the answer is still 2. Nothing that passed starts failing.

One consequence, and it is mine to carry rather than a change request.

`BindingOutcome` is `pub` in `sysknife-daemon`, which is published, and it is not `#[non_exhaustive]`. Under [Cargo's semver rules](https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new) adding a variant to such an enum is a major change, because a downstream `match` stops compiling. In the `0.y` series that lands on the middle digit, the same way v0.9.0 moved for removing `PRODUCTION_LISTEN_URI`. So this ships in 0.10.0, not 0.9.2. I cut 0.9.1 a few hours ago, which is timing rather than anything you did.

What I am **not** going to ask for is `#[non_exhaustive]` on that enum, even though it would make the next variant free. `sysknife-cli` matches `BindingOutcome` across a crate boundary in eleven places, and this repo leans on exhaustive matches as a compile-time guard on purpose. `audit_chain.rs` says so a few hundred lines above your change:

> This is an exhaustive `match`, not a sequence of `if let`s, and that matters more than it looks. With `if let` chains, adding a variant and forgetting its arm compiles fine and silently encodes the new generation byte-identically to `LegacyV1`.

Marking the enum non-exhaustive would force `_ =>` arms in the CLI and retire that guard for a version number. Your PR is the argument against it: adding `NotChecked` broke both readers at compile time, which is precisely why `mcp_server.rs` and `runner.rs` are in your diff at all. I would rather pay a minor bump and keep the compiler doing that work.

Merging. Thanks for claiming it in the thread first and for saying what you intended to do; it meant nobody else started on it.

#278 is unclaimed and in the same spirit as this one: the evidence artifact records the commit and timestamp a count was measured at, and nothing checks either, so a hand-edited number satisfies every gate. It needs a judgement call about squash-merge rather than just code, which is the interesting part.

Issue #71, opened 2026-07-22
feat(security): scoped sudoers.d management


vladimirrott:
## Motivation
Grant/revoke narrow sudo rights without hand-editing sudoers (risky — a syntax error locks out sudo).

## Proposed actions
- `grant_sudo_access` — write `/etc/sudoers.d/sysknife-<name>` (user, host=ALL, runas, NOPASSWD?, command list) via a helper that **`visudo -cf` validates before install** and rolls back on failure.
- `revoke_sudo_access` — remove the drop-in.
- `get_sudo_grants` (read-only) — list `/etc/sudoers.d/` entries.

## Verified
- `visudo -c` available (sudo pkg). Helper MUST validate with `visudo -cf <tmpfile>` before atomic move into `/etc/sudoers.d/` (0440). Command paths validated absolute; no wildcards unless explicitly flagged.

## Risk / role
get: Low → Observer. grant/revoke: **Admin (Critical)** — this is privilege escalation surface; require the strongest validation + audit-chain entry. Cross-distro.

**Wiring checklist (per action, 11 touch points):**
1. `KNOWN_ACTION_NAMES` (sysknife-types/src/lib.rs)
2. brain `KNOWN_ACTIONS` (propose_plan.rs) — same length + subset invariant
3. action module `specs()` + constructor
4. executor `build_action_spec` dispatch + per-arg validation
5. `policy.rs` `min_role_for_action`
6. `preview.rs` `PreviewProfile`
7. `prompt.rs` (optional NL hints)
8. `action_family.rs` (DEBIAN_ONLY / cross-distro classification)
9. `packaging/sysknife-sudoers` grant (+ scoped helper if free-form input)
10. tests/e2e install step for any new helper binary
11. unit + `action_consistency.rs` + e2e story tests

**Acceptance:** `cargo test` green (incl. `action_consistency`), clippy clean, live-verified in the noble QEMU VM (or syntax/context7-verified + flagged if the tool needs a network-gated package).

Issue #68, opened 2026-07-22
feat(security): PAM password policy (pwquality, faillock, chage)


vladimirrott:
## Motivation
Password-policy and account-lockout hardening (CIS/STIG). No PAM/password actions today.

## Proposed actions
- `set_password_policy` — pwquality (minlen, dcredit, etc.) via a scoped helper writing `/etc/security/pwquality.conf.d/`.
- `set_account_lockout` — faillock (deny, unlock_time) via `/etc/security/faillock.conf`.
- `set_password_aging` — `chage -M/-m/-W <user>`.
- `get_password_aging` (read-only) — `chage -l <user>`.

## Verified
- `chage` present on base image (live). pwquality/faillock need `libpam-pwquality` — **network-gated** in this sandbox; validate helper by syntax + document the package dep.

## Risk / role
get: Low → Observer. set_*: High → Admin. All numeric fields validated; username via existing `validated_username`. Debian-family focus (paths differ on Fedora → DEBIAN_ONLY unless dual-pathed).

**Wiring checklist (per action, 11 touch points):**
1. `KNOWN_ACTION_NAMES` (sysknife-types/src/lib.rs)
2. brain `KNOWN_ACTIONS` (propose_plan.rs) — same length + subset invariant
3. action module `specs()` + constructor
4. executor `build_action_spec` dispatch + per-arg validation
5. `policy.rs` `min_role_for_action`
6. `preview.rs` `PreviewProfile`
7. `prompt.rs` (optional NL hints)
8. `action_family.rs` (DEBIAN_ONLY / cross-distro classification)
9. `packaging/sysknife-sudoers` grant (+ scoped helper if free-form input)
10. tests/e2e install step for any new helper binary
11. unit + `action_consistency.rs` + e2e story tests

**Acceptance:** `cargo test` green (incl. `action_consistency`), clippy clean, live-verified in the noble QEMU VM (or syntax/context7-verified + flagged if the tool needs a network-gated package).

PR #290, opened 2026-08-24
fix(setup): refuse an install the pin file does not name

### LINKED ISSUE #223: SYSKNIFE_PINNED_SHA256SUMS silently stops enforcing when the sums file does not name the asset
`SYSKNIFE_PINNED_SHA256SUMS` lets an operator require a digest obtained out of band, because the release checksum and the release binary share one trust root. `verifySha256` is explicit that a broken pin must never degrade to no pin (`packages/setup/install-binary.js:369`):

```js
if (opts.pinnedSha256 !== undefined && opts.pinnedSha256 !== null) {
    const pinned = String(opts.pinnedSha256).trim().toLowerCase();
    if (!/^[0-9a-f]{64}$/.test(pinned)) {
      // Never treat an unusable pin as "no pin": that would silently disable a
      // control the operator deliberately switched on.
      throw new Error(...)
```

The lookup returns `null` for an asset the pinned file does not name (`install-binary.js:338`):

```js
return entry ? entry.split(/\s+/)[0].toLowerCase() : null;
```

and that `null` is what reaches the gate (`install-binary.js:646`):

```js
const pinFor = (asset) => (pinnedSumsText ? digestFor(pinnedSumsText, asset) : undefined);
```

`null` fails the `!== null` test, so the pin is skipped rather than enforced. The malformed-digest case throws; the missing-entry case passes silently.

## Why it matters

Asset names embed the release tag, so they change on every version bump. An operator pins digests for one release, the next `npx sysknife-setup` resolves `releases/latest` to a newer tag, no entry matches, and the installer prints:

```
✓ Pinning digests against /etc/sysknife/pins.txt
✓ SHA256 verified
```

Both lines are true in isolation and together they read as a pinned install. The first is printed at `install-binary.js:640`, before any lookup happens. Whoever can publish a release can then ship a root-capable daemon past a control the operator believes is active.

## Scope

- Throw when `pinnedSumsText` is set and `digestFor` returns `null`, naming the asset and the file, in the same shape as the malformed-digest error.
- Move or amend the `Pinning digests against …` line so it does not claim enforcement before the lookup has succeeded.

## Tests first

`packages/setup/tests/pinned-digest.test.mjs` covers a malformed digest and not a sums file that omits the entry, which is why this survived. Add the missing case first and watch it fail: a pinned file naming only `sysknife-v0.8.0-linux-x86_64` must refuse to install `sysknife-v0.9.0-linux-x86_64` rather than proceed unpinned.

## Difficulty

`easy`. One guard, one message, one test.

## Getting started

vsolano9:
Closes #223.

`digestFor` returns `null` for an asset the pin file does not name, `pinFor` passed that `null` to `verifySha256`, and the gate's guard was `!== undefined && !== null`. So the missing-entry case — the one an operator hits on the ordinary upgrade path, because asset names carry the release tag — skipped the check instead of failing it, while the installer printed that it was pinning.

### The change

- **`pinnedDigestFor(sumsText, filename, sourcePath)`**, beside `digestFor`, throws when the lookup comes back empty. `sourcePath` is in the message because editing that file is the operator's next move, and because "unpinned" and "pinned against the wrong file" need telling apart. `digestFor` keeps its return-`null` contract for any other caller.
- **`verifySha256`'s guard is now `!== undefined`.** A `null` arriving from anywhere is fatal through the existing not-a-valid-digest branch; only `undefined`, meaning no pin was requested, skips the gate. Belt and braces — `pinFor` throws first in practice.
- **The banner no longer claims enforcement before a lookup.** `Pinning digests against <path>` is now `Loaded pinned digests from <path>`, and each asset's verified line carries `(pinned)`, which can only print after that asset's digest was found and matched.

I put the throw in a helper rather than in `verifySha256` because the error has to name both the asset and the pin file, and `verifySha256` only knows the asset. It also keeps the new behaviour unit-testable instead of reachable only through a network install.

### Tests first

Added to `pinned-digest.test.mjs`, watched fail, then fixed — RED was 3 failing of 11:

```
✖ a pinned sums file that does not name the asset refuses the install
✖ a pinned sums file that names the asset returns its digest
✖ a null pin reaching verifySha256 is fatal, not treated as no pin
ℹ tests 11  ℹ pass 8  ℹ fail 3
```

The case from the issue is there literally: a pin file naming only `sysknife-v0.8.0-linux-x86_64` must refuse `sysknife-v0.9.0-linux-x86_64`. There is also a test that `digestFor` still returns `null` for a missing entry, so the split of responsibilities is pinned rather than implied, and one that an absent pin option still means no pin.

| mutation | result |
| --- | --- |
| missing-entry no longer throws | 1 failed |
| `null` treated as no pin again (the old guard) | 1 failed |
| error stops naming the asset and the file | 1 failed |
| unmutated | 0 failed |

### Validation

`npm test` in `packages/setup`: **90 tests, 86 pass, 4 fail.**

Those 4 are pre-existing and not from this change — `git stash` on the same machine gives **85 tests, 81 pass, the same 4 fail**. They are the `system-daemon-mode` cases, and they fail here because this is macOS: `installDaemonService` returns bare `undefined` when `hasSystemd()` is false (`install-daemon.js:302`), so the assertions get `undefined` instead of a result object. That is worth its own issue and I will file it separately rather than widen this PR — it is a real gap, not only a test-portability one, since `index.js:924` guards on `daemonInstall &&` and therefore prints no outstanding-steps warning on a host without systemd.

No Rust is touched (`git diff --name-only` matches no `*.rs`), so `tests/evidence/workspace-tests.json` and the three prose figures are untouched — the JS suite size is not a published figure.

AI disclosure, as on #224: prepared with AI assistance. I reviewed the complete diff and ran every command and mutation above myself.

### DIFF
--- a/packages/setup/install-binary.js
+++ b/packages/setup/install-binary.js
@@ -338,6 +338,33 @@ function digestFor(sumsText, filename) {
   return entry ? entry.split(/\s+/)[0].toLowerCase() : null;
 }
 
+/**
+ * Look up `filename`'s digest in an operator-supplied pin file, or throw.
+ *
+ * `digestFor` returns null for an asset the file does not name, which is the
+ * right answer to "is it in there". It is the wrong thing to hand to the gate:
+ * asset names carry the release tag, so a pin file written for one release
+ * names nothing in the next, and passing that null on skips the check instead
+ * of failing it — silently disabling a control the operator switched on, which
+ * is exactly what the malformed-digest branch in verifySha256 refuses to do.
+ *
+ * `sourcePath` is in the message because the operator's next move is to edit
+ * that file, and because "unpinned" and "pinned against the wrong file" need
+ * to be told apart.
+ */
+function pinnedDigestFor(sumsText, filename, sourcePath) {
+  const pinned = digestFor(sumsText, filename);
+  if (pinned === null) {
+    throw new Error(
+      `Pinned digests were requested, but ${sourcePath} does not name "${filename}". ` +
+      `Asset names carry the release tag, so a pin file written for an earlier release ` +
+      `will not name this one. Add the digest for "${filename}" or unset ` +
+      `SYSKNIFE_PINNED_SHA256SUMS. Refusing to install rather than ignore the pin.`
+    );
+  }
+  return pinned;
+}
+
 function verifySha256(data, sumsText, filename, opts = {}) {
   const lines = sumsText
     .split('\n')
@@ -366,11 +393,14 @@ function verifySha256(data, sumsText, filename, opts = {}) {
   // release itself is authentic. An operator who obtained the digest
   // independently — from the signed git tag, an internal mirror, or a
   // config-management value — can require it here.
-  if (opts.pinnedSha256 !== undefined && opts.pinnedSha256 !== null) {
+  if (opts.pinnedSha256 !== undefined) {
     const pinned = String(opts.pinnedSha256).trim().toLowerCase();
     if (!/^[0-9a-f]{64}$/.test(pinned)) {
       // Never treat an unusable pin as "no pin": that would silently disable a
-      // control the operator deliberately switched on.
+      // control the operator deliberately switched on. `null` counts as
+      // unusable — it is what a lookup returns when the pin file does not name
+      // the asset, and it used to fall through this guard as though unpinned.
+      // Only `undefined`, meaning no pin was requested at all, skips the gate.
       throw new Error(
         `Pinned SHA256 for ${filename} is not a valid digest: ${JSON.stringify(opts.pinnedSha256)}. ` +
         `Expected 64 hex characters. Refusing to install rather than ignore the pin.`
@@ -637,13 +667,16 @@ async function installBinaryIfMissing(opts) {
   if (pinnedSumsPath) {
     try {
       pinnedSumsText = fs.readFileSync(pinnedSumsPath, 'utf8');
-      ok(`Pinning digests against ${pinnedSumsPath}`);
+      // Says what has happened, not what will. Enforcement is reported per
+      // asset below, once a digest has actually been found and matched.
+      ok(`Loaded pinned digests from ${pinnedSumsPath}`);
     } catch (e) {
       err(`SYSKNIFE_PINNED_SHA256SUMS is set but ${pinnedSumsPath} could not be read: ${e.message}`);
       process.exit(1);
     }
   }
-  const pinFor = (asset) => (pinnedSumsText ? digestFor(pinnedSumsText, asset) : undefined);
+  const pinFor = (asset) =>
+    (pinnedSumsText ? pinnedDigestFor(pinnedSumsText, asset, pinnedSumsPath) : undefined);
 
   // --- Download and verify sysknife CLI ---
   console.log();
@@ -658,8 +691,10 @@ async function installBinaryIfMissing(opts) {
   try {
     const digest = verifySha256(cliBuf, sumsText, cliAsset, { pinnedSha256: pinFor(cliAsset) });
     // Print the digest rather than only "verified": a checksum nobody can see
-    // cannot be cross-checked against a signed tag or an internal mirror.
-    ok(`SHA256 verified: ${cliAsset}  ${digest}`);
+    // cannot be cross-checked against a signed tag or an internal mirror. The
+    // "pinned" marker is printed only on this line, after the pin was found and
+    // matched, so it can never appear for an asset that went unpinned.
+    ok(`SHA256 verified${pinnedSumsText ? ' (pinned)' : ''}: ${cliAsset}  ${digest}`);
   } catch (e) {
     err(e.message);
     process.exit(1);
@@ -678,7 +713,7 @@ async function installBinaryIfMissing(opts) {
     const digest = verifySha256(daemonBuf, sumsText, daemonAsset, { pinnedSha256: pinFor(daemonAsset) });
     // Print the digest rather than only "verified": a checksum nobody can see
     // cannot be cross-checked against a signed tag or an internal mirror.
-    ok(`SHA256 verified: ${daemonAsset}  ${digest}`);
+    ok(`SHA256 verified${pinnedSumsText ? ' (pinned)' : ''}: ${daemonAsset}  ${digest}`);
   } catch (e) {
     err(e.message);
     process.exit(1);
@@ -710,6 +745,7 @@ module.exports = {
   detectPlatform,
   verifySha256,
   digestFor,
+  pinnedDigestFor,
   isOnPath,
   stageForPrivilegedInstall,
   fetchLatestRelease,
--- a/packages/setup/install-binary.js
+++ b/packages/setup/install-binary.js
@@ -365,6 +365,23 @@ function pinnedDigestFor(sumsText, filename, sourcePath) {
   return pinned;
 }
 
+/**
+ * Build the per-asset pin lookup the download path hands to `verifySha256`.
+ *
+ * This is a named function rather than a closure at the call site so the wiring
+ * itself is testable. The defect this whole path fixes was a call site handing
+ * the gate a value the lookup never refused; a test that only calls
+ * `pinnedDigestFor` proves the lookup throws and proves nothing about what the
+ * installer asks. Reverting this to `digestFor` has to fail a test.
+ *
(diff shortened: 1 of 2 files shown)


vladimirrott (changes requested):
This is the best-argued PR the tracker has had. Three things in it I would not have asked for and am glad you did.

Splitting `pinnedDigestFor` out instead of putting the throw inside `verifySha256` is the right call for the reason you give: the error has to name the pin file, and `verifySha256` only knows the asset. `digestFor` keeping its null contract, with a test that pins that contract rather than leaving it implied, is what makes the split safe to rely on later.

Widening the guard to `!== undefined` so a stray `null` dies in the not-a-valid-digest branch is belt and braces, and I checked it earns its place. Reverting that one guard fails your `a null pin reaching verifySha256 is fatal` test.

And you found something #223 never mentioned: the banner promised enforcement before any lookup had happened. `Pinning digests against X` becoming `Loaded pinned digests from X`, with `(pinned)` moved onto the per-asset line that can only print after a match, is the reporting half of the same defect. A control that announces itself before it runs is how this stayed invisible in the first place.

**One gap, and it is the same shape as the bug.** I mutation-tested both halves. The guard is covered. The wiring is not:

```
BASELINE                                            90 tests, 90 pass
MUTATION A  pinFor back to digestFor (the bug)      90 tests, 90 pass   ← no test fails
MUTATION B  verifySha256 guard accepts null again   90 tests, 89 pass, 1 fail
```

Your refusal test calls `pinnedDigestFor` directly, so it proves the helper throws. Nothing proves `installBinaryIfMissing` calls the helper. Someone can revert that one line in `pinFor` and the suite stays green, which is precisely the failure this PR is fixing: the lookup was always right, and the call site handed its answer to the gate unchanged.

The smallest close I can see is lifting the closure out and exporting it:

```js
function pinLookup(pinnedSumsText, pinnedSumsPath) {
  return (asset) =>
    pinnedSumsText ? pinnedDigestFor(pinnedSumsText, asset, pinnedSumsPath) : undefined;
}
```

Then `installBinaryIfMissing` builds it with `pinLookup(pinnedSumsText, pinnedSumsPath)`, and a test asserts that a lookup built over a stale pin file throws for an asset it does not name, and returns `undefined` when no pin file was loaded. That puts the reverted line under test rather than the helper it calls.

If you would rather ship the fix now and let me add that test myself, say so and I will merge this as it stands. It is a live hole and I do not want to sit on it over a wiring test. Your call.

Separately: thank you for #291. That is the second thing you have found by reading the code around what you came for, and the comment three lines above the guard you cite says "Reporting MCP configuration as done while the daemon does not exist is what made this look like a successful install", which means the class was fixed once and this path was missed. I have added a note there about the `skip` path.

vsolano9:
Closed the gap rather than take the merge-as-is offer — you found the one thing that made the fix unfalsifiable, and shipping it untested would leave the same shape of hole the PR is about.

`pinLookup` is lifted out and exported exactly as you sketched, and four tests now cover what `installBinaryIfMissing` builds instead of the helper it calls:

- a lookup over a stale pin file throws for an asset it does not name, naming both the asset and the file
- it returns the pinned digest for an asset the file does name
- with no pin file loaded it yields `undefined` — deliberately not `null`, since `verifySha256` takes that down the unusable-pin path, so a lookup returning it would refuse every unpinned install
- the value it produces is what stops the install at `verifySha256`

Your table, re-run on `9982a1e`:

```
BASELINE                                       100 tests, 100 pass
MUTATION A  pinLookup body back to digestFor   100 tests,  98 pass, 2 fail
MUTATION B  verifySha256 guard accepts null    100 tests,  99 pass, 1 fail
```

A is the one that mattered — 90/90 before, now failing both new wiring tests. B is unchanged, still caught only by `a null pin reaching verifySha256 is fatal`.

I also ran a third control you did not ask for, because two of the new tests call `pinLookup` directly and I wanted to know what they do *not* prove: replacing the call site with `const pinFor = (asset) => undefined` leaves the suite 100/100. So the wiring is now pinned one level deeper than it was, not all the way to `installBinaryIfMissing`, which downloads over the network and calls `process.exit`. Testing that needs an injection seam through the whole download path, which is a larger change than this PR should carry. Saying so rather than implying full coverage.

**Rebased onto main.** The branch point predated #302 and carried four failing `system-daemon-mode` tests that main had already fixed — worth knowing they were never mine. The CI command `npm test --prefix packages/setup` is 100/100 on the rebased head, no pre-existing failures.

vlad-dev28 (dismissed):
Merged. I reproduced your table on `9982a1e`, including the control:

| mutation | you | me |
| --- | --- | --- |
| baseline | 100 pass | 100 pass |
| `pinLookup` body back to `digestFor` | 98 pass, 2 fail | 98 pass, 2 fail |
| `verifySha256` guard accepts `null` | 99 pass, 1 fail | 99 pass, 1 fail |
| call site ignores the lookup | 100 pass | 100 pass |

The two that fail under mutation A are #68 and #71, which is the wiring the earlier version could not falsify.

The third control is the reason this review is short. You ran an experiment whose only possible outcome was to weaken your own claim, reported it, and drew the boundary in the right place: the lookup is pinned, the call site is not, and pinning the call site means an injection seam through a path that downloads and calls `process.exit`. I would have accepted the PR without it and been wrong to.

Two things I checked beyond the table.

**Completeness.** Every asset that reaches `verifySha256` goes through the pin. There are exactly two call sites, cli and daemon, both passing `pinFor`, and `digestFor` is now unreachable from production code except through `pinnedDigestFor`. A third download added later would have to opt out deliberately rather than by omission.

**The `(pinned)` marker cannot lie.** `pinFor(asset)` is evaluated as an argument inside the same `try`, so a pin file that does not name the asset throws before `verifySha256` is entered and the run exits 1. Reaching the `ok(...)` line with `pinnedSumsText` set means the digest was found and matched. The comment says that; I wanted to confirm the control flow actually guarantees it.

The change is tighter on two independent axes, which is the right shape for this: the lookup refuses, and the gate refuses `null` even if a future lookup hands one over. Either alone would close #223. Both means a later edit to one does not reopen it.

On the deeper seam: open it as an issue if you want it, and say in the issue what the seam would have to expose. It is a design question about how much of `installBinaryIfMissing` should be injectable, and it deserves to be argued before it is written rather than bolted onto a bugfix.

Four for four now, and the AI disclosure on each has cost you nothing and told me exactly how much of the reasoning to re-derive. Keep it.

vladimirrott (approved):
Merged. I reproduced your table on `9982a1e`, including the control:

| mutation | you | me |
| --- | --- | --- |
| baseline | 100 pass | 100 pass |
| `pinLookup` body back to `digestFor` | 98 pass, 2 fail | 98 pass, 2 fail |
| `verifySha256` guard accepts `null` | 99 pass, 1 fail | 99 pass, 1 fail |
| call site ignores the lookup | 100 pass | 100 pass |

The two that fail under mutation A are #68 and #71, which is the wiring the earlier version could not falsify.

The third control is the reason this review is short. You ran an experiment whose only possible outcome was to weaken your own claim, reported it, and drew the boundary in the right place: the lookup is pinned, the call site is not, and pinning the call site means an injection seam through a path that downloads and calls `process.exit`. I would have accepted the PR without it and been wrong to.

Two things I checked beyond the table.

**Completeness.** Every asset that reaches `verifySha256` goes through the pin. There are exactly two call sites, cli and daemon, both passing `pinFor`, and `digestFor` is now unreachable from production code except through `pinnedDigestFor`. A third download added later would have to opt out deliberately rather than by omission.

**The `(pinned)` marker cannot lie.** `pinFor(asset)` is evaluated as an argument inside the same `try`, so a pin file that does not name the asset throws before `verifySha256` is entered and the run exits 1. Reaching the `ok(...)` line with `pinnedSumsText` set means the digest was found and matched. The comment says that; I wanted to confirm the control flow actually guarantees it.

The change is tighter on two independent axes, which is the right shape for this: the lookup refuses, and the gate refuses `null` even if a future lookup hands one over. Either alone would close #223. Both means a later edit to one does not reopen it.

On the deeper seam: open it as an issue if you want it, and say in the issue what the seam would have to expose. It is a design question about how much of `installBinaryIfMissing` should be injectable, and it deserves to be argued before it is written rather than bolted onto a bugfix.

Four for four now, and the AI disclosure on each has cost you nothing and told me exactly how much of the reasoning to re-derive. Keep it.

PR #293, opened 2026-08-24
fix(setup): report the daemon as outstanding when systemd is absent

### LINKED ISSUE #291: installDaemonService returns nothing when systemd is absent, so the wizard ends on "Setup complete"
## The defect

`installDaemonService` has four exit paths. Three return a result the caller can act on; the no-systemd path returns bare `undefined` (`install-daemon.js:302`):

```js
if (!hasSystemd()) {
  warn('systemd not detected — skipping daemon service install.');
  step('Start the daemon manually:  ' + daemonBinPath);
  return;                                    // ← no result
}
...
return { mode: 'skip',   daemonInstalled: false, manualSteps: [...] };   // :325
return await _installSystemService(daemonBinPath);                       // :329 → :443
return { mode: 'user',   daemonInstalled: true,  manualSteps: [] };      // :339
```

The caller guards on truthiness (`index.js:924`):

```js
if (daemonInstall && !daemonInstall.daemonInstalled && daemonInstall.mode === 'system') {
```

so on a host without systemd the outstanding-steps block is skipped and the wizard runs to `ok('Setup complete')` with a "Try it" section telling the user to ask their client for disk usage — against a daemon that was never installed.

To be fair to the current code: it is **not silent**. The `warn` at `:303` does print. But it prints roughly forty lines before the summary, the run still ends on `Setup complete`, and the structured signal the caller needs to say otherwise is absent. That is the same shape `system-daemon-mode.test.mjs` was written to close — its header says "Reporting MCP configuration as 'done' while the daemon does not exist is what made this look like a successful install." The fix landed on three paths and the fourth kept the old behaviour.

## Why it survived

`tests/system-daemon-mode.test.mjs` calls `installDaemonService` directly, so on any machine without systemd all four of its cases fail rather than skip:

```
✖ system mode reports that it did not install the daemon
  AssertionError: installDaemonService must return a result the caller can act on
    actual: undefined,  expected: true
✖ the printed install command uses sudo, as the Makefile requires
  TypeError: Cannot read properties of undefined (reading 'manualSteps')
✖ system mode explains how to finish the wizard afterwards
  TypeError: Cannot read properties of undefined (reading 'manualSteps')
✖ user and skip modes also report their outcome
  TypeError: Cannot read properties of undefined (reading 'mode')
```

CI never sees this: `npm test --prefix packages/setup` runs only on `ubuntu-latest` (`ci.yml:20`/`:51`, `release.yml:156`/`:173`), where `hasSystemd()` is true and the real branch is exercised. So the tests are green in CI and red on any macOS contributor's machine — reproduced today on macOS 26.6.2 / Node 26.7.0, `85 tests, 81 pass, 4 fail` on `main` at `a10e4d5`, unrelated to what I was working on (#290).

## Scope

Two things, and the second is arguably the more valuable:

- Return a result from the no-systemd path, in the same shape as its siblings — something like `{ mode: 'none', daemonInstalled: false, manualSteps: ['Start manually:  <path>'] }` — and widen the `index.js:924` condition so a not-installed daemon is reported as outstanding whatever the reason, rather than only when `mode === 'system'`.
- Make the tests provable off Linux. Either inject `hasSystemd` so the branch under test is chosen explicitly, or skip with a stated reason instead of failing. Skipping is the weaker option — a skip that hardens into "handled" is its own trap — so injection is worth the small refactor.

Whether `mode: 'none'` or reusing `'skip'` is right is a call for you: they mean different things to an operator (`skip` is a choice, `none` is a platform fact), but only one of them is already in the caller's vocabulary.

## Not in scope

Actually installing anything without systemd. The point is only that the function should say what it did.

## Offer

Happy to take this if you want it — it is a small change plus the test-injection refactor. Filing rather than pushing because #290 is already open from me and the `mode` naming is your call.

AI disclosure, as on #223/#224: prepared with AI assistance. I ran every command above and read each cited line myself.

vsolano9:
Closes #291. Taking it myself rather than leaving it parked — I picked `none` over reusing `skip` and explain why below, so it is still your call to overrule.

### What changed

`installDaemonService`'s no-systemd branch returned bare `undefined` while its three siblings returned a result. `index.js` guarded the outstanding-steps block on `daemonInstall &&` **and** `mode === 'system'`, so two not-installed outcomes ended on the success banner: a host with no systemd, and a deliberate skip.

- The no-systemd branch returns `{ mode: 'none', daemonInstalled: false, manualSteps }`.
- The caller reports **any** not-installed daemon, with a per-mode headline, and keeps the `--daemon-mode=skip` re-run hint only for the system path where a wizard step actually remains.
- `hasSystemd` is injectable, so a test picks the branch instead of inheriting the machine's init system.

### On the naming

`none`, not `skip`. A skip is a choice the operator made and a missing init system is a fact about the host; only one of them is worth re-running with a different flag, and an operator reading `mode: 'skip'` after never being asked would be misled.

It is deliberately **not** in `DAEMON_MODES`. That constant is the CLI's *input* vocabulary — `parseDaemonMode` validates against it and `daemon-mode.test.mjs` pins it to exactly three — and `--daemon-mode=none` could not do anything. There is a test asserting the separation so nobody widens one into the other later. `README.md` and `docs/quickstart.md` document only the three input modes and are unchanged.

### Why it survived

CI runs `npm test --prefix packages/setup` on `ubuntu-latest` only (`ci.yml:20`/`:51`, `release.yml:156`/`:173`), where `hasSystemd()` is true, so the bare `return` was unreachable there. On macOS all four `system-daemon-mode` cases failed instead of skipping:

```
✖ system mode reports that it did not install the daemon
  AssertionError: installDaemonService must return a result the caller can act on
    actual: undefined,  expected: true
✖ the printed install command uses sudo, as the Makefile requires
  TypeError: Cannot read properties of undefined (reading 'manualSteps')
```

**`main` on macOS 26.6.2 / Node 26.7.0: 85 tests, 81 pass, 4 fail. This branch: 90 tests, 90 pass.**

### Tests

RED first — 7 of 9 failing in `system-daemon-mode.test.mjs` before the fix. Added: the no-systemd branch returns a usable result; its `mode` is distinguishable from a skip; the branch wins over any `--daemon-mode`; `none` never leaks into `DAEMON_MODES`; and the reason is printed as well as returned.

The caller change is covered behaviourally rather than by matching source text — `setup-contract.test.mjs` spawns the real wizard in a throwaway cwd with `--no-binary` and asserts the outstanding block appears. It asserts the line both reasons share (`nothing will execute until these steps are done`), so it passes on Linux (`skip`) and macOS (`none`) alike.

| mutation | failures |
| --- | --- |
| caller gated on `mode === 'system'` again | 1 |
| no-systemd returns bare again | 4 |
| no-systemd masquerades as `skip` | 2 |
| injected `hasSystemd` ignored | 4 |
| unmutated | 0 |

### Verified end to end

Running the wizard on this macOS box, which previously printed no such block at all:

```
Next steps

  ⚠  No daemon service was installed: systemd was not detected. MCP clients are configured, but
  ⚠  nothing will execute until these steps are done:

    Start manually:  /usr/local/bin/sysknife-daemon
```

### Deliberately not changed

The run still ends on `ok('Setup complete')`. Softening that banner is a separate judgement about what "complete" means when the operator has outstanding steps, and #291 asked for the report, not the banner. Say the word if you want it conditional and I will follow up.

No Rust is touched, so `tests/evidence/workspace-tests.json` and the three prose figures are untouched; the JS suite size is not a published figure.

AI disclosure, as on #223/#224/#290: prepared with AI assistance. I reviewed the complete diff and ran every command, mutation and wizard run above myself.


vladimirrott (approved):
Merged. 91 tests pass here on Linux, where `hasSystemd()` is true, against your 90 on macOS. The delta is platform-gated cases, not a regression.

You picked the naming argument I would have picked, and you checked the thing that makes it safe. `daemonInstall.mode` had exactly one consumer, the line you rewrote, and `DAEMON_MODES` is read only by `parseDaemonMode`. So `none` cannot reach the input validator however many branches someone adds later, and your test says that out loud instead of trusting it.

The injectable `hasSystemd` is the part I looked at hardest, because a seam that lets a caller pick a branch is where a guard usually gets loosened. This one only lets a caller decline to install a daemon, which `--daemon-mode=skip` already allows, so there is nothing to escalate through it.

Leaving `ok('Setup complete')` alone was right for this PR. I will open the banner question on its own rather than have you guess what I want it to say.

PR #295, opened 2026-08-24
fix(audit): stop the Postgres auditor grading evidence it never read

### LINKED ISSUE #246: The Postgres auditor path reports a clean event chain for one it could not read
`sysknife audit verify --pubkey` is the auditor path: verify a chain with the exported public key, no signing key, no trust in the host that wrote it. On PostgreSQL that path discards every error from the approval-event read (`crates/sysknife-daemon/src/store/postgres.rs:238`):

```rust
let tx_rows = fetch_chain_rows_from_pool(&pool).await?;
// A chain written before the migration has no `audit_events` table.
// That is an absent feature, not a missing verification: report zero
// events rather than failing the whole verify.
let event_rows = fetch_event_rows_from_pool(&pool).await.unwrap_or_default();
```

The comment justifies one case. `unwrap_or_default()` covers all of them: a dropped table, a `SELECT` permission the auditor's role lacks, a connection lost mid-fetch, one malformed row. Each becomes "zero approval events".

Its sibling twelve lines down propagates instead (`postgres.rs:252`):

```rust
let event_rows = fetch_event_rows_from_pool(&self.pool).await?;
```

So the operator path with the signing key reports cannot-verify, and the third-party path the design is built around reports success.

## Two opposite false verdicts

**A clean bill for a trail nobody read.** An auditor granted `SELECT` on `transactions` but not `audit_events` gets `OK: N row(s) verified`, `OK: 0 approval event(s) verified`, and exit 0.

**A fabricated tamper.** If any transaction row committed a non-empty `event_tip`, the binding check over an empty slice returns `MissingEvent`, and the report becomes `BROKEN: ... approval events were deleted from the end of the chain` with exit 1. A permissions error gets published in the exact words `docs/the-audit-chain.md` teaches operators to read as erasure.

## Scope

Separate the case the comment describes from the rest. Probe `to_regclass('audit_events')` for the absent-table case, and map every other error to `VerifyOutcome::CannotVerify` with a binding outcome that is not `Consistent`.

`BindingOutcome` has no `CannotVerify` variant today, which is why the third state has nowhere to go. Adding it also fixes the neighbouring case where an unopenable database reports `binding: consistent, bindings_checked: 0`.

## Tests first

`crates/sysknife-daemon/tests/postgres_store.rs` runs against a fully migrated database, so the fallback never fires there. Write the failing case first: revoke `SELECT` on `audit_events`, then assert the verdict is cannot-verify rather than intact. Local Postgres for these tests runs under podman; docker's published ports drop connections on some hosts.

## Difficulty

`medium`. Small diff, and it needs a local Postgres plus a new enum variant threaded to the two renderers.

## Getting started

vsolano9:
Closes #246.

### One scope correction first

The issue says `BindingOutcome` has no variant for the third state and asks for one. It does now — `NotChecked` landed today in #289, six days after this was filed, wired to exit 2 and rendered by both the CLI and MCP paths. So **no new variant here**, and the neighbouring case the issue mentions (an unopenable database reporting `binding: consistent, bindings_checked: 0`) was already fixed there. What was left is the Postgres swallow itself, which is still on `main`.

### Both false verdicts, reproduced

One fixture produces both at once. Before the fix, an auditor granted `SELECT` on `transactions` and denied it on `audit_events`:

```
events  = Intact { rows_checked: 0 }
binding = MissingEvent { transaction_seq: 2, event_tip: "592ac378c67d874b0de993db78456809…" }
exit    = 1
```

The first line is the clean bill for a trail nobody read. The second is the fabricated tamper — a permissions error rendered in the words `docs/the-audit-chain.md` teaches operators to read as deleted approvals. Exit 1, so a pipeline treats it as a detected breach.

### The change

`to_regclass('audit_events')` is asked specifically for the absent-table case. It answers **existence, not access** — a table the caller may not `SELECT` still resolves — which is exactly the distinction needed. Everything else, including a probe that cannot answer, returns `CannotVerify` for the event chain and `NotChecked` for the binding.

The transaction chain and the attribution census are still reported. They were read, and an auditor who loses the event table should not lose the rest of the verdict too.

### Tests

RED first. Two live-Postgres cases, and the second matters as much as the first:

- **`an_auditor_denied_the_event_table_cannot_verify`** — the case you asked for, with a real `CREATE ROLE` + `GRANT SELECT` on `transactions` only. It also writes a second transaction *after* the approval events exist, so its row commits a non-empty event tip — that is what makes the fabricated-tamper verdict reachable rather than only the false clean bill.
- **`a_chain_predating_the_event_table_still_verifies`** — the branch I *preserved*. Without it, "return CannotVerify on any error" would pass every other test and quietly break pre-migration chains.

Both own a dedicated schema via `search_path` rather than the public one. The existing cases drop and recreate the shared tables with no serialisation between them, so a role grant made there could be revoked by a concurrent test dropping the table it names. Flagging that as an observation, not touching it here.

| mutation | result |
| --- | --- |
| restore `unwrap_or_default()` | 1 failed (`an_auditor_denied…`) |
| fail closed even when the table is absent | 1 failed (`a_chain_predating…`) |
| binding reports `Consistent { bindings_checked: 0 }` | 1 failed (`an_auditor_denied…`) |
| unmutated | 0 failed |

### Validation, and what I could not run

Local PostgreSQL 17.11, not a container, so the docker published-port problem you mention does not apply. `cargo test -p sysknife-daemon --test postgres_store`: **7 passed, 0 failed**, baseline before the change was 5.

`cargo fmt --all -- --check` clean. `cargo clippy -p sysknife-daemon --lib --tests -- -D warnings -A dead_code` clean.

Three things do not run on macOS, all pre-existing and all verified against pristine `main` on this machine:

- `cargo clippy --workspace --all-targets` — `sysknife-cli` does not compile: `no variant named Vsock found for enum SocketTarget` (Linux-gated), 3 errors on `main` too, plus `dead_code` on the vsock helpers in the daemon.
- `cargo test -p sysknife-daemon --lib` — 861 passed, 10 failed, **identical on `main`**: apt/distro cases that need a detectable host distribution.
- `tests/release/public-claims.test.sh` — `mapfile: command not found`; it needs bash 4+ and macOS ships 3.2.

**The test baseline is the one number I could not measure.** `scripts/test_baseline.sh` needs `cargo nextest run --workspace`, which cannot build here for the reason above, and this machine has no container runtime. So I derived it instead of measuring it, and I would rather say that plainly than imply otherwise:

- the recorded 1,769 is current for `main` — `a10e4d5` last wrote the artifact, and the only commit since (`ca67382`) touches no `.rs` file;
- `git diff upstream/main...HEAD -- '*.rs'` adds exactly **two** `#[tokio::test]` and removes none, in one existing test target;
- so 1,771, applied to the artifact and to the three prose figures, which I checked all derive from it.

I deliberately left `measured_at` and `commit` pointing at the last real measurement rather than stamping a run I did not perform; nothing reads those fields, `tests` is the only one checked. **CI's `scripts/test_baseline.sh` is the measurement** — if 1,771 is wrong this PR goes red on that step, which is the machinery working. Happy to push a correction the moment it says so.

AI disclosure, as on #223/#224/#290/#293: prepared with AI assistance. I ran every command, mutation and query above myself and reviewed the complete diff.


vsolano9:
Force-pushed one fix: CodeQL flagged a critical alert on the first push, and it was mine — `CREATE ROLE … PASSWORD 'audit-only-password'` is a hard-coded credential, throwaway role or not, and this repository is a poor place to write one down. The password is now generated at test time and never appears in source. All 12 checks are green.

Worth recording for the review: `rust` passing is the answer to the one number I said I could not measure. `scripts/test_baseline.sh` runs `cargo nextest run --workspace` there and fails when the artifact disagrees with the suite, so 1,771 is now measured rather than derived — my arithmetic held.

`postgres-contract` is green too, so the role-creation path in the new test works against the CI server and not only my local one.

vladimirrott (approved):
Merged. I re-ran this against Postgres 17 in a container rather than a local server and got your numbers: 7 passed on `postgres_store`, and restoring `unwrap_or_default()` fails `an_auditor_denied_the_event_table_cannot_verify` with `Intact { rows_checked: 0 }`, the false clean bill spelled out in the assertion message.

Three things worth naming.

The scope correction was right, and leading with it was right. `NotChecked` landed in #289 six days after this issue text was written, and reusing it instead of adding a fourth variant is what kept this off a minor version bump.

`to_regclass` is the correct probe, and I went looking for its boundary. It answers existence for a role denied `SELECT` on the table, which is your test:

```
role denied SELECT on audit_events   → to_regclass = 'audit_events'  (fails closed)
role denied USAGE on the schema      → to_regclass = NULL            (reads as absent)
```

So the second row still reaches the empty-events path. It is unreachable through this code: the migration creates both tables unqualified in one `search_path`, and `fetch_chain_rows_from_pool` propagates with `?`, so a role that cannot enter the schema fails on `transactions` before the event probe runs. Getting there needs a hand-split deployment where `transactions` is readable and `audit_events` sits in a schema the auditor cannot enter. Not a blocker. I would rather have it written here than found later.

On the baseline: you said you derived 1,771 instead of measuring it, and named CI as the measurement. That is how I want that handled. The `rust` job agreed with you.

Your observation about the existing cases sharing the public schema with no serialisation between them is a real defect in its own right. Take it as your next issue if you want it.

Issue #297, opened 2026-08-24
The wizard prints "Setup complete" and exits 0 after telling you nothing will execute


vladimirrott:
#293 made the wizard report an uninstalled daemon as outstanding, for all three reasons it can happen. It deliberately left the last two lines alone, and they still claim the opposite:

```js
  console.log();
  ok('Setup complete');      // packages/setup/index.js:982
```

So a run that installed nothing prints its own warning block and then, sixty lines later, `✓ Setup complete`. The operator sees a red flag followed by a green one, and the green one is last.

The exit code says the same thing. `main()` resolves, nothing calls `process.exit`, so the process exits 0. A scripted install reading `$?` cannot tell a finished run from one waiting on `sudo make install-daemon`.

### What to decide

Two questions, and the second is the one worth thinking about.

**The banner.** Something that names the state instead of asserting success when steps remain. `Setup incomplete: N step(s) left` reads honestly and keeps `Setup complete` for runs that are.

**The exit code.** Whether a deliberate `--daemon-mode=skip` is a failure. Argue it either way: the operator asked for it, so the wizard did what it was told; but a CI job that skips the daemon and then treats 0 as "sysknife is running" is wrong in a way no message on stdout fixes. A third code (2, matching this repo's use of 2 for inconclusive in `binding_outcome_to_exit_code`) separates "nothing to do" from "something failed" without breaking either reading.

Say which you picked and why in the PR. The choice is more of the work here than the diff.

### Scope

`packages/setup/index.js` only. Do not reopen the reporting block #293 just landed.

The contract test to extend is `packages/setup/tests/setup-contract.test.mjs`, which already spawns the real wizard with `--no-binary` in a throwaway cwd and asserts the outstanding block appears. Assert the banner and the exit code from the same run. There is no separate JS figure in `tests/evidence/`, so nothing needs regenerating.

### Difficulty

Small diff, real judgement call. The wizard's own behaviour is the whole subject, so read #291 and #293 first.

LunaMeerkats:
Autonomous OpenAI Codex agent here. I completed the current policy, assignment, and overlap preflight and am taking this from `main` at `0aa697b`.

[...]

PR #308, opened 2026-08-26
docs: stop asking contributors to chase a figure that moves without them

### LINKED ISSUE #306: The test baseline moves ~3x a week, so contributor PRs expire on a number they do not control
`90fc4a3` established that every published number derives from a run rather than from someone's memory. That was right and it should stay. The side effect is that a contributor's PR expires on a figure they do not control.

### The figure moves about three times a week

`tests/evidence/workspace-tests.json` over the last twenty days:

```
2026-08-05  1681      2026-08-13  1758      2026-08-24  1768
2026-08-06  1711      2026-08-16  1759      2026-08-24  1769
2026-08-08  1723 1727 1736 1737            2026-08-24  1771
2026-08-09  1739      2026-08-12  1743
2026-08-13  1745 1750
```

Fifteen moves in twenty days. Any PR that adds a Rust test and waits longer than a day or two needs its four figures re-derived, and the `rust` job goes red until it does.

### What it has cost, concretely

| PR | what happened |
| --- | --- |
| #284 | my review said the target was 1769. It is 1772 today. The number in my own review expired before the author acted on it. |
| #275 | recorded 1771 from a base of 1768. Rebased, the answer is 1774. |
| #260 | I quoted 1759 on 2026-08-20 and worked out 1766. The answer is now 1778. Two of my comments carried figures that stopped being true. |
| #295 | the author could not run `cargo nextest` on their machine, derived 1771 by counting `#[tokio::test]` in their own diff, and said so plainly rather than implying a measurement. |

Four of the last several outside PRs, and in two of them the stale number came from me.

### What not to do

**Do not relax the check to `tests >= recorded`.** Adding tests would stop failing, and the published figure would drift below reality and quietly stop being a measurement. That is the property `90fc4a3` bought and it is worth more than the friction.

**Do not have a bot rewrite the prose figures on merge.** README, `docs/introduction.md` and `docs/distro-support.md` carry published claims. A repo whose subject is approval-gated change should not have an unreviewed process editing its own claims about itself.

### What I think we should do

**The maintainer regenerates it, not the contributor.** I am the one merging the commits that move the target, so I should carry the consequence. When a PR is otherwise green and only the baseline is stale, push the regeneration to the branch (`maintainer_can_modify` is true on every fork PR so far) or land it and follow up. It preserves the invariant exactly, needs no code, and takes the one genuinely unrewarding step off the contributor.

`CONTRIBUTING.md` should then say so, because right now it tells contributors to run `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` and does not tell them what to do when the answer changes underneath them. It should say: add your tests, do not chase the figure, and say in the PR how many tests you added so the number can be derived at merge.

**Worth considering alongside it:** have the `rust` job print the expected value on failure, so a red run names the number instead of only the mismatch. That turns a puzzle into a copy-paste and helps whoever hits it anyway.

### Not a good first issue

The `CONTRIBUTING.md` half is small. The convention half is a maintainer decision about who owns a merge-time step, so I have labelled it accordingly.

vladimirrott:
Closes #306, and writes down the policy I gave in #297.

### The figure moves without the author

The recorded Rust test count moved fifteen times in the twenty days to 2026-08-24, 1681 to 1771. Any PR that adds a test and waits a day or two has its four figures invalidated by somebody else's merge, and `rust` stays red until the author re-derives a number they did not move.

It has cost four PRs, and in two of them the stale number came from me:

| PR | what happened |
| --- | --- |
| #284 | my review said 1769. It was 1772 by the time the author acted on it. |
| #275 | recorded 1771 from a base of 1768; needs 1774 rebased. |
| #260 | I quoted 1759 and worked out 1766. The answer is 1778 now. |
| #295 | the author could not run nextest at all, derived it from their own diff, and said so. |

So CONTRIBUTING now says: record it once, say how many tests you added, and if it has expired the maintainer regenerates it at merge. Whoever merges is what moves the figure. A `rust` job red only on the baseline does not block the author.

**One thing #306 proposed already exists.** I suggested making the `rust` job print the expected value on failure. `scripts/test_baseline.sh:126` already does:

```
test_baseline: rust suite has 1772 tests, baseline says 1771.
```

The run that made me think otherwise had hit the suite-failed path at `:91`, which withholds the count, correctly, since a failed suite has measured nothing. CONTRIBUTING now points at the behaviour rather than my adding it twice.

### A JavaScript change does not need the Rust gate

From #297, where a contributor was blocked on `cargo nextest run --workspace` for a change to `packages/setup` on a Windows host with no Linux build stack, and asked rather than bypassing it.

The workspace suite exists to stop a Rust regression reaching `main`. A diff with no `.rs` file in it cannot cause one, and that is checkable:

```sh
git diff --name-only upstream/main... | grep '\.rs$'   # empty means the gate does not apply
```

CONTRIBUTING now names what stands in its place: the Node suite for the package touched, `scripts/check_evidence_claims.py`, and CI on Ubuntu. It also says to run the claim screen anyway, since `packages/setup/index.js` is in `CLAIM_FILES` and the screen is pure Python with no subprocess calls, so it runs on Windows unchanged.

### Verified

Docs only. `markdownlint-cli2` clean on `CONTRIBUTING.md`, and the full board passed via the pre-commit hook with the baseline matching at 1,771.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,6 +118,32 @@ grep -rn 'Rust tests' README.md docs/introduction.md docs/distro-support.md
 CI gates both halves, so bumping the artifact alone turns `docs-and-hygiene` red
 after `rust` goes green.
 
+**Do not chase the figure.** It moved fifteen times in the twenty days to
+2026-08-24, so on any PR that waits a day or two the number you recorded stops
+being the answer, and that is not your problem to solve. Record it once, say in
+the PR how many tests you added, and if it has expired by the time the PR is
+ready the maintainer regenerates it at merge. Whoever merges is what moves the
+figure, so whoever merges carries it. If `rust` is red only on the baseline, the
+PR is not blocked on you.
+
+`scripts/test_baseline.sh` names both numbers when they disagree
+(`rust suite has 1772 tests, baseline says 1771`), so a red run tells you the
+answer rather than setting a puzzle.
+
+**A change that touches no Rust skips the Rust gate.** The workspace suite exists
+to stop a Rust regression reaching `main`, and a diff with no `.rs` file in it
+cannot cause one:
+
+```sh
+git diff --name-only upstream/main... | grep '\.rs$'   # empty means the gate does not apply
+```
+
+For such a diff the gate is the Node suite for the package you touched,
+`scripts/check_evidence_claims.py`, and CI on Ubuntu. Run the claim screen even
+for a JavaScript change: `packages/setup/index.js` is in `CLAIM_FILES`, and the
+screen is pure Python with no subprocess calls, so it runs anywhere Python does.
+Say which of the three you ran.
+
 **Which release your change lands in.** While SysKnife is in the `0.y` series, a
 change that breaks a consumer moves the middle digit and everything else moves the
 last one. [docs/release.md](docs/release.md#version-numbering) carries the rules
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,9 +126,12 @@ ready the maintainer regenerates it at merge. Whoever merges is what moves the
 figure, so whoever merges carries it. If `rust` is red only on the baseline, the
 PR is not blocked on you.
 
-`scripts/test_baseline.sh` names both numbers when they disagree
-(`rust suite has 1772 tests, baseline says 1771`), so a red run tells you the
-answer rather than setting a puzzle.
+`scripts/test_baseline.sh` names both numbers when they disagree, in the form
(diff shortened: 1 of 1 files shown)


vladimirrott:
CI caught this PR with the guard this PR is about. I illustrated the mismatch message with real digits, `check_evidence_claims.py` read 1772 as a published figure in a `CLAIM_FILES` entry, and refused it because it matches no recorded suite size. Placeholders now, with a note saying why, so the next person writing an example in a screened file does not spend a CI round finding out. Fixed in 9fed579.

PR #313, opened 2026-08-27
fix(test): fail postgres-contract when the database was never configured

### LINKED ISSUE #294: postgres-contract reports success when no database was configured
`postgres-contract` is one of the five required checks on `main`. Its job is to prove the live PostgreSQL store contract against a real server. Three of its five tests report success when no server was configured.

`crates/sysknife-daemon/tests/postgres_store.rs:12`:

```rust
fn test_url() -> Option<String> {
    std::env::var("SYSKNIFE_TEST_POSTGRES_URL").ok()
}
```

and at `:46`, `:372` and `:478`:

```rust
let Some(url) = test_url() else {
    eprintln!("SYSKNIFE_TEST_POSTGRES_URL is unset; live Postgres contract not requested");
    return;
};
```

An early `return` is a pass. Measured on `a10e4d5`:

```
with no postgres env var    5 passed; 0 failed; 0 ignored   0.30s
with SYSKNIFE_TEST_POSTGRES_URL set    5 passed; 0 failed; 0 ignored   1.21s
```

Same verdict, same counts, four times the duration. The two that behave identically are the URL-policy tests, which need no server. The three that carry the contract did nothing in the first run, and `0 ignored` says nothing was skipped.

## Why it matters

Nothing is wrong today: `.github/workflows/ci.yml:246` sets the variable and the service block starts `postgres:17-alpine`, so the job does the real work. The gap is that no signal distinguishes that from the alternative. Rename the variable, drop the `services:` block, change the image tag, or break the health check, and the job stays green while testing nothing, on the one check whose entire purpose is proving the thing it stopped doing.

The `eprintln!` is not that signal. nextest captures per-test stdout and prints it only for failures, so the line is invisible in exactly the run where it matters.

I walked into this while verifying a release: I ran the suite with the wrong variable name, got `5 passed`, and nearly recorded it as evidence the contract held.

## Scope

Make the skip loud where a server is expected, and visible where it is not. One shape:

- Read a second variable, `SYSKNIFE_REQUIRE_POSTGRES`, and `panic!` in the `else` arm when it is set. CI sets it beside the URL, so a broken service block fails the job instead of skipping it.
- Locally, with neither set, keep skipping, but mark those three `#[ignore]` so nextest reports them as ignored rather than passed. `0 ignored` is then a fact about the run instead of an accident.

Either half is an improvement; the pair is what makes the count readable.

## Tests first

This issue is reproducible in two commands, which is the test:

```sh
cargo test -p sysknife-daemon --test postgres_store --locked            # 5 passed
SYSKNIFE_REQUIRE_POSTGRES=1 cargo test -p sysknife-daemon --test postgres_store --locked   # must fail
```

`podman` works locally where `docker` gives ECONNRESET on the published port:

```sh
podman run -d --name sysknife-pg -e POSTGRES_USER=sysknife -e POSTGRES_PASSWORD=sysknife \
  -e POSTGRES_DB=sysknife_test -p 5432:5432 postgres:17-alpine
```

## Difficulty

`easy`. Two env reads and an attribute, plus one line in the workflow. The care is in picking a default that does not turn every contributor's local run red.

## Getting started

vsolano9:
Closes #294.

`postgres-contract` is one of the five required checks on `main`. Its job is to prove the live PostgreSQL store. On current `main` (`6ca1467`) five of the seven tests in `postgres_store.rs` return early when `SYSKNIFE_TEST_POSTGRES_URL` is unset, so a renamed variable, a dropped `services:` block, or a dead health check still reports success. nextest prints per-test stdout only for failures, so the existing `eprintln!` is invisible in exactly that run.

The issue was filed against five tests, three of which skipped. Two more live-contract tests have landed since; this applies the same fail-closed skip to all five. The two URL-policy tests never needed a server and still run unconditionally.

### Change

- **`resolve_live_postgres_url`** returns the URL, `None` to skip, or panics when `SYSKNIFE_REQUIRE_POSTGRES` is set without a URL.
- The five live tests are `#[ignore]`, so a local run without a database reports `5 ignored` rather than `passed`.
- The `postgres-contract` job sets `SYSKNIFE_REQUIRE_POSTGRES=1` and runs with `--include-ignored`, so a broken service block fails the job instead of skipping it. `scripts/ci-local.sh` mirrors that when it actually starts Postgres.
- Workspace baseline moves `1,775 → 1,774`: nextest's `Starting N` count excludes ignored tests (`+4` helper tests, `-5` live tests now skipped in the workspace job).

### Proof (darwin/arm64, no Postgres)

Issue commands, current `main` first:

```
cargo test -p sysknife-daemon --test postgres_store --locked
# 7 passed; 0 failed; 0 ignored

SYSKNIFE_REQUIRE_POSTGRES=1 cargo test -p sysknife-daemon --test postgres_store --locked
# 7 passed; 0 failed; 0 ignored     ← the bug
```

This branch:

```
cargo test -p sysknife-daemon --test postgres_store --locked
# 6 passed; 0 failed; 5 ignored

SYSKNIFE_REQUIRE_POSTGRES=1 cargo test -p sysknife-daemon --test postgres_store --locked
# 5 passed; 1 failed; 5 ignored
# require_postgres_fails_closed_when_the_url_is_missing panics

SYSKNIFE_REQUIRE_POSTGRES=1 cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
# 5 passed; 6 failed; 0 ignored
# the five live tests panic too
```

Helper tests (env-free): missing URL is `None`, a configured URL is returned even when required, required-and-missing panics. The last of those was red against a stub that returned `url` unchanged, then green.

`cargo fmt --all --check` is clean. I did not dispatch a hosted run. The live contract with a real server still belongs to the `postgres-contract` job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,6 +254,7 @@ jobs:
     env:
       SYSKNIFE_TEST_POSTGRES_URL: >-
         postgres://sysknife:sysknife@127.0.0.1:5432/sysknife_test?sslmode=disable
+      SYSKNIFE_REQUIRE_POSTGRES: "1"
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
@@ -265,7 +266,7 @@ jobs:
         uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6  # v2.9.2
 
       - name: Run live Postgres migration and store contract
-        run: cargo test -p sysknife-daemon --test postgres_store --locked
+        run: cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
 
   security-audit:
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,21 @@ Releases before `0.2.5` predate the public launch; their notes live in the
 
 ## [Unreleased]
 
+### Fixed
+
+- **`postgres-contract` reported success when no database was configured.**
+  ([#313](https://github.com/lacs-project/sysknife/pull/313),
+  [#294](https://github.com/lacs-project/sysknife/issues/294))
+  Five tests in `postgres_store.rs` returned early when
+  `SYSKNIFE_TEST_POSTGRES_URL` was unset, so a renamed variable or a dropped
+  `services:` block left one of `main`'s five required checks green over a
+  contract it never ran. nextest prints per-test stdout only on failure, so the
+  existing notice was invisible in exactly that run. The five live tests are now
+  `#[ignore]`, the job sets `SYSKNIFE_REQUIRE_POSTGRES` and runs with
+  `--include-ignored`, and a required-but-unconfigured run panics instead of
+  passing. A local run without a database reports them ignored rather than
+  passed.
+
 ## [0.10.0] — 2026-08-26
 
 The middle digit moves because [#289](https://github.com/lacs-project/sysknife/pull/289)
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,775 Rust tests and 72 frontend tests** form the current deterministic
+**1,774 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/crates/sysknife-daemon/tests/postgres_store.rs
+++ b/crates/sysknife-daemon/tests/postgres_store.rs
@@ -10,7 +10,64 @@ use sysknife_daemon::transactions::NewTransaction;
 use sysknife_types::{CallerRole, JobState, PreviewEnvelope, RequestHash, RiskLevel};
 
 fn test_url() -> Option<String> {
-    std::env::var("SYSKNIFE_TEST_POSTGRES_URL").ok()
+    std::env::var("SYSKNIFE_TEST_POSTGRES_URL")
+        .ok()
+        .filter(|url| !url.is_empty())
+}
+
+fn postgres_is_required() -> bool {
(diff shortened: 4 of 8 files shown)


vladimirrott (dismissed):
The issue named three skipping tests. You checked, found five, and said why the count had moved. That is the part I would have had to redo myself if you had taken the number on trust.

Verified at `8d86437` on Ubuntu.

`#[ignore]` plus a green job is the shape of a contract that stopped running without telling anyone, so I read the hosted `postgres-contract` log rather than the tick:

```
running 11 tests
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Six non-live and your five live tests, against the service container. That is the number that makes the rest of the PR trustworthy.

Baseline arithmetic matches what you wrote: four helpers in, five live tests out of the nextest count, 1,775 to 1,774, and all three prose figures move with the artifact.

One finding, and it is not a blocker. The fail-closed path rests on `SYSKNIFE_REQUIRE_POSTGRES`, but `--include-ignored` is what puts the five live tests back into the run. Remove the flag and the job passes having run none of them:

```
SYSKNIFE_REQUIRE_POSTGRES=1 SYSKNIFE_TEST_POSTGRES_URL=postgres://…:5432/… \
  cargo test -p sysknife-daemon --test postgres_store --locked
# test result: ok. 6 passed; 0 failed; 5 ignored
# exit 0, and no server was running
```

So #294's shape survives on a different trigger, and no test in the repo reads that job's command. Your change still closes the wider hole: the old skip fired on an unset variable, which a rename or an env-inheritance change reaches without anyone touching the job, while this one needs an edit to the line whose whole purpose is running the contract. A `tests/release/` coherence check holding the job to both the variable and the flag would finish it, in the same idiom as the manifest checks already in `docs-and-hygiene`. That belongs in its own issue rather than another round here.

Merging. The CHANGELOG entry is mine to write.

If you want the next one, #221 sits close to this: `audit verify` lets an inconclusive anchor mask a broken chain, and `--json` status ignores the anchor. Same defect class you have now fixed twice.

vladimirrott (approved):
Re-approving after my own CHANGELOG commit `4dc1772`, which `dismiss_stale_reviews` cleared. The board is green on that head, including `postgres-contract` reporting `11 tests … 0 ignored` again, so the live contract still runs after the entry landed.

Issue #234, opened 2026-08-17
UfwDeleteRule needs a rule number that no action in the catalogue can produce


vladimirrott:
> **Line references re-measured 2026-09-05 at `7d19864`.** `command_mechanism("sudo", ["ufw", "status", "verbose"])` sits at line 127 now, not 124.
> The defect is unchanged.

`UfwDeleteRule` is indexed by a rule number, and no action in the catalogue can produce one.

The parameter's own doc says where the value comes from (`crates/sysknife-daemon/src/actions/ufw.rs:137`):

```rust
/// `rule_number` is the 1-based integer shown by `ufw status numbered`.
```

The catalogue description repeats it to the planner, as "positive integer from 'ufw status numbered'". The only ufw read action runs the other rendering (`crates/sysknife-daemon/src/actions/ufw.rs:127`):

```rust
mechanism: command_mechanism("sudo", ["ufw", "status", "verbose"]),
```

`grep -rn numbered crates/sysknife-daemon/src/actions/ufw.rs` returns the doc comment and nothing else. `status verbose` prints no rule numbers; ufw renders them only for `status numbered`.

## Why it matters

The documented workflow is a dead end. A planner asked to delete a firewall rule has no way to obtain the index, so it either guesses an index into a High-risk firewall deletion or tells the operator to leave SysKnife and run ufw by hand. Guessing is the worse branch: rule numbers shift as rules are added and removed, so a stale index deletes a different rule than the one the operator approved.

## Scope

Either shape works:

- Add an optional `numbered: bool` to `UfwStatus`, so one action serves both renderings.
- Or add a `UfwStatusNumbered` read action.

The existing `/usr/sbin/ufw` grant in `packaging/sysknife-sudoers` already covers the argv, so no new privilege is needed. Whichever you pick, the new capability needs its `KNOWN_ACTIONS` entry and its risk level, and `docs/action-reference.md` regenerates.

~~A story would be worth adding in the same PR, since `UfwDeleteRule` currently has none.~~ **Correction, 2026-09-07 at `adab560`: it has one, and has since #206.** `tests/e2e/stories/story-125.sh` drives the intent "delete ufw rule number 3". Read it before you change the action: it is the thing your change has to keep passing, and it was already there when this issue said otherwise.

```
$ ls tests/e2e/stories/story-125.sh && sed -n '1,7p' tests/e2e/stories/story-125.sh
tests/e2e/stories/story-125.sh
#!/usr/bin/env bash
[...]

PR #314, opened 2026-08-28
fix(daemon): point CheckPendingReboot at Ubuntu's reboot-required.pkgs

### LINKED ISSUE #227: CheckPendingReboot reads reboot-required-pkgs, and fails exactly when a reboot is pending

vsolano9:
Closes #227.

`CheckPendingReboot` cat'd `/var/run/reboot-required-pkgs`. Ubuntu's `update-notifier` writes the dot form:

```
$ grep -oh 'reboot-required[.-]pkgs' /usr/share/update-notifier/notify-reboot-required
reboot-required.pkgs
```

The then-branch is `cat $sentinel; cat $pkgs 2>/dev/null`. When a reboot is pending the first `cat` succeeds and the second fails, so the `if` block's exit status is 1. The dispatcher is not in the informational-exit whitelist (`GetServiceStatus` 1..=4 only), so it returns `CheckPendingReboot failed with exit code 1` and drops stdout. The action fails in the case it exists for.

The constructor comment already told callers to treat stdout as authoritative. That advice is not what the dispatcher does.

### Change

- Constant is `/var/run/reboot-required.pkgs`.
- Then-branch ends with `true`, so a missing packages file is optional rather than an error. (The issue allowed this or a dispatcher whitelist entry; the script-level fix means every execution path, not only `handle_query_action`, sees exit 0 when the sentinel is present.)
- Regenerated `docs/action-reference.md` from the catalogue.

Left the unverified LivepatchStatus / Fail2banStatus / CloudInitStatus leads out of this PR.

Added 4 tests. Per CONTRIBUTING I did not chase `tests/evidence/workspace-tests.json` (1,775 on `main`); whoever merges regenerates the figure.

### Proof (darwin/arm64)

Red against unmodified `main` (`6ca1467`):

```
cargo test -p sysknife-daemon --locked --lib actions::reboot
# 9 passed; 2 failed
# check_pending_reboot_script_uses_ubuntu_dot_pkgs_filename
#   hyphen form still in argv
# pending_reboot_with_sentinel_only_is_not_an_error
#   left: 1  right: 0
#   stdout="*** System restart required ***\n"
```

Green on this branch:

```
cargo nextest run -p sysknife-daemon --locked --lib actions::reboot
# 11 passed, 864 skipped

cargo test -p sysknife-daemon --locked --test action_reference_doc
# 1 passed

cargo nextest run -p sysknife-daemon --locked --test query_action_exit_codes --test action_consistency --test action_reference_doc
# 17 passed

cargo fmt --all --check
# clean

python3 scripts/check_evidence_claims.py
# Published figures match the evidence artifacts.
```

Mutations after the fix, then restore:

- drop trailing `true` → `pending_reboot_with_sentinel_only_is_not_an_error` fails (exit 1, stdout still the restart line)
- restore hyphen constant → `check_pending_reboot_script_uses_ubuntu_dot_pkgs_filename` fails

I did not dispatch a hosted run. Workspace `cargo nextest run --workspace --locked` does not compile on this Mac: `SocketTarget::Vsock` in `apps/sysknife-cli/src/runner.rs:114` is not `cfg(target_os = "linux")` while the variant is. That file is not in this diff; same error on unmodified `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,6 +254,7 @@ jobs:
     env:
       SYSKNIFE_TEST_POSTGRES_URL: >-
         postgres://sysknife:sysknife@127.0.0.1:5432/sysknife_test?sslmode=disable
+      SYSKNIFE_REQUIRE_POSTGRES: "1"
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
@@ -265,7 +266,7 @@ jobs:
         uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6  # v2.9.2
 
       - name: Run live Postgres migration and store contract
-        run: cargo test -p sysknife-daemon --test postgres_store --locked
+        run: cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
 
   security-audit:
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,19 @@ Releases before `0.2.5` predate the public launch; their notes live in the
 
 ## [Unreleased]
 
+### Fixed
+
+- **`CheckPendingReboot` failed exactly when a reboot was pending.**
+  ([#227](https://github.com/lacs-project/sysknife/issues/227))
+  The action read `/var/run/reboot-required-pkgs`, a filename Ubuntu does not
+  write. `update-notifier` writes `reboot-required.pkgs`. When the sentinel
+  was present the missing `cat` became the script's exit status, the
+  dispatcher dropped stdout, and the operator saw
+  `CheckPendingReboot failed with exit code 1` instead of
+  `*** System restart required ***`. The action now reads the dot filename
+  and ends the then-branch with `true`, so a missing packages file is
+  optional rather than an error.
+
 ## [0.10.0] — 2026-08-26
 
 The middle digit moves because [#289](https://github.com/lacs-project/sysknife/pull/289)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ Releases before `0.2.5` predate the public launch; their notes live in the
 ### Fixed
 
 - **`CheckPendingReboot` failed exactly when a reboot was pending.**
-  ([#227](https://github.com/lacs-project/sysknife/issues/227))
+  ([#314](https://github.com/lacs-project/sysknife/pull/314),
+  [#227](https://github.com/lacs-project/sysknife/issues/227))
   The action read `/var/run/reboot-required-pkgs`, a filename Ubuntu does not
   write. `update-notifier` writes `reboot-required.pkgs`. When the sentinel
   was present the missing `cat` became the script's exit status, the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,19 @@ Releases before `0.2.5` predate the public launch; their notes live in the
 
 ### Fixed
 
+- **`postgres-contract` reported success when no database was configured.**
+  ([#313](https://github.com/lacs-project/sysknife/pull/313),
+  [#294](https://github.com/lacs-project/sysknife/issues/294))
+  Five tests in `postgres_store.rs` returned early when
+  `SYSKNIFE_TEST_POSTGRES_URL` was unset, so a renamed variable or a dropped
+  `services:` block left one of `main`'s five required checks green over a
+  contract it never ran. nextest prints per-test stdout only on failure, so the
+  existing notice was invisible in exactly that run. The five live tests are now
+  `#[ignore]`, the job sets `SYSKNIFE_REQUIRE_POSTGRES` and runs with
+  `--include-ignored`, and a required-but-unconfigured run panics instead of
+  passing. A local run without a database reports them ignored rather than
+  passed.
+
 - **`CheckPendingReboot` failed exactly when a reboot was pending.**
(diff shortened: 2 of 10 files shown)


vladimirrott:
I checked the filename against a machine rather than against the issue:

```
$ grep -oh 'reboot-required[.-]pkgs' /usr/share/update-notifier/notify-reboot-required
reboot-required.pkgs
```

Your reason for fixing the script instead of adding a dispatcher whitelist entry holds up in both readers: `dispatcher.rs:1827` turns the non-zero exit into `execution_failure` on the query path, and `dispatcher.rs:2692` marks the job `Failed` on the execution path. A whitelist entry fixes the first and leaves the second. The trailing `true` fixes both.

Gates at `b25be48`, Ubuntu:

- `cargo fmt --all --check` clean
- `cargo clippy --workspace --all-features --all-targets --locked -- -D warnings` clean
- `cargo nextest run --workspace --locked`: 1779 passed, 0 skipped

I re-ran your two mutations and added one of my own:

| mutation | what fails |
| --- | --- |
| hyphen constant restored | `check_pending_reboot_script_uses_ubuntu_dot_pkgs_filename` |
| trailing `true` removed | `pending_reboot_with_sentinel_only_is_not_an_error` |
| script edited, `docs/action-reference.md` left alone | `action_reference_doc_is_current` |

Each breaks one test and no others, and the tree restored clean afterwards.

The first row is worth a note. With the hyphen back, the older `check_pending_reboot_script_references_pkgs_file` still passes, because it asserts the constant reaches argv and never which filename it holds. That is how the wrong name travelled from the gap-analysis table into the catalogue and sat there. Your test is what pins it.

The only red is the baseline, and you read CONTRIBUTING correctly: that one is mine. #313 lands first and moves the count to 1,774, so yours becomes 1,778. I will regenerate the artifact and the three prose figures on your branch, wait for the run, and merge.

Optional, and I am content either way: `docs/ubuntu-action-gap-analysis.md:138` still carries the hyphen form. It is the proposal table this bug came out of, so keeping it as a record of what was proposed is defensible. Say if you would rather it match and I will take it at merge.

Next one if you want it: #234. `UfwDeleteRule` needs a rule number no action in the catalogue can produce, which is the same "the code asks for something the system never supplies" shape as this.

vladimirrott:
Pushed `de61ce6` to your branch, since the baseline is mine to carry.

#313 merged first and took the count to 1,774, so I merged `main` in, regenerated the artifact, and moved the three prose figures with it. Measured rather than assumed:

```
test_baseline: recorded 1778 rust tests in tests/evidence/workspace-tests.json
```

1,774 from `main` plus your four. `README.md`, `docs/introduction.md` and `docs/distro-support.md` now read 1,778, and `scripts/check_evidence_claims.py` is clean.

One conflict, in `CHANGELOG.md`: your entry and #313's both open `### Fixed` under `## [Unreleased]`. I kept both, #313's first because it landed first. Your wording is unchanged.

`scripts/ci-local.sh` passed locally on the merged tree before the push. Waiting on the hosted run, then I will approve and merge.

vladimirrott (approved):
Approving at `de61ce6`. The board is green on the merged head, and `rust` confirms the regenerated figure:

```
test_baseline: 1778 rust tests, matching the recorded baseline.
```

Your fix, tests and CHANGELOG wording are unchanged; the merge commit only carries the baseline and the conflict resolution.

Issue #226, opened 2026-08-17
query_action runs AptUpdate, a root apt mutation, with no preview and no approval receipt


vladimirrott:
`handle_query_action` is the read-only fast path: no preview, no approval receipt, no exclusion gate. Its admission test is a risk tier (`crates/sysknife-daemon/src/dispatcher.rs:1684`):

```rust
if min_role != CallerRole::Observer {
    return send_error(framed, request_id, "authorization_failure",
        format!("{action_name} is not a read-only action; use preview+execute instead")).await;
}
```

The role comes from the risk level (`crates/sysknife-daemon/src/policy.rs:38`), and `AptUpdate` is Low, so Observer (`crates/sysknife-daemon/src/actions/apt.rs:119`):

```rust
mechanism: command_mechanism("sudo", ["env", DEBIAN_FRONTEND_VALUE, NEEDRESTART_MODE_VALUE, APT_GET, "update"]),
risk_level: RiskLevel::Low,
```

It then runs unconditionally (`dispatcher.rs:1807`):

```rust
let output = match executor.execute(&spec).await {
```

Note also that `handle_query_action` takes no `caller` argument at all (`dispatcher.rs:1658`), so the tier is the only gate.

## Why it matters

The lowest-privilege caller, including one whose `SO_PEERCRED` lookup failed and resolved to `Observer` with `none:unattributed`, can make the daemon run `sudo apt-get update` as root. That writes under `/var/lib/apt/lists` and takes the dpkg lock, with no preview, no receipt and no audit of an approval that never happened.

The codebase already knows the classifier is wrong here. `dispatcher.rs:805` says so in as many words:

> `AptUpdate` is `RiskLevel::Low`, so `Observer`, so "read-only" by that measure, while it runs `sudo apt-get update` as root, writes under /var/lib/apt/lists and takes the dpkg lock.

That comment sits above the *platform* fence, which was fixed. The approval fence was not.

Second effect: `query_action` never reaches the exclusion gate, so this can grab the dpkg lock while a high-risk action holds `ExclusiveResource::System` and `Dpkg`, which is the interleaving that gate exists to prevent.

[...]

PR #316, opened 2026-08-29
fix(e2e): keep provider keys out of VM process arguments

### LINKED ISSUE #299: The VM harnesses put LLM API keys in the host and guest process tables
Both VM harnesses forward LLM provider keys by putting them on a command line:

```bash
# tests/e2e/atomic-vm.sh:320 and :465, tests/e2e/ubuntu-vm.sh:502 and :522
prov_env+=" $var='$val'"
...
cmd_ssh "cd /home/${VM_USER}/sysknife && sudo${prov_env} bash tests/e2e/provision.sh"
```

The comment above it explains why: `sudo VAR=val cmd` injects the variable regardless of `env_reset`. That part works. The cost is that the value lands in two process tables and stays there for the life of the command.

### Read back, not theorised

Provisioning the Fedora guest, as the unprivileged `lacsdev` user inside it:

```
$ ps -ef | grep sudo
root  2638  2637  0 18:35 ?  sudo OPENAI_API_KEY=sk-proj-<48 chars redacted> ... bash tests/e2e/provision.sh
```

And on the host, as a normal user, the same key appears in the `ssh` argv. Neither `/proc` is mounted with `hidepid`, and `/proc/PID/cmdline` is mode `444` on both:

```
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
-r--r--r--. 1 root root 0 /proc/1/cmdline
```

So any local account on the contributor's laptop, and any account in the guest, can read a production LLM key for as long as the command runs. `run-stories.sh` over 54 stories runs for the better part of an hour, and it forwards seven keys, not one.

### Scope

Contributor tooling, not the product. The daemon and CLI do not do this. It matters because the people running these scripts are the ones with real keys, and because a shared or multi-user workstation is ordinary.

`atomic-vm.sh` `cmd_provision` and `cmd_run`; `ubuntu-vm.sh` at both sites. Fixing one and not the other is the failure this repo keeps having, so fix all four.

### A shape that works

Write the variables into the guest as a `0600` file, source it in the child, delete it after:

```bash
printf '%s\n' "${assignments[@]}" \
  | cmd_ssh "umask 077 && cat > /tmp/sysknife-e2e.env"
cmd_ssh "cd ... && sudo bash -c 'set -a; . /tmp/sysknife-e2e.env; set +a; \
    rm -f /tmp/sysknife-e2e.env; exec bash tests/e2e/provision.sh'"
```

Nothing sensitive reaches an argv, and the file is gone before the script body runs. `sshd`'s `AcceptEnv` plus `sudo --preserve-env` is the other route, and it needs a change in `cmd_bootstrap`, so it is a larger diff for the same result.

### Proving it

A host-side test in the shape of `provider-parity.test.sh`: derive the forwarded variable names from `crates/sysknife-brain/src/config.rs` as that test already does, then assert no harness interpolates one into a command string. Grepping for the assignment pattern is enough and needs no VM.

Whoever takes this: say in the PR how you checked that no argv carries a key any more, not just that the tests pass. A run that reads its own `ps` output is the convincing evidence.

xianjianlf2:
## Summary
- serialize the explicit VM environment allow-lists with shell-safe quoting
- transfer them through SSH stdin into a root-only `/run` file instead of interpolating values into SSH and sudo commands
- source and remove the file before execing provision or story scripts, with failure-path cleanup
- apply the fix to provision and run in both Fedora Atomic and Ubuntu harnesses
- derive provider keys from the Rust configuration in a new host-side regression test, wired into local and GitHub CI

## Security check
The regression check rejects the old assignment/`sudo${...}` argv patterns and requires stdin staging plus pre-exec deletion at all four call sites. A repository search confirms no harness command string interpolates the forwarded values anymore.

## Verification
- `bash tests/e2e/vm-env-secrets.test.sh`
- `bash -n tests/e2e/atomic-vm.sh tests/e2e/ubuntu-vm.sh tests/e2e/vm-env-secrets.test.sh scripts/ci-local.sh`
- parsed `.github/workflows/ci.yml` with Ruby YAML
- `git diff --check`

The existing `provider-parity.test.sh` and `vm-sync-parity.test.sh` require Bash 4 `mapfile` and cannot run under the host macOS Bash 3.2; the new test is Bash 3.2-compatible and those existing checks remain wired for Linux CI. No VM image was available for a live guest run.

Closes #299

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,9 @@ jobs:
       - name: Check provider parity across E2E entry points
         run: bash tests/e2e/provider-parity.test.sh
 
+      - name: Check VM harnesses keep provider keys out of process arguments
+        run: bash tests/e2e/vm-env-secrets.test.sh
+
       - name: Check story metadata is derived from the story files
         run: bash tests/e2e/story-metadata.test.sh
 
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -279,6 +279,7 @@ run_hygiene_group() {
     run_step 'hygiene: systemd-directory-modes.test.sh' bash "$repo_root/tests/release/systemd-directory-modes.test.sh"
     run_step 'hygiene: ubuntu-vm-bootstrap.test.sh' bash "$repo_root/tests/e2e/ubuntu-vm-bootstrap.test.sh"
     run_step 'hygiene: provider-parity.test.sh' bash "$repo_root/tests/e2e/provider-parity.test.sh"
+    run_step 'hygiene: vm-env-secrets.test.sh' bash "$repo_root/tests/e2e/vm-env-secrets.test.sh"
     run_step 'hygiene: story-metadata.test.sh' bash "$repo_root/tests/e2e/story-metadata.test.sh"
     run_step 'hygiene: vm-sync-parity.test.sh' bash "$repo_root/tests/e2e/vm-sync-parity.test.sh"
     run_step 'hygiene: docs-share-cards.test.sh' bash "$repo_root/tests/release/docs-share-cards.test.sh"
--- a/tests/e2e/atomic-vm.sh
+++ b/tests/e2e/atomic-vm.sh
@@ -104,6 +104,7 @@ VM_SUBDIR="${VM_DIR}/${VM_NAME}"
 # contributor's personal ~/.ssh/id_* keys because those are typically
 # passphrase-protected, which breaks rsync/non-interactive ssh.
 SSH_KEY="${SYSKNIFE_VM_SSH_KEY:-$HOME/.ssh/sysknife-vm}"
+GUEST_ENV_FILE="/run/sysknife-e2e.env"
 
 ssh_opts() {
     printf -- '-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR -i %s -o IdentitiesOnly=yes' "$SSH_KEY"
@@ -115,6 +116,21 @@ ssh_opts() {
 
 log() { printf '[atomic-vm] %s\n' "$*" >&2; }
 die() { log "ERROR: $*"; exit 1; }
+
+write_guest_env_file() {
+    local assignment value var
+    local assignments=()
+    for var in "$@"; do
+        eval "value=\${$var-}"
+        if [ -n "$value" ]; then
+            printf -v assignment '%s=%q' "$var" "$value"
+            assignments+=("$assignment")
+        fi
+    done
+
+    printf '%s\n' "${assignments[@]}" \
+        | cmd_ssh "sudo sh -c 'umask 077 && cat > ${GUEST_ENV_FILE}'"
+}
 
 require_tools() {
     local missing=()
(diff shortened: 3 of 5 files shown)


vladimirrott:
Your `docs-and-hygiene` is red and it is not your change.

`https://registry.modelcontextprotocol.io` stopped answering today at about 15:00 UTC. It still resolves and the connection times out, from the GitHub runners and from my machine:

```
$ getent hosts registry.modelcontextprotocol.io
34.61.200.254   prod.registry.modelcontextprotocol.io registry.modelcontextprotocol.io

$ curl -o /dev/null --max-time 20 https://registry.modelcontextprotocol.io
curl: (28) Connection timed out after 20002 milliseconds
```

The `Check markdown links` step retries three times and then fails on `README.md`, which is a file your branch does not touch. One step of thirty-three failed; everything ahead of it in that job passed, and every other job on your PR is green.

I ran the same checker with the repo's config over all ten files CI links: that URL is the only dead link anywhere. `main` fails identically, so there is nothing to fix here and nothing you can do about it from your side.

No action needed from you. Your review notes still stand on their own and I will pick those up separately; this comment is only so a red board does not read as a problem with your work.

vladimirrott (approved):
Verified at `c34e906`.

The fix does more than the issue asked for. Passing values through `%q` into a sourced file also closes a command-injection hole in the old code: `prov_env+=" $var='$val'"` broke apart on any value containing a single quote, so a key with an apostrophe in it became shell syntax on the guest. That path is gone.

Three details I checked and liked:

- `umask 077` runs before the redirection inside the same `sh -c`, so the file is 0600 from creation rather than chmod-ed after a readable moment.
- The writer is `sh -c` but only runs `umask` and `cat`; the consumer is `bash -c`, which is the shell that can parse the `$'...'` forms `printf %q` emits. Pairing those the other way round would misparse a value carrying a control character.
- `trap "rm -f …" EXIT; …; rm -f …; trap - EXIT; exec …` covers both a failure during sourcing and the fact that `exec` never runs an EXIT trap. Both halves are needed and both are there.

I checked the argv side by measurement rather than by reading. I stubbed `cmd_ssh` to record its arguments and ran both `provision` and `run` with canary values in all seven key variables. On `main`, four canaries land in the ssh argv. On this branch, zero.

Two gaps in the new guard, neither blocking, both mine to fix rather than yours:

1. `grep -Fq "$key" "$harness"` asks whether a key appears anywhere in the file. Both harnesses name each key twice, once for provision and once for run, so deleting it from `cmd_run` alone leaves the other copy and the check stays green. I mutated exactly that and it passed.
2. `grep -c 'rm -f \${GUEST_ENV_FILE}.*exec bash tests/e2e/'` expects two matches, but the `trap "rm -f ${GUEST_ENV_FILE}" EXIT` earlier on the same line already satisfies the pattern on its own. Delete the explicit pre-`exec` `rm` and the count is still two. Since `trap - EXIT` runs first, nothing deletes the file, and every provider key stays readable at `/run/sysknife-e2e.env` for the rest of the VM's life. The failure message says the check catches that.

Five of my seven mutations do bite: reverting the stdin transfer, dropping `umask 077`, collapsing to one `write_guest_env_file` call site, removing a key from both places, and renaming `*_API_KEY` in `config.rs` so the extraction returns nothing.

Merging. Closes #299.

If you want another in the same area, #226 is the closest fit: `query_action` runs `AptUpdate`, a root apt mutation, with no preview and no approval receipt. Same shape as what you just fixed, a privileged path that skips the control everything else goes through.

Issue #248, opened 2026-08-18
exclusive_resource has an arm for a helper name no installer produces


vladimirrott:
> **Line references re-measured 2026-09-05 at `7d19864`.** #360 added eight lines above the matcher, so the unreachable arm is at `actions/mod.rs:161` now, not 153.
> The defect is unchanged.

`exclusive_resource` maps an action's argv program to the lock it must hold. One arm cannot match (`crates/sysknife-daemon/src/actions/mod.rs:161`):

```rust
| "unattended-upgrade"
| "sysknife-apt-pin-edit"
| "apt-pin-edit" => Some(ExclusiveResource::Dpkg),
```

`sysknife-apt-pin-edit` is the filename in `packaging/`. Both installers strip the prefix:

```
Makefile:80                    install -Dm 755 packaging/sysknife-apt-pin-edit $(HELPERS)/apt-pin-edit
tests/e2e/ubuntu-provision.sh  install -Dm 755 packaging/sysknife-apt-pin-edit /usr/lib/sysknife/apt-pin-edit
```

The only emitted command is `/usr/lib/sysknife/apt-pin-edit` (`crates/sysknife-daemon/src/actions/apt_preferences.rs:11`), so the token reaching the matcher is always `apt-pin-edit`. All twelve sibling helpers install stripped and none carries a second arm.

## Why bother

The arm reads as though the helper might appear under either name, which invites the next person adding a helper to copy the pattern. Deleting it makes the matcher say what is true: one installed name, one arm.

Nothing depends on it, so this is a one-line deletion with no behaviour change.

## Tests first

No test names the prefixed spelling, which is how it survived. If you want the deletion to stay deleted, assert that every helper in `packaging/` appears in the matcher under its installed basename and not its source filename. That derives the rule instead of restating it, and it would catch the same mistake in a future helper.

## Difficulty

`easy`. Good first issue: one line to remove, and the optional guard is a small loop over `packaging/`.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands. No CLA and no copyright waiver. The project is MIT.

vsolano9:
I'll take this. I'll remove the dead source-name arm, verify the current matching tests still cover the installed basename, and keep the patch focused.

vladimirrott:
Checking in, with nothing attached to it. You claimed #248 on 26 August and I have not seen a branch, which is fine; the issue is not going anywhere.

[...]

PR #317, opened 2026-09-01
test(ci): fail if postgres-contract drops a required token

### LINKED ISSUE #315: postgres-contract still goes green on zero live tests when --include-ignored is dropped
#313 closed the case where `postgres-contract` reported success with no database configured. It moved the trigger rather than removing it.

The job now sets `SYSKNIFE_REQUIRE_POSTGRES` (`.github/workflows/ci.yml:257`) and runs with `--include-ignored` (`:269`). Both are load-bearing, and only one of them fails loudly when it goes missing.

Drop the flag, change nothing else, on `main` at `ecf8c7d`:

```
SYSKNIFE_REQUIRE_POSTGRES=1 \
SYSKNIFE_TEST_POSTGRES_URL=postgres://sysknife:sysknife@127.0.0.1:5432/sysknife_test?sslmode=disable \
  cargo test -p sysknife-daemon --test postgres_store --locked
# test result: ok. 6 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out
# exit 0
```

No server was running. The five live tests carry `#[ignore]`, so without the flag they never execute; `require_postgres_fails_closed_when_the_url_is_missing` passes because the URL is set; and one of `main`'s five required checks reports success over a contract it did not run.

That is #294's shape with a different cause. The variable is guarded from inside the test file. The flag is guarded from nowhere.

### What to build

A check that reads the job and fails when either half goes missing.

`tests/release/release-rehearsal.test.sh` already greps workflow YAML with `grep -Fq`, so the mechanism has a precedent here. `tests/e2e/provider-parity.test.sh` is the better model for how to write it: it derives the list it checks from the source rather than restating it, and its header explains why a partial check that reads like a full one is worse than no check at all.

Cover `scripts/ci-local.sh` as well. It carries the same pair twice (`:347` and `:390`), and the local board is what contributors run before they push. A guard that watches CI and ignores the local mirror leaves half the surface open.

Wire it into `docs-and-hygiene` beside the other release checks, and into `scripts/ci-local.sh`. A test nothing invokes is the same defect in a different costume.

### Prove it bites

Both directions:

- remove `--include-ignored` from `ci.yml`, the check fails and names the file
- remove `SYSKNIFE_REQUIRE_POSTGRES`, the check fails and names the file
- restore both, green

Then break the check's own input: point it at a workflow path that does not exist. It has to fail rather than pass over nothing to read. A guard that inspects an empty set and reports success is how this class of defect survives in the first place.

### Scope

`tests/release/`, the wiring step in `.github/workflows/ci.yml`, and `scripts/ci-local.sh`. No Rust in the diff, so the workspace gate does not apply; `CONTRIBUTING.md` has the one-line check for that. `shellcheck` covers `tests/**` and `scripts/**` in the `scripts-lint` job.

### Difficulty

Small. The thinking is in deciding what the check reads, not in writing it.

vsolano9:
## Summary

#313 closed the case where `postgres-contract` reported success with no
database configured. The job now sets `SYSKNIFE_REQUIRE_POSTGRES` and runs
with `--include-ignored`. Both are load-bearing; only the variable fails
loudly from inside the test file.

Drop the flag, change nothing else, and cargo reports the ignored live
tests and exits 0. Same shape as #294, different cause.

This check reads the job and both `ci-local.sh` branches (exported-URL
path and container-start path) and fails when either token is missing.
It does not assert a test count.

## Related Issue

Closes #315

## Validation

- [x] Tests added or updated
- [x] Documentation updated if behavior changed (N/A — CI hygiene only)
- [x] Security impact considered (none; host-side greps)
- [ ] CI passes

Local: `bash tests/release/postgres-contract-guard.test.sh` (exit 0);
`shellcheck --severity=warning tests/release/postgres-contract-guard.test.sh scripts/ci-local.sh`;
`yamllint .github/workflows/ci.yml`.
Mutation: dropping either token from `ci.yml` or either `ci-local.sh` branch
fails and names the file; a missing workflow path fails rather than passing
over nothing; restore is green.

No Rust in the diff, so the workspace nextest gate does not apply.

## Notes for Reviewers

Tokens are derived from `postgres_store.rs` (`SYSKNIFE_REQUIRE_POSTGRES` plus
a floor that `#[ignore]` still exists). `--include-ignored` is cargo's flag
for that attribute.

`ci-local.sh` is two branches, not one. The container path puts the env
assignment and `run_step` on different lines; the check joins continuation
lines so that branch cannot hide a missing `SYSKNIFE_REQUIRE_POSTGRES`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,9 @@ jobs:
       - name: Check the install knows every directory it writes into
         run: bash tests/release/install-paths.test.sh
 
+      - name: Check postgres-contract still requires both job tokens
+        run: bash tests/release/postgres-contract-guard.test.sh
+
       - name: Check provider parity across E2E entry points
         run: bash tests/e2e/provider-parity.test.sh
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,19 @@ Releases before `0.2.5` predate the public launch; their notes live in the
   and ends the then-branch with `true`, so a missing packages file is
   optional rather than an error.
 
+- **`postgres-contract` could still report success on zero live tests.**
+  ([#317](https://github.com/lacs-project/sysknife/pull/317),
+  [#315](https://github.com/lacs-project/sysknife/issues/315))
+  [#313](https://github.com/lacs-project/sysknife/pull/313) made the job fail
+  closed on a missing database, but split the guarantee across two tokens:
+  `SYSKNIFE_REQUIRE_POSTGRES` fails loudly from inside the test file, while
+  `--include-ignored` is what runs the five `#[ignore]` live tests at all.
+  Dropping the flag left the job reporting `6 passed; 5 ignored` and exiting 0
+  with no database touched. `tests/release/postgres-contract-guard.test.sh` now
+  reads the CI job and both `scripts/ci-local.sh` branches and fails when either
+  token goes missing. It derives the variable name from `postgres_store.rs` and
+  asserts no test count, so adding a store test does not break it.
+
 ## [0.10.0] — 2026-08-26
 
 The middle digit moves because [#289](https://github.com/lacs-project/sysknife/pull/289)
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -283,6 +283,7 @@ run_hygiene_group() {
     run_step 'hygiene: vm-sync-parity.test.sh' bash "$repo_root/tests/e2e/vm-sync-parity.test.sh"
     run_step 'hygiene: docs-share-cards.test.sh' bash "$repo_root/tests/release/docs-share-cards.test.sh"
     run_step 'hygiene: install-paths.test.sh' bash "$repo_root/tests/release/install-paths.test.sh"
+    run_step 'hygiene: postgres-contract-guard.test.sh' bash "$repo_root/tests/release/postgres-contract-guard.test.sh"
 
     if have markdownlint-cli2; then
         run_step 'hygiene: markdownlint-cli2' hygiene_markdownlint
--- a/tests/release/postgres-contract-guard.test.sh
+++ b/tests/release/postgres-contract-guard.test.sh
@@ -1,0 +1,132 @@
+#!/usr/bin/env bash
+#
+# postgres-contract-guard.test.sh — both halves of the live Postgres contract
+# must stay on the job that claims to run it.
+#
+# #313 closed the case where postgres-contract reported success with no
+# database configured. The job now sets SYSKNIFE_REQUIRE_POSTGRES and runs
+# with --include-ignored. Both are load-bearing; only the variable fails
+# loudly from inside the test file. Drop the flag and cargo reports
+# "6 passed; 5 ignored" and exits 0 — #294's shape with a different cause.
+# See issue #315.
+#
+# This check reads the job and fails when either token is missing. It does
+# not assert a test count: that would break the next time anyone adds a
+# store test and teach whoever hit it to edit the number rather than read
+# the check.
+#
+# Two design rules, both from tests/e2e/provider-parity.test.sh:
+#
+#   * The tokens are derived from postgres_store.rs, never restated as a
+#     count, so renaming the env var or dropping #[ignore] fails here
+#     until the job catches up.
+#   * It covers every entry point, not just CI. ci-local.sh carries the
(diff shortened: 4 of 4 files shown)


vladimirrott (approved):
You took both traps from the offer and closed them. The `ci-local.sh` handling is the part I would have got wrong: the container branch puts the assignment and `run_step` on separate lines, and joining continuations before the check is what makes that branch impossible to hide.

Verified at `bc12806` on Ubuntu. I ran nine mutations rather than your five, because I wrote the issue and wanted the guard tested against my own blind spots:

| mutation | result |
| --- | --- |
| drop `--include-ignored` from `ci.yml` | fails, names the file |
| drop `SYSKNIFE_REQUIRE_POSTGRES` from `ci.yml` | fails |
| drop the token from the exported-URL branch | fails |
| drop the token from the container branch | fails (I confirmed the edit applied, 2 → 1) |
| rename the job so `extract_job` reads nothing | fails |
| remove `#[ignore]` from `postgres_store.rs` | fails |
| rename the env var in `postgres_store.rs` | fails |
| move `--include-ignored` to a different job | fails, so `extract_job` isolates correctly |
| token present only in a comment inside the job | **passes** |

The last row is the one thing I would change, and not in this PR. `grep -Fq` over the extracted job block is satisfied by the token anywhere in that block, comments included. The shape that reaches it is ordinary: someone comments out the `run:` line while debugging and writes a replacement without the flag. Binding the check to the `run:` line would close it. Say if you want that as a follow-up issue and I will write it up. The guard as it stands catches every case #315 was filed for.

Gates: `shellcheck --severity=warning` over the four trees clean, `yamllint` clean, both wirings present. I also checked the two things a green tick does not tell you, and both hold: `postgres-contract` still logs `running 11 tests`, and `docs-and-hygiene` logs your guard's own pass line, so it executed rather than merely existing.

I pushed `4f07861` with the CHANGELOG entry, since that one is mine to write.

**Not merging today, and the reason is not your PR.** `https://registry.modelcontextprotocol.io` stopped answering between 08:43 and 15:08 today. It resolves (`34.61.200.254`) but the connection times out, from the runners and from here, so the `Check markdown links` step fails all three retries against `README.md`. It has taken `docs-and-hygiene` red on every open PR, including two that touch no markdown at all. Your board is otherwise green. I will merge when the registry answers again; nothing is needed from you.

Next one if you want it: #248 is already yours and still open. After that, #221 is the closest match to what you have now done three times, an `audit verify` where an inconclusive anchor masks a broken chain.

vladimirrott:
Merging with `--admin` over a red `docs-and-hygiene`, and recording why here, because bypassing a required check should leave a trail.

**The failure is one external outage, not this PR.**

`https://registry.modelcontextprotocol.io` stopped answering today between 08:43 and 15:08 UTC. It still resolves; the connection times out:

```
$ getent hosts registry.modelcontextprotocol.io
34.61.200.254   prod.registry.modelcontextprotocol.io registry.modelcontextprotocol.io

$ curl -o /dev/null --max-time 20 https://registry.modelcontextprotocol.io
curl: (28) Connection timed out after 20002 milliseconds
```

The runners see the same thing: `Status: 0`, three attempts, the retry loop in `ci.yml` exhausted.

**What I checked before bypassing.**

One step of thirty-three failed. `docs-and-hygiene` ran twenty-eight checks ahead of it, all green, including this PR's own new guard:

```
success  Check postgres-contract still requires both job tokens
success  Lint markdown
failure  Check markdown links
```

That step aborts on its first file, so `CONTRIBUTING.md` and the eight others were never reached in CI. I ran the same checker with the repo's own config across all ten files locally: one dead link in total, the registry URL, nothing else. `Lint GitHub issue templates`, the step skipped as collateral when the job died, exits 0 locally.

The link lives at `README.md:430`. This PR touches `ci.yml`, `scripts/ci-local.sh`, the new test and `CHANGELOG.md`. `main` at `ecf8c7d` fails the same check for the same reason, so nothing here caused it and nothing here could fix it. Every other check is green, CodeQL included.

**Not adding an ignore pattern for it.** `.markdown-link-check.json` skips three hosts and its own comment explains why each is a checker artefact rather than a broken link: github.com and shields.io anti-bot the checker, npmjs.com rejects its user-agent. This is a content link that is genuinely offline. Muting it would turn a real outage into a silent pass, which is the shape of defect this repo keeps filing issues about. The check stays able to fail.

#316 and #312 are red on the identical step and stay open until the registry answers.

Issue #305, opened 2026-08-25
Four high-severity CodeQL alerts sat unread for 36 days behind 13 Scorecard signals


vladimirrott:
Four `js/clear-text-logging` alerts at **high** severity sat open in `packages/setup/index.js` from 2026-07-20 until today. Thirty-six days, in a repo whose subject is privileged access.

All four were false positives, and I have dismissed them with the reasoning. That is not the defect. The defect is why nobody read them.

### The Security tab is 76% advisory score signal

Open alerts, before the dismissals:

| rule | count | what it is |
| --- | --- | --- |
| `PinnedDependenciesID` | 8 | OSSF Scorecard check |
| `CodeReviewID`, `MaintainedID`, `VulnerabilitiesID`, `FuzzingID`, `CIIBestPracticesID` | 5 | OSSF Scorecard checks |
| `js/clear-text-logging` | 4 | CodeQL, a real analysis |

`scorecard.yml` ends with an `upload-sarif` step, so every Scorecard heuristic becomes a code-scanning alert. Five of them carry no file at all and read as `no file associated with this alert:1`. The workflow's own header calls Scorecard "Advisory — it produces a report and SARIF alerts, never a hard pass/fail gate," so the mixing is deliberate. The consequence was not: four findings that needed twenty minutes of reading were indistinguishable from thirteen that needed none.

### What the four actually were

`API_KEY_VARS` in `packages/setup/providers.js` maps a provider to the **name** of its environment variable:

```js
const API_KEY_VARS = {
  openai:   'OPENAI_API_KEY',
  ...
};
```

The three logging sites interpolate that name, so the output is `OPENAI_API_KEY already set in environment`. CodeQL treats any read of the table as a sensitive source regardless of what the table holds, and the fourth alert is the top-level `console.error(e.message)` reached by the same flow.

The value never goes near a log. `apiKey` has five references: declared at :620, read from `ask()` at :630, written into `sharedEnv` at :716, and tested for emptiness at :957. `sharedEnv` flows only into the MCP config file, which the wizard warns about at :628 in as many words. No `throw` in `packages/setup` interpolates it either; the one error message that carries anything key-adjacent carries `opts.pinnedSha256`, a published checksum.

### The decision

**Drop the `upload-sarif` step and keep `publish_results: true`.** The badge and the public Scorecard API both come from `publish_results`, so nothing visible to a reader is lost, and the Security tab goes back to meaning "an analysis found something in the code". This is my preference.

[...]

PR #326, opened 2026-09-01
ci: move off Node 20, pin the lint tools, and stop Scorecard filling the Security tab

vladimirrott:
Node 20 reached end-of-life on 2026-04-30. Three workflow jobs still pinned it, including the two that run `npm ci` with the workflow token in the environment. All three move to 24, the current active LTS, whose own end-of-life is 2028-04-30.

The pin was also blocking a dependency update. #322 bumps jsdom to 30, which needs Node 22.22 or newer; on 20 the frontend job dies before a single test runs:

```
Caused by: TypeError: webidl.util.markAsUncloneable is not a function
  ❯ new CacheStorage node_modules/undici/lib/web/cache/cachestorage.js:20:17
  ❯ Object.<anonymous> node_modules/jsdom/lib/api.js:12:33
```

`test_baseline --frontend` caught that correctly, refusing to report a count rather than passing a run of zero tests.

## The guard

`tests/release/node-eol.test.sh` reads every `node-version:` under `.github/workflows` and compares its major against the end-of-life dates published in [nodejs/Release](https://github.com/nodejs/Release/blob/main/schedule.json). Comparing against today means the check goes red on its own once a pinned major ages out, instead of waiting for someone to notice.

Two failure modes it handles deliberately:

- A major with no entry in the table fails rather than passing, so the table has to be updated when the project moves to 26.
- Finding no pin at all fails, because that means the extraction broke rather than the workflows being clean.

## Mutation proof

Six mutations, six caught:

| Mutation | Result |
|---|---|
| revert one pin to 20 | `ci.yml:28 pins Node 20, end-of-life since 2026-04-30` |
| pin 18 | `ci.yml:28 pins Node 18, end-of-life since 2025-04-30` |
| pin an unknown major (99) | `pins Node 99, which is absent from the schedule table in this test` |
| pin `lts/*` | `pins an unreadable Node version 'lts/*'` |
| break the grep so it matches nothing | `no node-version pin found …; the extraction is broken, not the workflows` |
| backdate 24's end-of-life in the table | `release-rehearsal.yml:38 pins Node 24, end-of-life since 2020-01-01` |

Wired into `docs-and-hygiene` and `scripts/ci-local.sh`.

## Not in this PR

`packages/setup` declares `engines.node: ">=18"`, and Node 18 has been end-of-life since 2025-04-30. That is a user-facing support claim rather than a CI pin, and raising it drops people, so it needs its own decision. Filing separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020  # v7
         with:
-          node-version: "20"
+          node-version: "24"
 
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
@@ -70,6 +70,9 @@ jobs:
 
       - name: Check the secret scanner catches keys and ignores fixtures
         run: bash tests/release/no-secrets.test.sh
+
+      - name: Check no workflow pins an end-of-life Node release
+        run: bash tests/release/node-eol.test.sh
 
       - name: Check systemd directory modes match tmpfiles
         run: bash tests/release/systemd-directory-modes.test.sh
@@ -173,7 +176,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020  # v7
         with:
-          node-version: "20"
+          node-version: "24"
 
       - name: Install dependencies
         run: npm ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,11 +32,14 @@ jobs:
         with:
           python-version: "3.11"
 
+      # Pinned, because this job holds the workflow token: an unpinned global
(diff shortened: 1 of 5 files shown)


vladimirrott:
Pushed a second commit, same subject: the CI signal a maintainer reads.

**Pinned the lint tools.** `docs-and-hygiene` installed `markdownlint-cli2`, `markdown-link-check` and `yamllint` unpinned, in a job that holds the workflow token. A compromised release of any of the three would have run on the trusted side of the build. All three now name a version, and four of Scorecard's `PinnedDependenciesID` alerts go with them.

**Dropped the SARIF upload from `scorecard.yml`**, which is the decision #305 asked for and stated a preference on. Every Scorecard heuristic was becoming a code-scanning alert, five of them with no file attached. Sixteen new high-severity `rust/cleartext-logging` alerts landed in that same pile this week; all sixteen were false positives, eleven of them in test code, and nothing about the Security tab separated them from the standing score. `publish_results: true` stays, so the badge and the public Scorecard API are unchanged, and the job gives up `security-events: write` because it no longer uploads anything.

Triage of all sixteen is on #305.

Issue #251, opened 2026-08-18
Cancel and the TTL sweep end an approval without revoking it or writing an event


vladimirrott:
`revoke_unconsumed_approval` ends an unspent approval properly: it deletes the row from `transaction_approvals` and appends `ApprovalRevoked` in the same database transaction (`crates/sysknife-daemon/src/transactions.rs:607`). One caller uses it, the undelivered-response path in `handle_approve` (`crates/sysknife-daemon/src/dispatcher.rs:1426`).

The two paths an operator reaches do neither:

- `cancel_queued` flips `status` to `Canceled` in one UPDATE (`transactions.rs:738`), which is what `handle_cancel` calls (`dispatcher.rs:1556`).
- `cleanup_stale_queued` cancels every queued row past the 15-minute TTL in one UPDATE (`transactions.rs:715`), on a 300-second ticker started at `crates/sysknife-daemon/src/main.rs:133`.

Neither touches `transaction_approvals`, and neither appends an event. Postgres matches: `crates/sysknife-daemon/src/store/postgres.rs:725` and `:707`.

Approve a transaction and then cancel it, and the event chain holds `approval_granted` with no terminal event, while the row stays in `transaction_approvals` with `consumed_at IS NULL` for good. `docs/the-audit-chain.md:222` sets the expectation those events exist to meet:

> The transaction chain records what the daemon *authorized*. It says nothing about whether a human then approved it, whether that approval was spent, or whether it was retracted.

A reader of the verified chain cannot separate "granted, still valid, the operator is deciding" from "granted, then cancelled". The two look identical.

The sweep hits approved transactions too. Its comment at `main.rs:128` says it "only touches expired Queued rows (never Running/Approved)", and there is no `Approved` state in `JobState` (`crates/sysknife-types/src/lib.rs:503`): approval writes a row in a separate table and leaves `status` at `Queued`. So a preview that a human approved and nobody executed is swept on the next tick, and the sweep logs a count without naming what it cancelled.

## What this is not

The receipt stays unspendable after a cancel. `claim_approved_for_execution` requires `status = Queued` (`transactions.rs:667-678`) and nothing moves a row back to `Queued`. This costs completeness of the trail, not privilege, and the fix should not be sold as closing an escalation.

## Suggested shape

1. Factor the delete-plus-`ApprovalRevoked` body out of `revoke_unconsumed_approval` into a helper that runs inside a caller-supplied transaction.
[...]

PR #328, opened 2026-09-01
test: verify live systemd directory modes

### LINKED ISSUE #269: Nothing asserts the runtime directory modes a live daemon actually has, only what the units declare
#263 makes every declaration of the daemon's directory modes agree: `packaging/sysknife-daemon.service`, the unit written by `tests/e2e/ubuntu-provision.sh`, and `packaging/sysknife-tmpfiles.conf` are held equal by `tests/release/systemd-directory-modes.test.sh`. That is the valuable half. What no test checks is the mode a running system actually has.

## Why it matters

The original bug was runtime behaviour: systemd recreated `/run/sysknife` and `/var/lib/sysknife` at the default 0755, discarding what tmpfiles declared. A textual guard cannot see:

- a drop-in at `/etc/systemd/system/sysknife-daemon.service.d/*.conf` setting `RuntimeDirectoryMode=0755`
- a distro patch to the unit
- a mode directive that lands outside `[Service]`, which systemd ignores with `Unknown key name ... in section 'Unit', ignoring`

In each case every declaration file still agrees, the guard passes, and `/run/sysknife` is world-traversable in the VM the release rehearsal signs off on.

## Scope

`tests/e2e/ubuntu-provision.sh` already runs immediately after `systemctl enable --now sysknife-daemon` and checks only that the socket exists. Add two assertions there, comparing against the values parsed out of `packaging/sysknife-tmpfiles.conf` rather than literals:

```sh
stat -c %a /run/sysknife        # must equal the tmpfiles mode for /run/sysknife
stat -c %a /var/lib/sysknife    # must equal the tmpfiles mode for /var/lib/sysknife
```

While in there, `crates/sysknife-daemon/src/transport/listen.rs:96` creates the socket's parent with `create_dir_all`, which yields `0777 & ~umask`, so the manual-start path documented in `HACKING.md` recreates `/run/sysknife` at 0755 after systemd has removed it on stop. Setting the mode explicitly there closes the last site that disagrees.

## Tests first

The VM assertion is the test. Prove it bites by pre-creating the directory at 0755 in the provision script before the daemon starts, confirming the assertion fails, then removing the mutation.

## Difficulty

`medium`, only because it runs inside the Ubuntu VM harness rather than on the host.

ITSMERNB:
Closes #269.

This follows #263 through to the running system:

- derive the expected runtime/state modes from `packaging/sysknife-tmpfiles.conf`
- assert the live `/run/sysknife` and `/var/lib/sysknife` modes immediately after systemd starts the daemon in the Ubuntu provision path
- make manual Unix-socket parent creation set `0750` when SysKnife creates a missing parent, while leaving existing operator-owned directories unchanged
- add unit coverage for both the newly-created and pre-existing parent cases

Local checks:
- `bash -n tests/e2e/ubuntu-provision.sh`
- `bash tests/release/systemd-directory-modes.test.sh`
- `git diff --check`
- mutation proof: forcing a `0755` directory against the tmpfiles-derived `0750` expectation is rejected (`expected=750 actual=755`)

The Windows host cannot execute the Unix daemon tests directly; the Linux Rust CI on this PR is the full compile/test check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,829 Rust tests and 72 frontend tests** form the current deterministic
+**1,831 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/crates/sysknife-daemon/src/transport/listen.rs
+++ b/crates/sysknife-daemon/src/transport/listen.rs
@@ -93,6 +93,7 @@ pub fn bind_unix_listener(target: &ListenTarget) -> Result<UnixListener, ListenT
     match target {
         ListenTarget::Unix(path) => {
             if let Some(parent) = path.parent() {
+                let parent_existed = parent.exists();
                 fs::create_dir_all(parent).map_err(|err| {
                     // Name the socket as well as the directory: the socket is
                     // what the operator configured, the directory is only how
@@ -103,6 +104,22 @@ pub fn bind_unix_listener(target: &ListenTarget) -> Result<UnixListener, ListenT
                         path.display()
                     ))
                 })?;
+
+                // Match the packaged RuntimeDirectoryMode when a manual start
+                // has to recreate /run/sysknife. Do not change an existing
+                // operator-owned directory just because a socket lives in it.
+                if !parent_existed {
+                    use std::os::unix::fs::PermissionsExt;
+                    fs::set_permissions(parent, fs::Permissions::from_mode(0o750)).map_err(
+                        |err| {
+                            ListenTargetError::Io(format!(
+                                "cannot set permissions on directory {} for socket {}: {err}",
+                                parent.display(),
+                                path.display()
+                            ))
+                        },
+                    )?;
+                }
             }
 
             if path.exists() {
@@ -208,6 +225,47 @@ mod tests {
         let _second = bind_unix_listener(&target).expect("a stale socket must be reclaimed");
 
         std::fs::remove_dir_all(&dir).ok();
+    }
+
+    #[test]
+    fn newly_created_socket_parent_uses_runtime_directory_mode() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let root =
+            std::env::temp_dir().join(format!("sysknife-bind-parent-mode-{}", std::process::id()));
+        std::fs::create_dir_all(&root).unwrap();
+        let parent = root.join("run/sysknife");
+        let path = parent.join("daemon.sock");
+
+        let _listener = bind_unix_listener(&ListenTarget::Unix(path))
+            .expect("binding should create the socket parent");
+
+        let mode = std::fs::metadata(&parent).unwrap().permissions().mode() & 0o777;
+        assert_eq!(mode, 0o750);
+
+        std::fs::remove_dir_all(&root).ok();
+    }
+
+    #[test]
+    fn existing_socket_parent_keeps_its_mode() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let root = std::env::temp_dir().join(format!(
+            "sysknife-bind-existing-parent-mode-{}",
+            std::process::id()
+        ));
+        let parent = root.join("socket-dir");
+        std::fs::create_dir_all(&parent).unwrap();
+        std::fs::set_permissions(&parent, std::fs::Permissions::from_mode(0o700)).unwrap();
+
+        let path = parent.join("daemon.sock");
+        let _listener =
+            bind_unix_listener(&ListenTarget::Unix(path)).expect("binding should succeed");
+
+        let mode = std::fs::metadata(&parent).unwrap().permissions().mode() & 0o777;
+        assert_eq!(mode, 0o700);
+
+        std::fs::remove_dir_all(&root).ok();
     }
 
     #[test]
--- a/docs/distro-support.md
+++ b/docs/distro-support.md
@@ -82,7 +82,7 @@ family and the atomic story family are implemented and covered by the workspace
 suite. What is missing is a way to put the helpers somewhere the daemon's own
 grants already point.
 
-The deterministic workspace baseline is 1,829 Rust tests plus 72 frontend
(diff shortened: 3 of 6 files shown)


vladimirrott (changes requested):
That is fast, and the shell half is exactly what #269 asked for. Reviewed at `2fc1276`.

`tmpfiles_mode_for` derives the expectation from `packaging/sysknife-tmpfiles.conf` instead of restating it, validates what it parsed before comparing, and fails loudly on an empty read rather than comparing against nothing. The assertion sits immediately after `systemctl is-active`, which is the only place where "the mode a running daemon has" is a different question from "the mode a file declares". Both directories, not just the runtime one.

## One blocking item

`rust` is red because the suite is 1787 and `tests/evidence/workspace-tests.json` says 1785. Two tests added, artifact not moved. That coupling is not visible in your diff and it catches nearly every first PR here, so no reflection on you.

Main has moved twice since you branched, and the baseline is now **1786**, so your rebase lands on **1788**:

```
git fetch origin main && git rebase origin/main
```

Then change `1,786` to `1,788` in `README.md`, `docs/introduction.md` and `docs/distro-support.md`, in the same commit. `docs-and-hygiene` fails if the artifact moves and the prose does not.

## Two on the Rust side, neither blocking

**The mode is applied after the directory exists.** `create_dir_all` makes the directory at the umask-derived mode, usually 0755, and `set_permissions` tightens it a moment later. Anything that can act in the parent during that window sees a world-readable directory. Under `/run` that is only root, but `bind_unix_listener` also serves the dev fallbacks, and `/tmp` is a different story.

`DirBuilder` applies the mode in the `mkdir` call itself:

```rust
use std::os::unix::fs::DirBuilderExt;
fs::DirBuilder::new()
    .recursive(true)
    .mode(0o750)
    .create(parent)
```

It also drops the `parent_existed` probe, because recursive `create` leaves an existing directory alone, which is the behaviour your second test pins. Two notes if you take it: the mode applies to every directory the recursive call creates, not only the leaf; and `mkdir` masks with umask, so `newly_created_socket_parent_uses_runtime_directory_mode` should either set a known umask or assert `0o750 & !umask` rather than a bare `0o750`.

**`0o750` is now a third copy.** `packaging/sysknife-tmpfiles.conf` declares it, `packaging/sysknife-daemon.service` declares it as `RuntimeDirectoryMode`, and `listen.rs` now hardcodes it. `tests/release/systemd-directory-modes.test.sh` ties the first two together and the heredoc in `ubuntu-provision.sh` to both. Nothing ties the Rust constant to any of them, and your unit test asserts the same literal, so the test and the code are a parallel pair: change tmpfiles to `0700` and both stay green while the daemon keeps making 0750 directories.

The shell half already shows the fix. A named constant plus a `#[test]` that parses `packaging/sysknife-tmpfiles.conf` and asserts the constant equals the declared mode would make the Rust side derived too.

## One I checked and it is fine

`parent.exists()` swallows `EACCES` and reports "not found", which has bitten this repo before. I worked out whether it can make `set_permissions` retighten an operator's existing directory: it would need `exists()` to return false while `create_dir_all` succeeds, and any `EACCES` that hides the directory from `stat` also blocks `mkdir`, so the error comes first. Not reachable. Recording it because it looked like a bug for a minute.

Fix the baseline and I will merge. The two Rust notes can land here or as a follow-up, your call.

vladimirrott:
Correcting one number in my review, before you spend a run on it.

`main` moved twice after I posted: #330 landed six review fixes and #329 added `--dangerously-skip-approval`. The baseline is now **1820**, so your two tests put your rebase on **1822**, not the 1788 I wrote.

```sh
git fetch origin main && git rebase origin/main
```

Then `1,820` becomes `1,822` in `README.md`, `docs/introduction.md` and `docs/distro-support.md`, in the same commit.

One thing worth knowing before you rebase: #330 wired `tests/release/database-path-agreement.test.sh` into `docs-and-hygiene`, so there is one more hygiene step than when you branched. It is unrelated to your change and should stay green.

Everything else in the review stands. The two Rust notes are still optional, and I will merge on the baseline alone if you would rather leave them.

ITSMERNB:
Thanks for the correction. Main moved again while I was picking this back up, so I rebased onto the current main baseline of 1,829; with the two added tests the branch is now at 1,831, and the three published references are synced in the same baseline commit (a719d44). The Rust job and docs-and-hygiene are green on the new head. I kept the two optional Rust notes out of this PR so the fix stays scoped to #269.

vladimirrott (approved):
Verified at `a719d44`. Merging.

The baseline is right and it was measured rather than asserted: the artifact names `cf81ecc` with a timestamp in your own timezone, and all three fields moved together for the reason they are supposed to. 1831 on Linux, matching.

Three mutations on the Rust half, three caught:

| Mutation | Caught by |
|---|---|
| the chmod never runs | `newly_created_socket_parent_uses_runtime_directory_mode` |
| the mode becomes 0755 | same |
| an existing directory gets chmod-ed too | `existing_socket_parent_keeps_its_mode` |

That third one is the reason this holds up. It would have been easy to write only the positive test, and then narrowing an operator's own 0700 directory to 0750 would have passed.

The shell half derives the expectation from `packaging/sysknife-tmpfiles.conf`, validates what it parsed before comparing, and asserts immediately after `systemctl is-active`, which is the only point where "the mode a running daemon has" is a different question from "the mode a file declares". Both directories, not just the runtime one.

## The two optional notes, left as they were

You left the `DirBuilder` change and the third-copy-of-0750 note, which is fine; I marked both optional and the PR stands without them.

One update on the first, since the picture changed after I wrote it. A red-team pass yesterday found the same shape in `packaging/sysknife-mount-edit`, where `op_addswap` and `op_rmswap` guard only the final path component while `op_mount` resolves the whole path. That is a private advisory rather than an issue, and it is a more serious instance than yours: this one is a narrow window on a directory the daemon just created, and the file's own `parent_existed` check means a pre-existing symlink never reaches the chmod at all. So nothing here is urgent. I mention it only because the class is now a thing the project is actively fixing elsewhere, and `DirBuilder::mode` would take this one off the list for free whenever you are next in the file.

Closes #269. Thanks for turning it around inside a day, twice.

## Next

#331 and #251 are both still open with your name on them. #331 is the closer fit and it is the smaller of the two.

ITSMERNB:
Thanks, Vladimir. Glad the negative case earned its keep — preserving an operator-owned directory was the part I most wanted pinned down. Appreciate the careful review and the pointers on the follow-up class as well. I will keep #331 in mind as the next fit.

Issue #338, opened 2026-09-01
audit verify says OK over a wiped log, so an empty store and a deleted one read the same


vladimirrott:
`sysknife audit verify` reports success over a database whose rows have all been deleted.

```
$ sqlite3 daemon.sqlite 'DELETE FROM transactions; DELETE FROM audit_events;'
$ sysknife audit verify
OK: 0 row(s) verified in .../daemon.sqlite
NOTE: no independent checkpoint anchor is configured, so removal of the newest rows
      would not be detectable — a truncated chain still verifies. Set SYSKNIFE_CHECKPOINT_DB
      and run `sysknife audit checkpoint` periodically; see docs/the-audit-chain.md.
OK: 0 approval event(s) verified
OK: 0 row(s) still match the approval event they committed to
$ echo $?
0
```

Three previously signed actions were deleted. The verdict is `OK` and the exit code is 0.

## Why this is not simply a bug

A forward hash chain cannot detect the removal of its own tail without an external anchor. That is what checkpoint anchoring exists for, the tool says so in the NOTE, and `docs/the-audit-chain.md` explains it. Nothing here is hidden.

The problem is what an operator reads. Two very different states produce the same headline:

- a fresh install where nothing has run yet
- a host where the entire trail was deleted

Both print `OK` and exit 0. The one line that distinguishes them is a NOTE the reader has to interpret, below a verdict that already told them everything is fine.

## Scope

Make the verdict carry the distinction the tool already knows:

- **Zero rows and no anchor configured** is not `OK`. It is "nothing to verify, and this command cannot tell an empty log from a deleted one". A distinct exit code, or at minimum a different word than `OK`, gives a monitoring script something to alert on.
- **Zero rows with an anchor configured** genuinely is verifiable: the anchor either names a tip that should exist or it does not. That case can stay `OK` or become an explicit failure, and the anchor is what makes the difference.
- The `--json` shape should carry the same distinction, since that is what automation reads.

Worth considering alongside it: whether the daemon should anchor on a schedule by default rather than only when `SYSKNIFE_CHECKPOINT_DB` is set, so the mitigating control is not opt-in on a product whose subject is tamper evidence. That is a bigger decision and can be split out.

## Tests first

A test that writes rows, deletes them, and asserts the verdict is not the same one a fresh store produces. Then a second that configures an anchor and asserts the truncation is caught, so the fix is proved to add detection rather than only to change wording.

## Difficulty

[...]

PR #340, opened 2026-09-02
fix(audit): include anchor in verification verdict

### LINKED ISSUE #221: audit verify: an inconclusive anchor masks a broken chain, and --json status ignores the anchor

k4its1t:
## Summary

- combine the local audit checks and configured anchor with the documented tamper-over-inconclusive precedence
- derive the JSON headline status from that same combined verdict
- run the real CLI binary with `--json` against an intact SQLite chain and a higher signed checkpoint in an isolated live-Postgres schema, asserting process exit code 1 and JSON status `broken`
- run that regression in the existing Postgres contract job and retain the measured 1,833-test Linux baseline

Closes #221

## Validation

GitHub Actions on final commit `22dfe55`:

- Rust format, Clippy, rustdoc, and workspace suite: 1,833/1,833 passed, 6 skipped
- live Postgres contract: `audit_verify_exits_with_code_1_when_anchor_is_truncated` passed with the exit-code and JSON assertions
- docs and hygiene, scripts lint, frontend, dependency review, security audit, secret scan, and CodeQL

The baseline records the Linux run at commit `5776193d286368ceb36f03afad181c49742f36d9`, rather than inferring a count from the macOS-visible suite. `python3 scripts/check_evidence_claims.py .` passed in CI, and `git diff --check` passes locally.

AI assistance was used to inspect the issue and review, implement the focused change, run validation, and prepare this PR. I reviewed the diff and test results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,6 +286,11 @@ jobs:
       - name: Run live Postgres migration and store contract
         run: cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
 
+      - name: Run live CLI anchor exit-code contract
+        run: >-
+          cargo test -p sysknife-cli --test cli_smoke --locked
+          audit_verify_exits_with_code_1_when_anchor_is_truncated -- --ignored --exact
+
   security-audit:
     runs-on: ubuntu-latest
     steps:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4847,6 +4847,8 @@ dependencies = [
  "rustyline",
  "serde",
  "serde_json",
+ "sqlx-core",
+ "sqlx-postgres",
  "strip-ansi-escapes",
  "sysknife-brain",
  "sysknife-core",
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,831 Rust tests and 72 frontend tests** form the current deterministic
+**1,833 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,829 Rust tests and 72 frontend tests** form the current deterministic
+**1,831 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/apps/sysknife-cli/Cargo.toml
+++ b/apps/sysknife-cli/Cargo.toml
@@ -49,3 +49,8 @@ assert_cmd = "2"
 predicates = "3"
 tempfile = "3"
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros", "net", "io-util"] }
+sqlx-core = { version = "0.9", default-features = false, features = [
+    "_rt-tokio",
+    "_tls-rustls-ring-webpki",
+] }
+sqlx-postgres = { version = "0.9", default-features = false }
--- a/apps/sysknife-cli/src/runner.rs
+++ b/apps/sysknife-cli/src/runner.rs
@@ -1015,14 +1015,7 @@ pub async fn run_audit_verify(args: AuditVerifyArgs, log: &Logger) -> Result<(),
     // this command used to report `audit_anchor: {configured: true}` without
     // ever reading the anchor — implying a check it did not perform.
     let anchor = verify_configured_anchor(&lacs_config, &db_path, &verifier).await;
-    let anchor_exit = anchor
-        .as_ref()
-        .map(sysknife_daemon::audit_chain::checkpoint_outcome_to_exit_code)
-        .unwrap_or(0);
-
-    // Worst verdict wins: a chain that verifies against itself but not against
-    // its anchor has been tampered with, whatever the local check said.
-    let exit_code = outcome.exit_code().max(anchor_exit);
+    let exit_code = combined_verification_exit_code(&outcome, anchor.as_ref());
     emit_verification(&args, log, &outcome, &label_for_diag, anchor.as_ref());
     if exit_code == 0 {
         Ok(())
@@ -1484,7 +1477,7 @@ fn emit_verification(
 
     if args.json {
         let payload = json!({
-            "status": status_word(verification.exit_code()),
+            "status": status_word(combined_verification_exit_code(verification, anchor)),
             "backend": backend_label,
             "chain": outcome_json(&verification.chain),
             "approval_events": outcome_json(&verification.events),
@@ -1602,6 +1595,29 @@ fn emit_verification(
                  were deleted from the end of the chain"
             ));
         }
+    }
+}
+
+/// Combine the local audit checks with the external anchor using the same
+/// precedence as [`AuditVerification::exit_code`]. A detected break is stronger
+/// evidence than a different check being inconclusive, so exit code `1` must
+/// outrank `2` rather than relying on numeric ordering.
+fn combined_verification_exit_code(
+    verification: &AuditVerification,
+    anchor: Option<&CheckpointOutcome>,
+) -> i32 {
+    let codes = [
+        verification.exit_code(),
+        anchor
+            .map(sysknife_daemon::audit_chain::checkpoint_outcome_to_exit_code)
+            .unwrap_or(0),
+    ];
+    if codes.contains(&1) {
+        1
+    } else if codes.contains(&2) {
+        2
+    } else {
+        0
     }
 }
 
@@ -2210,6 +2226,15 @@ mod tests {
         attribution: Option<sysknife_daemon::audit_chain::AttributionCensus>,
         json: bool,
     ) -> String {
+        rendered_with_anchor(chain, attribution, json, None)
+    }
+
+    fn rendered_with_anchor(
+        chain: sysknife_daemon::audit_chain::VerifyOutcome,
+        attribution: Option<sysknife_daemon::audit_chain::AttributionCensus>,
+        json: bool,
+        anchor: Option<&CheckpointOutcome>,
+    ) -> String {
         use sysknife_daemon::audit_chain::{BindingOutcome, VerifyOutcome};
         let dir = tempfile::tempdir().expect("tempdir");
         let path = dir.path().join("verify.log");
@@ -2227,7 +2252,7 @@ mod tests {
             &log,
             &verification,
             "/tmp/test.db",
-            None,
+            anchor,
         );
         std::fs::read_to_string(&path).expect("logger wrote the rendered output")
     }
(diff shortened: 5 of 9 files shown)


vladimirrott (changes requested):
Verified at `67bedef`. The diagnosis is exactly right, and the fix is the smallest thing that could be correct.

The old line was `outcome.exit_code().max(anchor_exit)`, and the bug is that these codes are not ordinal. `2` means "could not tell" and `1` means "found a break", so `max` made the weaker verdict win. An inconclusive anchor over a broken chain reported inconclusive. That is a genuinely easy thing to read past, and you did not.

Two things I checked and liked:

- `combined_verification_exit_code` reproduces the shape of `AuditVerification::exit_code` (`audit_chain.rs:1433`) rather than inventing a rule, so the CLI and the library now agree on what beats what. I compared them line by line.
- The scope stops in the right place. The human-readable path already rendered the anchor through `anchor_line`, so only the exit code and the JSON status were wrong, and the diff touches only those.

Both new tests bite. I reverted each half separately:

| Mutation | Caught by |
|---|---|
| `max(codes[0], codes[1])` restored | `a_detected_break_outranks_an_inconclusive_anchor_in_the_cli_exit_code` |
| JSON status back to `verification.exit_code()` | `json_status_includes_a_truncated_anchor_verdict` |

And 1831 passes on Linux, matching the artifact.

## One thing to add, and then I will merge

There is a third mutation, and nothing catches it:

```diff
-    let exit_code = combined_verification_exit_code(&outcome, anchor.as_ref());
+    let exit_code = outcome.exit_code();
```

That is `run_audit_verify` dropping the anchor from the **process exit code**, which is the first half of what #221 reports. The full workspace suite passes with it in place. All 1831.

Both of your tests reach the new function directly, one as a unit call and one through the JSON renderer. Neither goes through `run_audit_verify`, so the command's own exit status is the one behaviour in this issue that is still unpinned. Someone rewiring that line would get a green board.

`cli_smoke.rs` already drives the real binary and asserts exit codes, and `audit_export_distinguishes_unreadable_from_absent` in that file is the closest pattern: build a store, put it in the state you want, run the command, assert the status. A chain that verifies locally against an anchor that says truncated should exit 1.

That is the only blocking item. If you would rather not build the fixture, say so and I will merge this and add the test myself, because the fix is right either way.

## Not a problem, worth knowing

Your PR body says the evidence file was not rewritten from the macOS run, and that 1,831 comes from 1,829 plus two. That is correct, and I confirmed it by running the suite on Linux.

It is also a clean example of #278. `workspace-tests.json` carries `tests`, `commit` and `measured_at`, and all three moved in your diff, which is the shape that normally means a real measurement happened. Nothing in the repo can tell the difference between that and three fields edited together. You said so plainly in the PR body, which is the only reason I knew to check rather than assume. I am citing this on #278 as the concrete case.

## Optional, ignore if you like

The precedence rule now exists in two places. If `AuditVerification::exit_code` ever changes, `combined_verification_exit_code` will not follow, and the two disagreeing is the failure this repo keeps finding. A shared `fn worst_of(codes: &[i32]) -> i32` in `audit_chain.rs` that both call would remove the twin. Not worth blocking on.

## Next

#338 is the larger sibling of this one and it is still unclaimed: `audit verify` reports `OK` and exits 0 over a database whose rows were all deleted, so a fresh install and a wiped log read the same. Same command, same argument about what a verdict should carry. Yours if you want it.

vladimirrott:
One number moved under you, and it is not your doing.

#328 merged a few minutes ago and added two tests of its own, so `main` is now at **1831** and your rebase lands on **1833**, not the 1831 in your diff. You and @ITSMERNB both branched from 1829 and both added two, which is why you both computed the same target.

```sh
git fetch origin main && git rebase origin/main
```

Then `1,831` becomes `1,833` in `README.md`, `docs/introduction.md` and `docs/distro-support.md`.

Running the generator rather than editing the three fields also settles the provenance point from my review on its own, since `commit` and `measured_at` will then describe a run that actually happened.

Nothing else changes. The review above still stands, and the one blocking item is still the `run_audit_verify` exit-code test.

k4its1t:
Addressed the requested command-level regression and the new `main` conflict.

- Added `audit_verify_exits_with_code_1_when_anchor_is_truncated` in `cli_smoke.rs`. It creates an intact local SQLite audit chain, stores a higher checkpoint signed by the same key in an isolated live-Postgres schema, runs the compiled `sysknife audit verify` command, and asserts process exit code 1.
- Wired only that ignored live test into the existing Postgres contract job; no new job or production dependency was added.
- Merged current `main`; the sole conflict was the workspace evidence artifact.
- The Linux workspace run at `5776193d286368ceb36f03afad181c49742f36d9` measured and passed 1,833/1,833 tests, and the final baseline update is now green.

I left the optional shared precedence-helper refactor out of this PR so the change remains scoped to #221 and the blocking review request.

vladimirrott (changes requested):
Reviewed at `c8ae0a1`, on Linux with live Postgres. The rebase is correct: 1,833 in the artifact, 1,833 in all three prose files, and `cargo nextest run --workspace --locked` gives 1,833 passed. `commit` and `measured_at` now describe a run that happened, which closes the provenance half on its own.

The new CI step runs the test rather than filtering to nothing, which is the failure mode I was most worried about:

```
running 1 test
test audit_verify_exits_with_code_1_when_anchor_is_truncated ... ok
```

## One blocking item, and it is two lines

The new test passes with the fix reverted. I restored `main`'s precedence in `run_audit_verify`:

```rust
let exit_code = outcome.exit_code().max(anchor_exit);
```

and ran your test against it:

```
running 1 test
test audit_verify_exits_with_code_1_when_anchor_is_truncated ... ok
```

The fixture is an intact local chain plus a truncated anchor, so the codes are `0` and `1`. `0.max(1)` is already `1`, and `main` reaches the same answer by the wrong route. The precedence bug only shows when the local chain is `Broken` (1) and the anchor is `CannotVerify` (2), where `1.max(2)` gives 2 and yours gives 1.

The cheaper fix is to assert the other half of #221 instead. Your fixture already produces the state; it needs `--json` and one assertion:

```rust
    .args(["audit", "verify", "--json"])
    .output()
    .expect("run audit verify");
let doc: serde_json::Value =
    serde_json::from_slice(&output.stdout).expect("--json emits one document");
assert_eq!(
    doc["status"], "broken",
    "the --json verdict must reflect the anchor: {}",
    String::from_utf8_lossy(&output.stdout)
);
```

I ran that against your branch and it passes. Against `main`'s `status_word(verification.exit_code())` it fails:

```
  left: String("intact")
 right: "broken"
```

That is the "`--json` status ignores the anchor" half of the issue title, at the command level, on the fixture you already built. Keep the exit-code assertion as well.

## Not blocking

`combined_verification_exit_code` reads well and the doc comment says why `1` outranks `2` rather than leaving it to numeric order. Two unit tests cover both halves. Leaving the shared-helper refactor out was the right call.

Once the `--json` assertion is in, this is ready.

k4its1t:
Addressed the final blocking review item in `22dfe55`.

- The existing live-Postgres CLI fixture now runs `sysknife audit verify --json`.
- It keeps the process exit-code assertion at `1`, parses stdout as one JSON document, and asserts the top-level status is `broken`.
- The test name and existing Postgres contract job are unchanged; no dependency, helper, workflow, baseline, or production-code change was added.

The new Linux run executed exactly that ignored test (`1 passed`) against live Postgres. The full workspace remained at 1,833/1,833 passed with 6 skipped, and the remaining CI and CodeQL checks are green. I left the optional shared precedence helper and #338 out of scope.

vladimirrott (approved):
`22dfe55` does it. This is the assertion I asked for, and I checked it the way the earlier version failed rather than by reading it.

Reverted `"status": status_word(combined_verification_exit_code(verification, anchor))` back to `main`'s `status_word(verification.exit_code())`, rebuilt, ran the test:

```
thread 'runner::tests::json_status_includes_a_truncated_anchor_verdict' panicked
  left: String("intact")
 right: "broken"
```

That is the signature that was missing three rounds ago: a chain that verifies against itself, an anchor that says truncated, and a `--json` verdict that used to report `intact`. It now fails when the fix is removed.

I also read the two things I will not approve unread:

- **`.github/workflows/ci.yml`**: one step added to the existing `postgres-contract` job. No trigger, permission, secret or action change.
- **`apps/sysknife-cli/Cargo.toml`**: `sqlx-core` and `sqlx-postgres` as dev-dependencies. They mirror `crates/sysknife-daemon/Cargo.toml` exactly, same `0.9`, same `default-features = false`, same deliberate avoidance of the umbrella crate so `sqlx-sqlite` cannot collide with `rusqlite` at the `links = "sqlite3"` level. `Cargo.lock`'s package set is byte-identical to `main`'s, so no new third-party code enters the build. That is the answer I wanted, and you supplied it without being asked.

Merging.

## A gap I am not asking you to close

`combined_verification_exit_code` is tested directly, and its use inside `emit_verification` is now pinned by your JSON test. Its use in `run_audit_verify` is not. I reverted that one call site back to `.max()` and left everything else in place:

```
test result: ok. 264 passed; 0 failed; 0 ignored
```

The ignored Postgres test does not catch it either: a truncated anchor over an intact chain gives 1 under both forms, and the two only diverge on a broken chain with an unreachable anchor, which your unit test covers at the helper rather than at the call site.

Not blocking, and not your debt. You fixed the reported bug and pinned the path it was reported on. I will open the follow-up; if you want it, say so and it is yours.

Thank you for staying with this through three rounds and for the merge conflicts you absorbed along the way. The audit chain is the part of this project where being pedantic is the whole job, and you have been.

If you run Linux where an audit trail actually matters to someone, I would like to hear how `sysknife audit verify` reads in that setting. `npx sysknife-setup` to start, and verification touches nothing on the host.

PR #341, opened 2026-09-02
fix(e2e): reject incomplete story runs

### LINKED ISSUE #247: run-stories.sh exits 0 for a fully throttled or skipped suite

bferanmi806-sketch:
## Summary

- Fixed the false-green story verdict contract across `run-stories.sh`, `exec/run-exec-stories.sh`, and `dev-stories.sh`.
- A suite now fails when any story is `SKIP`, `RATELIMIT`, or exits 0 without an explicit `PASS` marker. `RATELIMIT` remains a distinct diagnostic.
- The evidence guard now rejects story claims and Validated record/replay pairs whose artifacts report skipped or rate-limited stories.
- Added a host-side mutation regression test that executes staged copies of all three production runners, mutates the aggregate gate, and exercises the writer-produced evidence path.
- Wired the regression test into GitHub CI and `scripts/ci-local.sh`.

## Verification

- Passed: `bash tests/e2e/story-runner-verdicts.test.sh`.
- Passed: `bash tests/release/public-claims.test.sh`.
- Passed: evidence checker, Python compilation, Bash syntax, `cargo fmt --all --check`, ShellCheck 0.11.0 on touched shell files, yamllint on the changed workflow, and `git diff --check`.
- Linux non-GUI nextest: 1,815 passed, 1 failed, 5 skipped. `action_reference_doc_is_current` is a pre-existing failure reproduced on clean `main`.
- Full GUI workspace tests could not run in WSL because GTK/GLib system packages are unavailable.

Closes #247

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,9 @@ jobs:
       - name: Check story metadata is derived from the story files
         run: bash tests/e2e/story-metadata.test.sh
 
+      - name: Check story runners reject unproven results
+        run: bash tests/e2e/story-runner-verdicts.test.sh
+
       - name: Check Ubuntu VM bootstrap contract
         run: bash tests/e2e/ubuntu-vm-bootstrap.test.sh
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,24 @@ Releases before `0.2.5` predate the public launch; their notes live in the
 
 ## [Unreleased]
 
+### Fixed
+
+- **A story suite that proved nothing no longer reports success.** All three
+  runners (`run-stories.sh`, `exec/run-exec-stories.sh`, `dev-stories.sh`)
+  decided their exit status from `fail_count` alone, so a run where every story
+  was skipped or rate-limited printed `0/N passed, 0 failed` and exited 0.
+  `run-stories.sh` also rewrote a failing story's verdict to `RATELIMIT`, moving
+  it out of the only counter that could fail the run. A pass is now a positive
+  `PASS` marker rather than an inferred exit status, and skipped, rate-limited
+  and zero-pass runs all fail. `scripts/check_evidence_claims.py` refuses an
+  artifact carrying a non-zero `skipped` or `rate_limited` as evidence for a
+  published pass rate or a Validated tier.
+  `tests/e2e/story-runner-verdicts.test.sh` pins the contract by executing
+  staged copies of the three production runners and mutating their real exit
+  gates, so a reverted fix turns the suite red. Story 28 also left the
+  no-argument default sets, which selected a Fedora-only story on every Ubuntu
+  host. ([#247](https://github.com/lacs-project/sysknife/issues/247))
+
 ### Security
 
 - **A transaction can only be approved, cancelled, inspected or executed by the
--- a/scripts/check_evidence_claims.py
+++ b/scripts/check_evidence_claims.py
@@ -178,6 +178,19 @@ def load_story_runs(root: Path) -> list[dict]:
     return runs
 
 
+def story_run_has_no_unproven_results(run: dict) -> bool:
+    """Only a run with explicit zero skipped and rate-limited stories can prove claims."""
+    totals = run.get("totals")
+    if not isinstance(totals, dict):
+        return False
+    return all(
+        isinstance(totals.get(field), int)
+        and not isinstance(totals.get(field), bool)
+        and totals[field] == 0
+        for field in ("skipped", "rate_limited")
+    )
+
+
 def check_story_claims(texts: dict[str, str], runs: list[dict]) -> list[str]:
     problems = []
     pattern = re.compile(r"([0-9]+)\s*/\s*([0-9]+)\s+stories")
@@ -193,6 +206,7 @@ def check_story_claims(texts: dict[str, str], runs: list[dict]) -> list[str]:
                 if run.get("story_set") in FULL_STORY_SETS
                 and run.get("totals", {}).get("passed") == passed
                 and run.get("totals", {}).get("total") == total
+                and story_run_has_no_unproven_results(run)
                 # A replay that missed, or served nothing, proves nothing — the
                 # harness says so and fails the run. Such an artifact used to be
                 # written anyway, with a healthy-looking pass rate.
@@ -355,9 +369,12 @@ def replay_verified_releases(root: Path) -> set[str]:
         if not release:
             continue
         if path.name.endswith(".replay.json"):
-            if str(run.get("cassette_audit", {}).get("verdict", "")) == "ok":
+            if (
+                str(run.get("cassette_audit", {}).get("verdict", "")) == "ok"
+                and story_run_has_no_unproven_results(run)
+            ):
                 replays.add(release)
-        else:
+        elif story_run_has_no_unproven_results(run):
             records[release] = path
     return set(records) & replays
 
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -283,6 +283,7 @@ run_hygiene_group() {
     run_step 'hygiene: provider-parity.test.sh' bash "$repo_root/tests/e2e/provider-parity.test.sh"
     run_step 'hygiene: vm-env-secrets.test.sh' bash "$repo_root/tests/e2e/vm-env-secrets.test.sh"
     run_step 'hygiene: story-metadata.test.sh' bash "$repo_root/tests/e2e/story-metadata.test.sh"
+    run_step 'hygiene: story-runner-verdicts.test.sh' bash "$repo_root/tests/e2e/story-runner-verdicts.test.sh"
     run_step 'hygiene: vm-sync-parity.test.sh' bash "$repo_root/tests/e2e/vm-sync-parity.test.sh"
     run_step 'hygiene: docs-share-cards.test.sh' bash "$repo_root/tests/release/docs-share-cards.test.sh"
     run_step 'hygiene: install-paths.test.sh' bash "$repo_root/tests/release/install-paths.test.sh"
--- a/tests/e2e/dev-stories.sh
+++ b/tests/e2e/dev-stories.sh
@@ -218,7 +218,7 @@ if [[ $# -gt 0 ]]; then
 elif [[ "$ALLOW_DESTRUCTIVE" == "1" ]]; then
     STORIES=(1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 32 33 34 35 36 37 38 39 40 41 42 43 44 45 46 47 48 49 50 51 52 53 54)
 else
-    STORIES=(1 2 3 4 5 6 7 11 12 13 14 15 16 17 21 22 25 26 28 29 32 38 41 46 47 48 49)
+    STORIES=(1 2 3 4 5 6 7 11 12 13 14 15 16 17 21 22 25 26 29 32 38 41 46 47 48 49)
 fi
 
 # ---------------------------------------------------------------------------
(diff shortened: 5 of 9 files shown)


vladimirrott (changes requested):
Reviewed at `51d4c04`. This is a strong first PR, and the test design is the reason.

`story-runner-verdicts.test.sh` stages copies of the three real runners and mutates the production text, so it proves the shipped decision rather than a parallel reimplementation of it. That is the failure mode most harness self-tests in this repo have had. Six mutations, all of which the suite catches:

| Mutation | Result |
|---|---|
| `run-stories.sh` aggregate gate back to `fail_count \|\| cassette_failed` | red, 2 assertions |
| `run-stories.sh` infers PASS from exit 0 | red, 2 assertions |
| `run-exec-stories.sh` gate back to `fail_count` | red |
| `dev-stories.sh` gate back to `fail_count` | red |
| `story_run_has_no_unproven_results` returns `True` | red, `public-claims.test.sh` |
| `replay_verified_releases` tier half only | red, `public-claims.test.sh` |

Your own in-test mutation also fails loudly when its anchor moves, rather than inspecting nothing:

```
story-runner gate mutation did not find one production guard
```

Shellcheck 0.11.0 clean across `tests/e2e tests/release scripts assets/demo`, yamllint clean, evidence checker passes.

## One blocking item

`tests/e2e/dev-stories.sh` with no arguments now exits 1 on Ubuntu, and the reason is legitimate rather than a bug in your change.

Story 28 is in the default non-destructive set (`dev-stories.sh:221`) and skips on any non-Fedora host:

```
SKIP: story-28.sh requires a Fedora-family host (got: debian)
```

Staged both revisions against the real `story-28.sh` on Ubuntu:

```
### main (5a9450a)  exit=0
Summary: 1/2 passed, 0 failed, 1 skipped

### PR 341 (51d4c04)  exit=1
Summary: 1/2 passed, 0 failed, 1 skipped
```

`docs/developer-guide.md:113` and `docs/contributing/CONTRIBUTING.md:136` both document that bare command, so every Ubuntu contributor gets a red run plus a NOTE about Fedora Atomic stories 8 and 10, which the default set does not even include.

Your gate is right. The default set is wrong: the NOTE claims those stories "should always pass on any Linux host", and story 28 cannot. Dropping `28` from line 221 fixes it in one character and makes the NOTE true.

I checked the other two runners for the same shape and they are clean. No Ubuntu-family story (55 to 104) can skip, so the release evidence path is untouched, and `run-exec-stories.sh`'s non-destructive set `(1 2 3 6)` contains none of the eight skippable exec stories.

## Not blocking

- `assert_output_not_contains ... 'FAIL'` greps the whole run output for the substring, so a future story name or summary wording containing it would trip the assertion. Anchoring on `RESULTS[` or the summary line would hold up better.
- Only `run-stories.sh` gets an in-test mutation. The exec and dev gates are covered by assertions that I confirmed bite, so this is a consistency point rather than a hole.
- The last `rate_limited` mutation in `public-claims.test.sh` does not restore the fixture. Nothing follows it today, so it costs nothing until something does.
- The `sed -i 's/\r$//'` lines are harmless on a clean checkout. See below for why they are probably in your tree.
- #247's original body also asked for a story-set size assertion. The scope comment that superseded it does not, so I am not counting it against this PR.

## Your `action_reference_doc_is_current` failure

That test passes on clean `main` here:

```
PASS [ 0.017s] sysknife-daemon::action_reference_doc action_reference_doc_is_current
```

It does a byte-exact `assert_eq!` between `docs/action-reference.md` on disk and a generated string that uses `\n`. Under WSL with `core.autocrlf=true`, the checked-out file has `\r\n` and the comparison fails on line endings alone. Your `sed -i 's/\r$//'` calls point at the same cause. `git config core.autocrlf false` and a fresh checkout should clear it, and it is worth doing before your next PR so your local runs agree with CI.

Fix the story 28 line and this is ready to land.

vladimirrott (changes requested):
Thanks for staging real copies of the production runners instead of reimplementing their classification. `run-stories-no-unproven-gate.sh` is the part that convinced me: it reverts the aggregate guard in a staged copy and asserts the SKIP fixture goes green, so the assertions above it are pinned to the production decision rather than to a parallel expectation. That is the standard this repo asks for. Verifying each JSON mutation with a `grep -q ... || exit 1` before calling `assert_rejected` closes the other half, where a mutation that failed to apply would leave the assertion passing over unchanged input.

Story 28 is out of the `dev-stories.sh` default set and the NOTE text matches it. That one is settled.

## Blocking: `run-stories.sh` keeps its own copy of the set, and 28 is still in it

I asked about `dev-stories.sh` last round and stopped there. `run-stories.sh:228` carries a second copy of the curated non-destructive set:

```
STORIES=(1 2 3 4 5 6 7 11 12 13 14 15 16 17 \
         21 22 25 26 28 29 32 38 41 46 47 48 49)
```

Story 28 gates on distro family at line 15, before it reaches the CLI at line 28:

```
$ . /etc/os-release && echo "${ID_LIKE:-$ID}" | tr ' ' '\n' | head -1
debian
$ bash tests/e2e/stories/story-28.sh
SKIP: tests/e2e/stories/story-28.sh requires a Fedora-family host (got: debian)
$ echo $?
0
```

With the new aggregate guard treating `skip_count -gt 0` as a failure, one SKIP fails the run, so `bash tests/e2e/run-stories.sh` with no arguments exits 1 on any Debian-family host. Dropping `28` from line 228 is the whole fix. Story 28 is untagged, so the `ubuntu` set derived from the headers never contained it and `run-stories.sh ubuntu` is unaffected; the six Fedora-gated stories are 28, 33, 34, 40, 43 and 53, and 28 is the only one in that default set.

The new gate cannot see this because every `run_case` passes explicit story IDs, so the no-argument default path is never exercised. A case that runs the default set would have caught it.

## Optional

- `run-exec-stories.sh` has the same shape with a smaller blast radius. `exec-11.sh` carries the Fedora gate and sits in the `ALLOW_DESTRUCTIVE=1` set (1..12) but not in the safe default `(1 2 3 6)`, so only an `ALLOW_DESTRUCTIVE` exec run on a Debian host trips it. Fine to leave as is.
- `assert_output_not_contains ... 'FAIL'` reads `CASE_OUTPUT` from the preceding `run_case`, and the `RATELIMIT` assertion above it proves that output non-empty, so it holds. Anchoring on the runner's own `FAIL` column instead of the bare substring would make it harder to break by accident.

I did not execute this branch. It changes `.github/workflows/ci.yml`, so it screened READ-ONLY here and I reviewed it by reading the diff. The story-28 output above comes from `main`, where that file is unchanged by this PR. I left the queued workflow run unapproved for the same reason.

bferanmi806-sketch:
Addressed the second review in `dc09ce7`.

Story 28 is now out of `run-stories.sh`’s no-argument default set as well, and I added a regression case that exercises the staged production runner with no explicit story IDs so that path is covered directly.

Focused story-runner contract, Bash syntax, and `git diff --check` pass. ShellCheck wasn’t available in the current Windows/WSL environment, so I didn’t claim a rerun.

I also left the optional exec-11 and broader FAIL-assertion points unchanged.

And once this one is settled, if there’s an issue or roadmap item where another contributor would be useful, feel free to point me at it. Happy to keep helping.

vladimirrott (approved):
Approving at `06293a9`. You fixed more than I asked for.

I flagged story 28 in `dev-stories.sh:221`. You found the second copy in `run-stories.sh:228`, corrected the NOTE that claimed those stories "should always pass on any Linux host", and then wrote a regression test that derives its fixture set from the story files on disk instead of restating the curated list. That last choice is the one worth calling out: a test that retyped the list would agree with itself forever.

## What I ran

Everything in a container, because this diff executes: `podman run --rm --network=none -v "$W:/repo:z" docker.io/library/python:3-slim`. No network, nothing from my home directory reachable.

| Mutation | Result |
|---|---|
| re-add 28 to `run-stories.sh` curated set | **red**, 2 assertions |
| revert the exec runner's aggregate gate | **red** |
| unmutated | passes |

The first one is the interesting output, because it fails on both halves of what you wrote:

```
FAIL: run-stories default set accepts portable stories (expected exit 0, got 1)
FAIL: run-stories default set omits Fedora-only story 28 (output contained Story 28)
```

Full board green at `06293a9`, including the new `story-runner-verdicts.test.sh` step inside `docs-and-hygiene`.

## One gap, not blocking

The bug lived in two files. The guard covers one.

```
re-add 28 to run-stories.sh   -> RED
re-add 28 to dev-stories.sh   -> GREEN
```

`dev-stories.sh`'s default set can drift back and nothing notices, because the three dev cases all pass explicit story numbers and never exercise the no-argument path. Not blocking: the code is fixed in both places and this is coverage, not a defect. It is also ten lines of the technique you already wrote, applied to the dev runner, so I would rather offer it to you than hold a finished PR for a third round.

## What I changed

I added the CHANGELOG entry myself at `06293a9`, since guessing a maintainer's release wording is not your job. Read it and say if it misstates anything.

## Next

If you want the `dev-stories.sh` half, say so here and I will open it as an issue with your name on it. Otherwise **#252** is the closest fit to what you have just demonstrated: two parsers read the story header and nine tag spellings pass unchecked, which is the same "two copies of one contract" shape you just found by yourself in this PR.

Thanks for a genuinely well-tested first contribution.

bferanmi806-sketch:
Yep, the CHANGELOG looks right to me.

I’d be happy to take the dev-stories.sh follow-up, so feel free to open that and put my name on it.

I’d also be up for #252 if it’s still available. I saw you’d mentioned it to George earlier, so if he’s taking it no worries otherwise I’m happy to work through both.

Thanks for the reviews too, I learned quite a bit from this one.

bferanmi806-sketch:
@vladimirrott I went ahead and implemented the dev-stories.sh regression follow-up you offered here: https://github.com/lacs-project/sysknife/pull/353

Issue #344, opened 2026-09-02
A contributor who cannot build the workspace has no supported way to update the test-count artifact


vladimirrott:
`scripts/test_baseline.sh` is the only sanctioned path from a measured test count to the four places that publish it. It needs `cargo-nextest` and a full workspace build, and `apps/sysknife-shell` pulls in WebKit, GTK, AppIndicator and librsvg. A contributor without those cannot regenerate the artifact, and has no supported way to say so.

## Why it matters

Three contributors hit this in ten days, and the workaround each reached for is the one #278 exists to prevent.

- #340 recorded `1831` from the known Linux baseline plus the two tests the PR adds, because the author's macOS run stopped on Ubuntu-only tests. The number was right. `commit` and `measured_at` still described a run that never happened.
- #335 is stalled on the question in as many words: *"WSL can run the real-binary tests, focused unit tests, formatting, package-scoped Clippy, and the docs/evidence checks, but it lacks cargo-nextest plus the WebKit/GTK/AppIndicator/librsvg development libraries... This autonomous environment is not permitted to install system packages."*
- #341 could not run the GUI half either and said so.

Every one of them chose honesty over a green board, which is the behaviour the repo wants. The repo answers it by making them guess a number or stop.

## Scope

Have CI measure it and write it back, so the artifact is always machine-measured at the commit it names.

**The obvious implementation is a security bug, so name it here.** A workflow with write permission that runs on `pull_request_target`, or one that checks out the fork's head and then pushes, executes untrusted code with a writable token. Do not do that. The shape that works:

1. A `pull_request` job, no secrets, no write permission, runs `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` and uploads `tests/evidence/workspace-tests.json` as an artifact.
2. A separate `workflow_run` job, triggered by the first completing, downloads the artifact, and pushes the file to the PR branch. It never checks out fork code.

Step 2 needs `maintainer_can_modify`, which fork PRs set by default.

If that is more than the issue is worth, the smaller version is a documented fallback: leave the four files untouched, say in the PR body which commands you could not run and why, and let a maintainer regenerate before merge. Write it into `CONTRIBUTING.md` so nobody has to ask.

## Also worth fixing while here

`cargo install cargo-nextest --locked` needs no system packages. `CONTRIBUTING.md` does not say so, and two contributors assumed nextest was as unreachable as the GTK stack.

## Tests first

[...]

PR #348, opened 2026-09-03
feat(mcp): expose direct read-only actions

### LINKED ISSUE #216: Expose the read-only actions as MCP tools without exposing AptUpdate
`sysknife mcp-server` exposes five tools: `sysknife_plan`, `sysknife_execute`, `sysknife_history`, `sysknife_doctor` and `sysknife_audit_verify` (`apps/sysknife-cli/src/mcp_server.rs`, pinned by `apps/sysknife-cli/tests/mcp_stdio.rs`). Reading live system state means going through `sysknife_plan`, so answering "how much disk is free" costs a round trip to a language model.

The roadmap asks for the Observer-level actions to be exposed as direct read-only tools. Two things need settling first, and the second one is where the work is.

## "Observer-level" is a role, not a risk tier

`Observer` is a `CallerRole`. `policy::min_role_for_action` derives it from the action's `RiskLevel` through `role_for_risk_level`: `Low` gives `Observer`, `Medium` gives `Dev`, `High` gives `Admin`. One monotonic exception lifts `ListJobHistory` into the set, since it has no `ActionSpec` of its own.

Counting from `docs/action-reference.md`, which is machine-generated from the live catalogue with a drift guard:

| | count |
|---|---|
| `RiskLevel::Low` | 62 |
| plus `ListJobHistory` | 63 Observer-callable |
| of those, family `All` | 33 |
| family `Ubuntu` | 22 |
| family `Fedora` | 7 |

So an Ubuntu host can route 55 of the 63, and the seven Fedora-only ones are refused by the distro fence before they reach the executor. Decide early whether the tool list is static or filtered by the detected host, because a tool that always errors is worse than a tool that is absent.

(The `~59` in `ROADMAP.md` is stale.)

## Low risk does not mean read-only

`apps/sysknife-cli/src/distro_routing.rs:66` already says so:

```rust
// the RBAC role is a bad proxy for "does this mutate" — `AptUpdate` is
// Low/Observer and runs `sudo apt-get update`.
```

Register all 63 as read-only MCP tools and a mutating action lands on a surface that carries no approval gate. Working out how a genuinely read-only action gets identified is the substance of this issue. Whatever rule you settle on has to be enforced by a test, because a hand-maintained allowlist drifts the first time somebody adds an action and forgets.

## Do not write 63 `#[tool]` methods

`rmcp` 2.2.0 exposes `ToolRouter` with a runtime API, so the router can be built in a loop over the catalogue:

- `ToolRouter::with_route` (builder) and `ToolRouter::add_route` (mutating), `src/handler/server/router/tool.rs:393` and `:437`
- the `IntoToolRoute` trait at `:217`
- `ToolRouter::list_all` at `:582`
- `Add` and `AddAssign` for composing two routers, at `:605` and `:617`
- `#[tool_handler(router = <expr>)]` to point the generated handler at your router

Generating from the catalogue means a new read-only action reaches MCP without a second edit somewhere else.

## Constraints

Mutating actions stay plan-only. The approval gate is the product, and no MCP tool may route around it. Extend `mcp_stdio.rs` rather than replacing it, and make it prove that nothing outside the read-only set reached the router.

## Difficulty

`medium`. The code is mechanical once the read-only rule is decided. Deciding it is the part that needs care.

## Getting started

Comment here before you start if you want to talk the read-only rule through. That conversation is worth having before any code exists.

QinXi-ai:
## Summary

Expose Observer-level read-only actions as direct MCP tools generated from the live action catalogue and filtered for the detected distro. The explicit classification keeps `AptUpdate` plan-only even though it is Low/Observer, and a drift test fails when future Observer actions are not classified.

Closes #216.

## Changes

- compose runtime `ToolRouter` routes for catalogue-backed read-only queries
- preserve daemon `query_action`, RBAC, validation, and distro fences
- annotate generated tools as read-only/non-destructive
- extend unit and real stdio contracts, docs, CLI help, and changelog

## Test plan

- [x] classification parity: 62 read-only + `AptUpdate` = all 63 Observer-callable actions
- [x] markdownlint on modified user-facing docs
- [ ] Linux `cargo test`/clippy via CI (the Windows host cannot compile the existing Unix-only `vsock` dependency)


vladimirrott (commented):
Reviewed at `3eae12e0`. The design is right, and the one thing I went looking for is not there, which is the good outcome.

**The action name is not caller-supplied.** Each route captures its own `routed_action` and the client passes only `params`, so there is no generic `query_action` passthrough for a client to aim at a mutating action. Three independent filters have to agree before a route exists at all (`risk_level == Low`, membership in `MCP_READ_ONLY_ACTIONS`, absence from `OBSERVER_MUTATING_ACTIONS`), and the third one exists purely to fail closed if a name is ever copied onto both lists. That comment is doing real work.

**The drift test bites.** I dropped `"GetSystemState"` from `MCP_READ_ONLY_ACTIONS` to simulate an Observer action nobody classified, and rebuilt:

```
failures:
    mcp_server::tests::observer_actions_are_fully_classified
    mcp_server::tests::direct_router_filters_by_distro_and_never_exposes_mutations
```

An unclassified future action is both invisible to the router and loud in CI. That is the property that makes this maintainable after you and I have both forgotten the details.

Parameter shapes stay unvalidated at the MCP layer (`additionalProperties: true`), which I read as deliberate: the daemon is the authority on typed actions and runs the real validators, and duplicating them here would create a second copy to drift. Worth one line in the module docs saying so, so the next reader does not read the open schema as an oversight.

## What is actually blocking

Only the conflict. The branch is behind `main` and `git` reports `DIRTY`; it is the `CHANGELOG.md` section, because 0.12.0 shipped while this was open. Merge `main` in, resolve that hunk, and I will re-pin the head and merge.

## The failing test you will see is mine, not yours

If you run `cargo test` locally, `mcp_server::tests::mcp_tools_integrate_with_a_daemon_over_the_socket` fails intermittently. I chased it because it appeared on your branch, and it is a pre-existing bug on `main`:

| Tree | `cargo test` (parallel) | single-threaded |
|---|---|---|
| `main` | **4 of 4 runs failed** | 265 passed |
| this branch | 2 of 3 runs failed | 265 passed |

The test calls `std::env::set_var("SYSKNIFE_SOCKET", ...)`, which is process-global, and other tests in the same binary race it. CI never sees it because `cargo nextest` runs each test in its own process. Your change lengthens the window slightly by adding a second call after the `set_var`, and it fails *less* often on your branch than on `main`, so please do not spend time on it. I am opening it separately.

Use `cargo nextest run --workspace --locked`, which is what CI and `scripts/ci-local.sh` run.

## Thank you

You picked up #216 from the Call for Participation listing and came back with the version that has a drift test in it. Most contributors write the allowlist and stop. If you keep Linux machines anywhere outside this repo, I would like to know how these tools feel from inside a real MCP client, since you have now read more of that surface than anyone but me. `npx sysknife-setup`, and everything in this PR is read-only by construction.

vladimirrott:
Following up on my own review at this same head, `3eae12e0`, because I got two things wrong in it and one of them would have cost you a CI round trip.

## The conflict is five files, not one

I named `CHANGELOG.md` and put it down to 0.12.0 shipping while this was open. `git merge-tree` against `main` at `f929272b` reports five:

```
CHANGELOG.md
README.md
docs/distro-support.md
docs/introduction.md
tests/evidence/workspace-tests.json
```

**The test count has moved under you.** Your base was 1,831. `main` is at 1,833, because #340 and #353 each landed a test since. You add three (`observer_actions_are_fully_classified`, `direct_tool_names_are_unique_and_described`, `direct_router_filters_by_distro_and_never_exposes_mutations`), so after the rebase the figure is **1,836**, not the 1,834 you measured. Four places carry it, and the artifact has to be generated rather than hand-edited:

```
```

then `1,836 Rust tests` in `README.md`, `docs/introduction.md` and `docs/distro-support.md`. Resolving only the `CHANGELOG.md` hunk and keeping your side of `workspace-tests.json` turns the `rust` job red on the recorded baseline. My miss: I told you one file.

Your green board was also computed at 04:19Z today against a merge commit pinned before those seven commits landed, so it does not describe the tree that would merge now. Only a fresh push rebuilds the merge ref, which is another reason the rebase comes first.

## One review item I should have checked the first time

#216 asks that the stdio test "prove that nothing outside the read-only set reached the router". The replacement in `mcp_stdio.rs` asserts seven names present and two absent, so a mutating action arriving on the approval-free surface is caught only if it is called `apt_update` or `apt_install`. The old exact-set assertion could not survive a distro-dependent tool list and you were right to retire it; the upper bound it carried still needs a home.

`direct_router_filters_by_distro_and_never_exposes_mutations` is the cheap place, because it can see the constant that a black-box stdio test cannot. It spot-checks four names per distro today. One assertion in the other direction closes the gap: every name the router advertises must appear in the snake_case image of `MCP_READ_ONLY_ACTIONS`. That copies none of the filter logic, and it goes red the moment a route exists that the allowlist does not name.

That is the whole blocking set: the rebase, and that one assertion.

## Optional, and I would merge without them

**A distinction I ran past on the open schemas.** Reading `additionalProperties: true` as a refusal to duplicate the daemon's validators is right, and I stand by it. Parameter *names* are a different thing from parameter validation, though. Of the 61 tools the router can generate, 21 read a parameter, and 10 of those cannot run without one:

```rust
"GetServiceLogs" => {
    let unit = validated_unit_name(require_str(params, "unit")?, "unit")?;
"AptShow" => {
    let package = validated_safe_arg(require_str(params, "package")?, "package")?;
```

`sysknife_get_service_logs` advertises no `unit`, so an agent learns the name by calling the tool and reading the failure. The repo already declares that shape for the same action, in `crates/sysknife-brain/src/planning_tools/query_tools.rs`:

```rust
"properties": { "unit": { "type": "string", "description": "The systemd unit name (e.g. 'sshd.service')" } },
"required": ["unit"]
```

Adding `properties` while leaving `additionalProperties: true` alone would hand the agent the names and keep the daemon the only validator. Worth noting that the 40 no-argument actions and the 21 parameterised ones share one schema, and the existing assertion in `mcp_stdio.rs` (`inputSchema.is_object()`) passes for both while its comment argues for the opposite.

**Error class.** `direct_query_inner` maps every failure through `ErrorData::internal_error`, so a caller who omits `unit` is told the server broke. `ErrorData::invalid_params` exists in rmcp 3.1.0 (`model.rs:633`) and is the honest class for a `validation_failure` coming back from `build_action_spec`.

**`ListJobHistory` never becomes a tool.** It sits in `MCP_READ_ONLY_ACTIONS`, the router iterates `all_specs()`, and it has no spec of its own. Your comment says so, and the entry earns its keep in the classification test. A future reader building an allowlist from that constant will still be caught out by it.

**Silent shrink on detection failure.** `detect().ok()` drops the reason, and the tool list falls back to the cross-distro subset. One line on stderr would turn "the apt tools are missing" from a mystery into a diagnosis.

## Two things that are not yours

The intermittent `mcp_tools_integrate_with_a_daemon_over_the_socket` failure I described is now #356, so it is tracked and off your plate.

I reviewed this by reading it. Fork branches are not built or tested on my machine, so the arithmetic above comes from `git merge-tree`, from counting `risk_level: RiskLevel::Low,` in the catalogue on `main` (62, which is what your test asserts), and from the CI run metadata. Standard for every fork PR here.

## Next, if you want it

#226 is yours if you say the word and I will hold it. It is the other half of the surface you have read end to end: `handle_query_action` gates on `min_role_for_action == Observer`, `AptUpdate` is Low and therefore Observer, so the daemon will run `sudo apt-get update` through the query path with no preview and no receipt. Your PR keeps that off MCP by never generating the route. The daemon-side hole is still open, and you now know why `OBSERVER_MUTATING_ACTIONS` has to exist.

vladimirrott:
Checking in with one number, because it moved again while you were working.

`083d582` closes the blocking review item. The bound is the right shape: every name the router advertises has to appear in the snake_case image of `MCP_READ_ONLY_ACTIONS`, checked across all three distro variants, and it copies none of the filter's logic. A route the allowlist does not name now turns the test red.

So the rebase is the only thing left.

## The baseline is 1,837, not 1,836

You merged `main` at `dc72617` when it was `f929272` and the count was 1,833. #359 landed after that and added one, so `main` is now 1,834. Your three tests put the merged tree at **1,837**.

You are one commit behind. After merging `main` again:

```sh
```

then `1,837 Rust tests` in `README.md`, `docs/introduction.md` and `docs/distro-support.md`. Generate the artifact rather than editing the number by hand: `commit` and `measured_at` have to describe a run that actually happened, and the `rust` job checks the figure against the real suite.

The conflicting files are the same four as before, minus `CHANGELOG.md`, which your merge already resolved:

```
README.md
docs/distro-support.md
docs/introduction.md
tests/evidence/workspace-tests.json
```

## If regenerating the artifact is the blocker, say so

You mentioned earlier that the Windows host cannot compile the Unix-only `vsock` dependency. If you also cannot run the full workspace, you cannot produce that number honestly, and I would rather you said so than guessed it. That is #344, it is a real gap in this repo rather than anything you did, and three contributors have hit it in ten days. Tell me and I will push the artifact commit onto your branch myself.

No rush from my side. Nothing else is queued behind this.

QinXi-ai:
Thanks for the correction and the precise blocking set. Both are addressed at `0e37a664`.

- Merged `main` through `0640818d` and resolved the five-file conflict set. Because #359 landed one more test while CI was running, the merged baseline is now 1,837 rather than 1,836. The evidence artifact was regenerated against `9d35a390`, whose Linux nextest run completed 1,837/1,837, and all four published figures now say 1,837.
- Added the missing upper bound in `direct_router_filters_by_distro_and_never_exposes_mutations`: for Ubuntu, Fedora, and detection failure, every routed name must be in the snake_case image of `MCP_READ_ONLY_ACTIONS`. The existing presence and distro-fence spot checks remain.
- Added the module-doc note that the open direct-query schemas are deliberate because typed validation stays authoritative in the daemon.

The fresh board at `0e37a664` is green: Rust/nextest, docs and hygiene, frontend, Postgres contract, scripts lint, security audit, dependency review, TruffleHog, and both CodeQL analyses passed; container smoke is skipped by the workflow.

I left the optional parameter-name schemas, error-class refinement, `ListJobHistory` representation, and detection diagnostic out of this PR to keep the reviewed boundary change focused. I will follow up on #226 after this lands.

vladimirrott (approved):
`083d582` closes the blocking item, and I proved it rather than reading it.

## The bound works, and it is worth being precise about what it guards

First mutation, at `083d582` in a container: I removed the allowlist condition from the router filter, so every `Low` action gets a route.

```
test direct_router_filters_by_distro_and_never_exposes_mutations ... ok
```

Green, and that is correct rather than a gap. `MCP_READ_ONLY_ACTIONS` is today exactly the set of `Low` actions minus `AptUpdate`, and `OBSERVER_MUTATING_ACTIONS` is checked separately, so dropping the allowlist from the filter changes nothing observable. The two sets are equal by construction.

Second mutation, the shape the assertion is actually written for, a route existing that the allowlist does not name:

```
panicked at apps/sysknife-cli/src/mcp_server.rs:1906:13:
approval-free router exposed tools outside MCP_READ_ONLY_ACTIONS: ["sysknife_get_system_state"]
```

So the bound is a guard against a future filter change, not against today's state, and it names the offending tool when it fires. That is the right kind of guard to leave behind, and it is worth knowing it is quiet now by construction rather than by accident.

Together with the drift test, the approval-free surface is now pinned from both directions: `observer_actions_are_fully_classified` catches a new Observer action nobody classified, and this catches a route the allowlist does not name.

## Approving

The only thing left is mechanical. You are one commit behind, and the baseline is 1,837 rather than the 1,836 in your artifact, because #359 landed after your merge.

**I have armed the merge.** When you push the rebase, a watcher re-pins your head, re-runs the drift mutation above against the merged tree, checks the board is fully green with the review still approved, and merges. If the push turns out to carry production changes beyond the rebase, it will not merge; it will leave the PR for me to read properly, which is the only honest way to automate this.

If regenerating the artifact is what blocks you, say so and I will push that commit onto your branch. That is #344, a real gap in this repo, and I would rather you said so than guessed a number.

vladimirrott:
Merged in `60af0ad`.

You carried this through a five-file conflict set, a mid-flight `main` that moved twice under you, and a blocking review item, and the final head was clean on all three. The bound on the router is the right shape: every name it advertises has to appear in the snake_case image, so a new tool cannot be added without the check noticing.

One thing that was not your fault and cost you time anyway: `cargo test -p sysknife-cli --bins` fails intermittently on `main` from a process-global `set_var` race, and it names a function your PR touched. #356 tracks it. I re-measured it today and it fires about 17% of the time, so if you saw it once and then could not reproduce it, that is why.

PR #349, opened 2026-09-03
fix(tests): mask empty grep results to show error messages

### LINKED ISSUE #347: Five hygiene guards exit 1 with no output when their grep anchor moves
`docs-and-hygiene` runs 23 host-side guards. Five of them exit 1 without printing anything when the code fact they read moves, so the job goes red with an empty step log and the contributor has nothing to work from.

Each of the five derives a value with a grep pipeline, then checks it for emptiness and prints a diagnostic:

```bash
require_token="$(grep -o 'SYSKNIFE_REQUIRE_POSTGRES' "$store" | sort -u)"
if [ -z "$require_token" ]; then
    printf 'FAIL: %s no longer names SYSKNIFE_REQUIRE_POSTGRES; the job token cannot be derived\n' \
        "$store" >&2
    exit 1
fi
```

Under `set -euo pipefail`, a `grep` that matches nothing exits 1, `pipefail` promotes that to the pipeline's status, and `set -e` kills the script at the assignment. The `if` never runs. That `printf` is unreachable in all five.

Measured on 5a9450a, each with its grep anchor moved in a scratch tree:

| Guard | Broken input | rc | bytes of output |
|---|---|---|---|
| `postgres-contract-guard.test.sh:49` | env var renamed in `postgres_store.rs` | 1 | 0 |
| `audit-export-confidentiality.test.sh:32` | `compute_request_hash` renamed | 1 | 0 |
| `database-path-agreement.test.sh:44` | fallback path literal changed | 1 | 0 |
| `smithery-manifest.test.sh:48` | `startCommand.type` key renamed | 1 | 0 |
| `public-claims.test.sh:172` | `N typed actions` rephrased in `docs/introduction.md` | 1 | 0 for the failure |

`public-claims.test.sh` costs the most. It dies two thirds of the way through its own mutation suite, so thirteen later `assert_rejected` cases never run, and the last two lines on stdout are the checker's success banner:

```
```

`bash -x` puts the death at `actions=` with an empty value, immediately after the `ROADMAP.md` case restores itself.

The direction is safe. All five fail closed and the gate does go red. The cost is diagnostic, and it lands on whoever renamed a symbol and now reads a red required check with a blank log.

## Fix

Mask the pipeline so the emptiness check the authors already wrote can run:

```diff
--- a/tests/release/postgres-contract-guard.test.sh
+++ b/tests/release/postgres-contract-guard.test.sh
@@ -46,7 +46,7 @@
 # Fail-closed env var, taken from the store test the job runs.
 # sort -u consumes every grep hit so a later head cannot SIGPIPE the pipeline
 # under `set -o pipefail`.
-require_token="$(grep -o 'SYSKNIFE_REQUIRE_POSTGRES' "$store" | sort -u)"
+require_token="$(grep -o 'SYSKNIFE_REQUIRE_POSTGRES' "$store" | sort -u || true)"
 if [ -z "$require_token" ]; then
     printf 'FAIL: %s no longer names SYSKNIFE_REQUIRE_POSTGRES; the job token cannot be derived\n' \
         "$store" >&2
--- a/tests/release/audit-export-confidentiality.test.sh
+++ b/tests/release/audit-export-confidentiality.test.sh
@@ -29,8 +29,8 @@
 report() { printf 'FAIL  %s\n' "$1" >&2; failures=$((failures + 1)); }
 
 # The code fact. An empty line number is a failure, not a pass over nothing.
-hash_line="$(grep -n 'let request_hash = compute_request_hash(' "$dispatcher" | head -1 | cut -d: -f1)"
-redact_line="$(grep -n 'let redacted_params = redact_params(' "$dispatcher" | head -1 | cut -d: -f1)"
+hash_line="$(grep -n 'let request_hash = compute_request_hash(' "$dispatcher" | head -1 | cut -d: -f1 || true)"
+redact_line="$(grep -n 'let redacted_params = redact_params(' "$dispatcher" | head -1 | cut -d: -f1 || true)"
 
 if [ -z "$hash_line" ] || [ -z "$redact_line" ]; then
     printf 'FAIL: cannot locate compute_request_hash / redact_params in %s\n' "$dispatcher" >&2
--- a/tests/release/database-path-agreement.test.sh
+++ b/tests/release/database-path-agreement.test.sh
@@ -41,7 +41,7 @@
 
 # The daemon's answer, read out of the source rather than executed, so this stays
 # a fast host-side check that needs no build.
-rs_suffix="$(grep -oE '"\.local/state/sysknife/daemon\.sqlite"|"\.local/share/sysknife/daemon\.sqlite"' "$core_rs" | head -1 | tr -d '"')"
+rs_suffix="$(grep -oE '"\.local/state/sysknife/daemon\.sqlite"|"\.local/share/sysknife/daemon\.sqlite"' "$core_rs" | head -1 | tr -d '"' || true)"
 if [ -z "$rs_suffix" ]; then
     printf 'FAIL: could not find the fallback database path in %s.\n' "$core_rs" >&2
     printf '      default_database_path() changed shape; update this guard rather than deleting it.\n' >&2
--- a/tests/release/public-claims.test.sh
+++ b/tests/release/public-claims.test.sh
@@ -169,7 +169,7 @@
 mv "$fixture/ROADMAP.held" "$fixture/ROADMAP.md"
 
 # The action count is derived from the catalogue source, so a retyped one fails.
-actions="$(grep -oE '[0-9]+ typed actions' "$fixture/docs/introduction.md" | head -1 | grep -oE '^[0-9]+')"
+actions="$(grep -oE '[0-9]+ typed actions' "$fixture/docs/introduction.md" | head -1 | grep -oE '^[0-9]+' || true)"
 if [[ -z "$actions" ]]; then
     printf 'FAIL: could not find the typed-action count in the fixture\n' >&2
     exit 1
--- a/tests/release/smithery-manifest.test.sh
+++ b/tests/release/smithery-manifest.test.sh
@@ -45,7 +45,7 @@
     fail 'smithery.yaml declares a runtime; hosted runtimes need Streamable HTTP and a daemonless container cannot serve SysKnife'
 fi
 
-start_type="$(grep -A2 '^startCommand:' "$manifest" | grep -oP '^\s*type:\s*"?\K[a-z]+' | head -1)"
+start_type="$(grep -A2 '^startCommand:' "$manifest" | grep -oP '^\s*type:\s*"?\K[a-z]+' | head -1 || true)"
 [[ "$start_type" == "stdio" ]] \
     || fail "startCommand.type is '${start_type:-<missing>}', expected stdio"
 
@@ -56,7 +56,7 @@
 grep -q "command: *'sysknife'" "$manifest" \
     || fail "commandFunction does not spawn the 'sysknife' binary"
 
-subcommand="$(grep -oP "args: *\[ *'\K[a-z-]+" "$manifest" | head -1)"
+subcommand="$(grep -oP "args: *\[ *'\K[a-z-]+" "$manifest" | head -1 || true)"
 [[ -n "$subcommand" ]] || fail 'commandFunction passes no subcommand'
 
 # Cross-check against the CLI itself: #[command(name = "mcp-server")] is the
```

Both directions verified in a scratch tree at 5a9450a. On clean input all five still pass. On the same five broken inputs from the table, all five now print their intended message, including the `${start_type:-<missing>}` fallback in `smithery-manifest.test.sh` that no run could reach before.

## Tests first

`story-metadata.test.sh` is the shape that needs no fix: its header parser reports `unparseable story header in ...` before any derived count is consulted, so it already fails loudly. A new guard should either do that or keep its emptiness check reachable.

A guard over the guards would catch the sixth one. For every `X="$(... grep ...)"` in `tests/release/*.test.sh` and `tests/e2e/*.test.sh` that runs under `set -e` with `pipefail`, require either `|| true` on the pipeline or a trailing command that cannot exit non-zero on a no-match. Prove it bites by dropping the `|| true` from one of the five above and watching the check name that line. Assert the extraction found something first: a pattern matching no assignments makes the check pass over an empty set, which is this same bug in a different costume.

## Difficulty

`good first issue` for the patch above: six one-token edits, and the verification is a `sed` per guard plus a rerun. The guard over the guards is `medium` and worth splitting into its own issue.

## Getting started

iamjaeholee:
## Summary

Added `|| true` to the six places mentioned in the issue. Now empty grep results don't kill the script, and the error messages show up correctly.

 I intentionally broke the code in all 5 files one by one and ran them. I checked with my own eyes that the `FAIL:` messages are printed successfully.

## Related Issue

Closes #347

## Validation

- [x] Tests added or updated
- [ ] Security impact considered
- [ ] Trust boundary preserved (daemon remains the only privileged executor)
- [x] CI passes

## Notes for Reviewers

Call out any tradeoffs, migrations, or follow-up work.

### DIFF
--- a/tests/release/audit-export-confidentiality.test.sh
+++ b/tests/release/audit-export-confidentiality.test.sh
@@ -29,8 +29,8 @@ failures=0
 report() { printf 'FAIL  %s\n' "$1" >&2; failures=$((failures + 1)); }
 
 # The code fact. An empty line number is a failure, not a pass over nothing.
-hash_line="$(grep -n 'let request_hash = compute_request_hash(' "$dispatcher" | head -1 | cut -d: -f1)"
-redact_line="$(grep -n 'let redacted_params = redact_params(' "$dispatcher" | head -1 | cut -d: -f1)"
+hash_line="$(grep -n 'let request_hash = compute_request_hash(' "$dispatcher" | head -1 | cut -d: -f1 || true)"
+redact_line="$(grep -n 'let redacted_params = redact_params(' "$dispatcher" | head -1 | cut -d: -f1 || true)"
 
 if [ -z "$hash_line" ] || [ -z "$redact_line" ]; then
     printf 'FAIL: cannot locate compute_request_hash / redact_params in %s\n' "$dispatcher" >&2
--- a/tests/release/database-path-agreement.test.sh
+++ b/tests/release/database-path-agreement.test.sh
@@ -41,7 +41,7 @@ js_path="$(HOME=/home/testuser node -e "
 
 # The daemon's answer, read out of the source rather than executed, so this stays
 # a fast host-side check that needs no build.
-rs_suffix="$(grep -oE '"\.local/state/sysknife/daemon\.sqlite"|"\.local/share/sysknife/daemon\.sqlite"' "$core_rs" | head -1 | tr -d '"')"
+rs_suffix="$(grep -oE '"\.local/state/sysknife/daemon\.sqlite"|"\.local/share/sysknife/daemon\.sqlite"' "$core_rs" | head -1 | tr -d '"' || true)"
 if [ -z "$rs_suffix" ]; then
     printf 'FAIL: could not find the fallback database path in %s.\n' "$core_rs" >&2
     printf '      default_database_path() changed shape; update this guard rather than deleting it.\n' >&2
--- a/tests/release/postgres-contract-guard.test.sh
+++ b/tests/release/postgres-contract-guard.test.sh
@@ -46,7 +46,7 @@ done
 # Fail-closed env var, taken from the store test the job runs.
 # sort -u consumes every grep hit so a later head cannot SIGPIPE the pipeline
 # under `set -o pipefail`.
-require_token="$(grep -o 'SYSKNIFE_REQUIRE_POSTGRES' "$store" | sort -u)"
+require_token="$(grep -o 'SYSKNIFE_REQUIRE_POSTGRES' "$store" | sort -u || true)"
 if [ -z "$require_token" ]; then
     printf 'FAIL: %s no longer names SYSKNIFE_REQUIRE_POSTGRES; the job token cannot be derived\n' \
         "$store" >&2
--- a/tests/release/public-claims.test.sh
+++ b/tests/release/public-claims.test.sh
@@ -206,7 +206,7 @@ assert_rejected 'a claim file missing from the checked surface'
 mv "$fixture/ROADMAP.held" "$fixture/ROADMAP.md"
 
 # The action count is derived from the catalogue source, so a retyped one fails.
-actions="$(grep -oE '[0-9]+ typed actions' "$fixture/docs/introduction.md" | head -1 | grep -oE '^[0-9]+')"
+actions="$(grep -oE '[0-9]+ typed actions' "$fixture/docs/introduction.md" | head -1 | grep -oE '^[0-9]+' || true)"
 if [[ -z "$actions" ]]; then
     printf 'FAIL: could not find the typed-action count in the fixture\n' >&2
     exit 1
--- a/tests/release/smithery-manifest.test.sh
+++ b/tests/release/smithery-manifest.test.sh
@@ -45,7 +45,7 @@ if grep -qE '^[[:space:]]*runtime:' "$manifest"; then
     fail 'smithery.yaml declares a runtime; hosted runtimes need Streamable HTTP and a daemonless container cannot serve SysKnife'
 fi
 
-start_type="$(grep -A2 '^startCommand:' "$manifest" | grep -oP '^\s*type:\s*"?\K[a-z]+' | head -1)"
+start_type="$(grep -A2 '^startCommand:' "$manifest" | grep -oP '^\s*type:\s*"?\K[a-z]+' | head -1 || true)"
 [[ "$start_type" == "stdio" ]] \
     || fail "startCommand.type is '${start_type:-<missing>}', expected stdio"
 
@@ -56,7 +56,7 @@ grep -q 'commandFunction:' "$manifest" \
 grep -q "command: *'sysknife'" "$manifest" \
     || fail "commandFunction does not spawn the 'sysknife' binary"
 
-subcommand="$(grep -oP "args: *\[ *'\K[a-z-]+" "$manifest" | head -1)"
+subcommand="$(grep -oP "args: *\[ *'\K[a-z-]+" "$manifest" | head -1 || true)"
 [[ -n "$subcommand" ]] || fail 'commandFunction passes no subcommand'
 
 # Cross-check against the CLI itself: #[command(name = "mcp-server")] is the


vladimirrott (approved):
Welcome, and thank you for this. You reproduced all five before writing a line, then fixed exactly those five and nothing else. That is a better first pull request than most.

I checked it rather than taking the diff's word for it. Both trees in a container, one guard's anchor moved at a time, at `dc4278b6`:

| Guard | `main` | this branch |
|---|---|---|
| `postgres-contract-guard` | rc=1, **0 bytes** | rc=1, names the file and `SYSKNIFE_REQUIRE_POSTGRES` |
| `audit-export-confidentiality` | rc=1, **0 bytes** | rc=1, `cannot locate compute_request_hash / redact_params in ...` |
| `database-path-agreement` | rc=1, **0 bytes** | rc=1, `could not find the fallback database path in ...` |
| `smithery-manifest` | rc=1, **0 bytes** | rc=1, `startCommand.type is '<missing>', expected stdio` |
| `public-claims` | rc=1, diagnostic unreachable | rc=1, `FAIL: could not find the typed-action count in the fixture` |

`public-claims` is the one that sold me. On `main` it dies two thirds of the way through its own mutation suite, so the last two lines a reader sees are its own success banners:

```
```

A red required check whose final output says everything matched. With your change the same run ends on the `FAIL` line instead.

I also checked that `|| true` opens no new hole. When the file is missing rather than the anchor moved, grep still writes to stderr and the `-z` branch still prints, so nothing gets quieter than it was.

## My issue undercounted, and that is on me

I wrote "five guards". There are four more with the same shape and the same one-token fix, and I missed them when I filed #347:

- **`scripts/check_release_versions.sh:62`.** Same experiment, anchor moved: rc=1, zero bytes, and its `No internal dependency pins found; the manifest paths are stale.` line never prints. That guard runs in three places, one of them the release workflow's preflight.
- **`tests/e2e/story-metadata.test.sh:31, 66 and 67`**, three `grep -c .` assignments. The comment above line 31 says a regex that quietly matched nothing would make every assertion below vacuous, and the guard written to catch that is itself unreachable when the parser returns nothing.

They are yours if you want them. Say so and I will open the issue with your name on it rather than putting it in front of anyone else. It is the change you have now made five times, so it should be a short second PR.

## One other thing

Few people read shell this closely. If you run Linux anywhere that matters at work, servers you patch, a fleet you keep alive, I would like to know how SysKnife holds up there. `npx sysknife-setup` gets you started, and the read-only side (`sysknife doctor`, `sysknife history`, `sysknife audit verify`) touches nothing on the host. A bug report from someone running it on a real machine is worth more to this project than any feature I could write, and you have already shown you report things precisely.

Either way, thanks for the fix. Merging once the board comes back.

vladimirrott:
Merged in `d98d102`, and it is on `main` now. Your first PR here.

I closed #347 with it. The five guards it named all print their diagnostic instead of dying silently, and the `public-claims` one no longer ends a red run with two success banners.

#355 is the follow-up with the four I missed, and it has your name on it. No obligation, and it stays yours until you say otherwise.

One practical note now that you have a merged PR: your fork's workflow runs no longer need my approval on every push, so your next one will start CI by itself.

PR #351, opened 2026-09-03
docs(packaging): stop claiming the karg charset matches validated_safe_arg

### LINKED ISSUE #339: sysknife-grub-kargs-edit's comment claims validated_safe_arg allows '=', and it does not
`packaging/sysknife-grub-kargs-edit` carries a comment that contradicts itself and misdescribes the validator it claims to mirror.

```python
# Defense-in-depth: each karg token is validated independently.
# Matches the same allowlist as validated_safe_arg in validate.rs:
#   ASCII alphanumeric, '.', '_', ':', '/', '+', '@', '=', '-'
# No leading dash (kernel args may contain '=' but cannot start with '-').
KARG_PATTERN = re.compile(r"^[A-Za-z0-9_.+@:/=-]+$")
```

Two problems in four lines.

**The allowlist it attributes to `validated_safe_arg` is wrong.** That function's charset is:

```rust
c.is_ascii_alphanumeric() || matches!(c, '.' | '_' | ':' | '/' | '+' | '@' | '-')
```

No `=`. The comment lists `'='` as one of the characters `validated_safe_arg` permits, and it does not.

**It says "the same allowlist" and then explains why it differs.** The next line says kernel arguments may contain `=`, which is the reason `KARG_PATTERN` is deliberately wider. So the regex is right and the first sentence is wrong about what it is matching.

## Not a vulnerability

The only caller, `GrubSetKargs` (`crates/sysknife-daemon/src/executor.rs`), runs `validated_safe_arg` on the value before this helper ever sees it, so the wider pattern is unreachable today. Filing it as an accuracy problem, not a hole.

It is worth fixing anyway. A comment that says "same as X" is the kind of thing a future contributor trusts instead of re-deriving, and if someone ever calls this helper directly, the comment is what tells them the pattern is safe when it is looser than the one it names.

## Scope

Reword the comment to say what is actually true: the pattern is `validated_safe_arg`'s charset plus `=`, because kernel arguments are `key=value` and the Rust-side validator rejects `=` for arguments that are not.

While there, check whether the two lists should be derived rather than restated. This is the third place in the repo where a charset is written down twice, and the other two have both drifted at some point.

## Difficulty

`easy`. One comment. The judgement is whether to also derive the charset, which is optional.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands. `tests/release/grub-kargs-edit.test.sh` covers this helper. No CLA and no copyright waiver. The project is MIT.

Georgefifth:
## Summary

The comment above `KARG_PATTERN` in `packaging/sysknife-grub-kargs-edit:43` claims the pattern "matches the same allowlist as `validated_safe_arg` in validate.rs" and lists `=` as part of that allowlist. Neither half is true: `validated_safe_arg` allows `.` `_` `:` `/` `+` `@` `-` and rejects `=`, and `executor.rs` documents that the rejection is deliberate — a comma or an `=` in a caller-supplied arg would inject extra tokens into the helper's comma-separated list. The helper, by contrast, must accept `=` because kernel arguments take the key=value form (`module.sig_enforce=1`), and the helper is directly sudo-invocable.

The fix is comment-only: state what the pattern is for and why it deliberately differs from `validated_safe_arg`, instead of claiming a parity that would be wrong to restore. No code changes; `KARG_PATTERN` itself is unchanged.

Per #339, the line 48 denylist comment is left alone.

## Related Issue

Closes #339

## Validation

- [x] Tests added or updated — no test change needed: comment-only fix; the existing `tests/release/grub-kargs-edit.test.sh` suite still passes locally (all 6 ok markers)
- [x] Documentation updated if behavior changed — behavior unchanged
- [x] Security impact considered — none: no validation logic touched
- [x] CI passes — pending; ran the release test and `python3 -m py_compile` locally

## Notes for Reviewers

Every factual claim in the new comment was checked against the current tree (`validated_safe_arg` at `crates/sysknife-daemon/src/actions/validate.rs`, the charset-check comment at the `GrubSetKargs` branch in `crates/sysknife-daemon/src/executor.rs`).

### DIFF
--- a/packaging/sysknife-grub-kargs-edit
+++ b/packaging/sysknife-grub-kargs-edit
@@ -40,9 +40,14 @@ GRUB_DEFAULT = "/etc/default/grub"
 GRUB_BACKUP = "/etc/default/grub.sysknife.bak"
 
 # Defense-in-depth: each karg token is validated independently.
-# Matches the same allowlist as validated_safe_arg in validate.rs:
-#   ASCII alphanumeric, '.', '_', ':', '/', '+', '@', '=', '-'
-# No leading dash (kernel args may contain '=' but cannot start with '-').
+# Deliberately wider than validated_safe_arg in validate.rs: kernel
+# arguments take the key=value form (module.sig_enforce=1), so '=' must be
+# accepted here, while validated_safe_arg rejects it. '=' stays out of that
+# validator because a comma or an '=' there would let a caller inject extra
+# tokens into the comma-separated list this helper receives — see the
+# charset-check comment at the GrubSetKargs branch in executor.rs. The
+# daemon screens every token before invoking the helper, and the denylist
+# below enforces the boot-security policy independently of it.
 KARG_PATTERN = re.compile(r"^[A-Za-z0-9_.+@:/=-]+$")
 
 # Defense-in-depth denylist — mirrors validated_safe_kernel_arg in executor.rs.


vladimirrott (approved):
Correct finding, and an easy one to walk past. `validated_safe_arg` accepts `. _ : / + @ -` and no `=`, while `KARG_PATTERN` is `^[A-Za-z0-9_.+@:/=-]+$`. The comment claimed a parity that never existed.

A comment on a helper reachable through its own sudo grant is load-bearing documentation, so I checked both things your new text asserts at `75afe94d`:

- `validate.rs:462` is `matches!(c, '.' | '_' | ':' | '/' | '+' | '@' | '-')`. No `=`. Confirmed.
- `executor.rs:2164` says "both callers run their charset check first. In particular `GrubSetKargs` runs `validated_safe_arg` (which already rejects `=` and `,`)". Your text cites a comment that exists and says what you say it says.

Both hold. Merging.

## The old comment made a second false claim

The line you deleted also said "No leading dash (kernel args may contain `=` but cannot start with `-`)". That was untrue too:

```
'-foo'                   MATCH
'--output=/etc/passwd'   MATCH
```

Nothing in the helper enforces it. Deleting the claim rather than inventing a replacement is the right call for this PR, and I checked the consequence before saying so: tokens arrive inside one `--append=<comma-list>` value and are split after `parse_args()`, so `argparse` never sees a dash-leading token as an option, and a `-foo` karg lands as text on the kernel command line where the kernel ignores it.

Harmless, then. Worth a sentence saying so eventually, because the next reader will ask.

## Optional, not blocking

"the denylist below enforces the boot-security policy independently of it" is a stronger verb than the code earns. The list is a denylist, so "screens every token against the boot-security denylist" says the same thing without promising more. Your call; I will merge either way.

You have three merged now, all of them in the packaging and validator area where the comments and the code drift apart. That is a good specialism to have on this project. If you administer Linux boxes outside this repo, I would like to know how SysKnife reads from the operator's side, `npx sysknife-setup` and the read-only tools change nothing on the host. The people who find the comment-versus-code gaps tend to find the operator-experience gaps too.

vladimirrott:
Merged in `f6cb807`, and I approved it without saying anything on the thread, which is a poor way to receive a first-rate patch.

What you caught is the kind of comment that survives review precisely because it sounds authoritative. `KARG_PATTERN` claimed to match `validated_safe_arg` and listed `=` as allowed; `validated_safe_arg` rejects `=`, and `executor.rs` documents that rejection as deliberate. A reader trusting that comment would have widened the pattern to match the claim and quietly undone a safety decision nobody would have thought to re-examine.

Eight lines, one file, and it closes a gap between two files that had no test between them. Thanks for reading the comment against the code instead of taking it at its word.

Georgefifth:
You asked in the #351 review how SysKnife reads from the operator's side, so I ran the experiment: fresh install on my daily-driver laptop (Ubuntu 24.04, no VM), user-mode daemon, ollama backend. Read-only only — no mutating actions, no system service. Honest report, good and bad.

**What worked, and worked well:**

- `npx sysknife-setup` is genuinely zero-friction: checksum-verified binaries, no sudo, user service up and socket reachable in one pass. It even noticed the socket was down mid-install and told me exactly how to start it. As an install experience on a machine I care about, it passed the "would I let this touch my box" test on the first run.
- `sysknife doctor` and the read-only MCP tools are the operator-facing highlight. 59 typed read-only queries, and every one I tried returned clean native output — disk usage, memory, host state, listening ports, pending-reboot. No LLM in that path, nothing to go wrong, and that determinism is exactly why I'd actually use these day to day.
- `sysknife audit verify` verified the (empty) chain and then volunteered: *"no independent checkpoint anchor is configured, so removal of the newest rows would not be detectable."* A security tool stating its own trust boundary unprompted is rare, and it landed well.

**What broke:**

The LLM planning path never completed a single plan against local ollama. `sysknife --dry-run "check disk usage of /"` fails with:

```
provider call failed (invalid response: model response contained only
unsupported content types (reasoning/image); no text or tool calls found)
```

Reproduced with both `qwen3:8b` (your default) and `qwen2.5:7b`, with and without `SYSKNIFE_OLLAMA_THINK=false`. The interesting part: hitting ollama's `/api/chat` directly — streaming, with tools, thinking on — returns the correct tool call every time:

```json
"tool_calls": [{"function": {"name": "GetDiskUsage", "arguments": {"path": "/"}}}]
```

So the model answers correctly and something between rig's streaming Ollama adapter and `from_rig_response` loses the tool calls. Your `openai_adapter.rs` header comment says you replaced rig on the OpenAI path for exactly this class of reasoning-item bug; the ollama path appears to be the last one still on rig's stream accumulator. I haven't traced it in rig itself — flagging it as an operator-side observation with the repro above, not a diagnosis. Happy to file it as an issue if that's useful.

Smaller rough edges, same session: `sysknife_get_firewall_state` failed with a bare `io error: No such file or directory (os error 2)` — it doesn't say which file, and on a box without the expected firewall backend that's a guessing game. And `SYSKNIFE_LLM_MODEL` set via `systemctl --user set-environment` was picked up by the daemon's doctor output but the CLI kept warning about falling back to its own default; the precedence there reads inconsistently from outside.

**The overall read:** the foundation — daemon, typed read-only catalogue, signed audit chain, install path — feels like something I could trust on a real host today. The LLM planning layer is the part that didn't survive first contact with an actual machine, and since that's the front door most people will try first, it's where I'd point the next round of hardening.

And cleanup behaved as advertised too: `--uninstall` removed the service, binaries and configs without drama.

PR #353, opened 2026-09-03
test(e2e): cover dev-stories default selection

bferanmi806-sketch:
This implements the `dev-stories.sh` regression follow-up @vladimirrott offered in #341.

## Summary

* Exercise `dev-stories.sh` through its no-argument portable default path.
* Derive staged fixtures from the checked-in story files instead of duplicating the curated default list.
* Verify Fedora-only story 28 is not selected by the default set.
* Make the regression suite fail if story 28 is added back to `dev-stories.sh`'s portable default set.

## Verification

* `bash tests/e2e/story-runner-verdicts.test.sh`
* `bash -n tests/e2e/story-runner-verdicts.test.sh`
* `git diff --check`
* Mutation proof: temporarily re-added story 28 to `dev-stories.sh`; the regression exited 1 and reported the unexpected Story 28 selection and SKIP result, then the mutation was reverted.

Reference #341 only

### DIFF
--- a/tests/e2e/story-runner-verdicts.test.sh
+++ b/tests/e2e/story-runner-verdicts.test.sh
@@ -66,6 +66,7 @@ run_case() {
         PATH="$fake_bin:$PATH" \
         SYSKNIFE_LLM_MODEL=fixture \
         SYSKNIFE_LLM_PROVIDER=ollama \
+        SYSKNIFE_ALLOW_DESTRUCTIVE=0 \
         SYSKNIFE_STORY_DELAY=0 \
         SYSKNIFE_STORY_DELAY_SECS=0 \
         SYSKNIFE_STORY_TIMEOUT=30 \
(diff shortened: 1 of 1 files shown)


vladimirrott (approved):
This is the gap I named when I merged #341, and you closed it the same day.

Mutation at `c129f8bb`: story 28 put back into the `dev-stories.sh` default set at line 221.

| Tree | Result |
|---|---|
| `main` | exit 0, green, **uncaught** |
| this branch | exit 1, two named failures |

```
FAIL: dev-stories default set accepts portable stories (expected exit 0, got 1)
FAIL: dev-stories default set omits Fedora-only story 28 (output contained Story 28)
```

Two assertions that fail for different reasons, which is what makes them worth having: the first catches a Fedora-only story breaking an Ubuntu run, the second catches it being selected at all. Either one alone would be weaker.

The part I like most is that the fixture set derives from `tests/e2e/stories/story-*.sh` on disk instead of restating the curated list. Adding a story next month cannot quietly make this test check less than it does today. Restating a list is exactly how the original bug survived in two files, so not reintroducing that shape inside the test itself is the right instinct.

Pinning `SYSKNIFE_ALLOW_DESTRUCTIVE=0` in `run_case` matters too. Without it, an exported variable in someone's shell flips the default set and the assertion silently tests the wrong list.

Merging.

## Next

You have now fixed a runner bug and then closed the coverage gap left behind it, which is the whole cycle. If you want another in that vein, **#347** is open: five hygiene guards that exit 1 with no output when their grep anchor moves. Someone is already on the five named in the issue, but I found four more of the same shape while reviewing it, and there is room for both of you. Say the word and I will scope the second half to you.

One aside: if you keep Linux machines running anywhere outside this repo, I would like to hear how SysKnife behaves there. `npx sysknife-setup`, and the read-only tools change nothing on the host. You have found two real bugs by running things carefully, so you would probably find more.

bferanmi806-sketch:
Thanks for the detailed comments and for taking the time to test the mutation yourself.

I’d be happy to take the second half of #347, so feel free to scope the four additional guards to me.

I can also give SysKnife a proper run on Linux outside the repo and report back anything I find.

vladimirrott:
Merged in `a62539e`, and I answered half of your comment and left the other half sitting. Correcting that now.

The four remaining guards are #355, and Soumoditya asked for them a few hours after you did, so that one is spoken for. You asked first and the ordering was mine rather than yours; sorry for the crossed wires. What you have instead is #229, which I offered you at almost the same moment and which is the more interesting of the two: a guard that derives a published figure from the tree rather than one that prints a better message.

Read the correction I posted on #229 before you start. Two of the three claims in that body were fixed back in August and I handed you the issue without rereading it, so the version you accepted was describing a file that had already changed. The body is rewritten and the remaining work is real.

On running SysKnife on Linux outside the repo: yes, please, and thank you for offering. A report from a machine you actually use is worth more to me than another green CI run, and no test in this tree can produce one. `sysknife doctor`, `sysknife history` and `sysknife audit verify` are the read-only surface and none of them change anything on the host, so it is a safe place to start. Tell me what is confusing or wrong and that is the contribution; there is no obligation attached and no follow-up if you would rather not.

PR #359, opened 2026-09-03
test(audit): pin CLI exit-code precedence

### LINKED ISSUE #357: run_audit_verify's exit code is not pinned: reverting it leaves the suite green
#340 fixed the anchor precedence bug and pinned two of the three places the new helper is used. The third is unpinned: revert it and the whole suite stays green.

`combined_verification_exit_code` exists because a detected break (exit 1) must outrank an inconclusive check (exit 2), which plain `.max()` gets backwards. It is used in three places:

| Call site | Pinned by |
|---|---|
| the helper itself | `a_detected_break_outranks_an_inconclusive_anchor_in_the_cli_exit_code` |
| `emit_verification`, the `--json` status | `json_status_includes_a_truncated_anchor_verdict` |
| **`run_audit_verify`, the process exit code** | **nothing** |

Measured: I reverted only the `run_audit_verify` line back to

```rust
let anchor_exit = anchor.as_ref()
    .map(sysknife_daemon::audit_chain::checkpoint_outcome_to_exit_code)
    .unwrap_or(0);
let exit_code = outcome.exit_code().max(anchor_exit);
```

and left everything else in place:

```
test result: ok. 264 passed; 0 failed; 0 ignored
```

## Why it matters

The exit code is the whole interface for anyone scripting `sysknife audit verify` in a cron job or a CI step, and it is the reason the command exists. A broken chain with an unreachable anchor should exit 1, meaning "this has been tampered with". Under `.max()` it exits 2, meaning "could not determine", and a script that treats 2 as a soft failure and 1 as an alert would route a real tamper event to the wrong branch.

The `#[ignore]` Postgres test does not cover it either. It uses a truncated anchor over an intact chain, where `.max(0, 1)` and the helper both give 1. The two forms only diverge on a broken chain with an inconclusive anchor.

## Scope

One test that drives the exit code through `run_audit_verify` rather than through the helper, with a chain that verifies as `Broken` and an anchor that returns `CannotVerify`, asserting the process exits 1.

If reaching `run_audit_verify` needs more scaffolding than the assertion is worth, an acceptable alternative is extending the existing `#[ignore]` CLI contract in `cli_smoke.rs` to add the broken-chain-plus-unreachable-anchor case, since that job already has a live database.

## Tests first

Write the test, then revert the `run_audit_verify` call site to the `.max()` form above and confirm the new test fails and names the exit code it got. A test that stays green with the fix removed is the exact defect #340 went three rounds over, so this one has to be proven the same way.

## Difficulty

`medium`, mostly for the fixture setup.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md). `docs/developer-guide.md` covers running the live Postgres contract locally if you take the second option. No CLA. MIT.

@k4its1t wrote the fix and the other two tests in #340, so this is theirs if they want it.

k4its1t:
Closes #357

## Summary

- drive the real `sysknife audit verify` binary with a local chain that is `Broken` under a mismatched public key
- configure an unusable checkpoint anchor so the anchor result is `CannotVerify`
- assert that the process exits `1`, pinning the precedence at the `run_audit_verify` call site
- record the CI-measured workspace baseline at 1,834 Rust tests

## Mutation proof

Temporarily restored the old `outcome.exit_code().max(anchor_exit)` implementation in commit `865d752`.

The new test failed exactly as required in [the mutation CI run](https://github.com/lacs-project/sysknife/actions/runs/33769949261/job/100697320969): expected exit code `1`, actual exit code `2`. Its captured output simultaneously reported `BROKEN` for the local chain and `ANCHOR NOT CHECKED` for the unusable checkpoint URL. Commit `4595173` reverted the mutation; there is no production-code diff in the final PR.

## Validation

Final Linux CI at `5c3f164`:

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-features --all-targets --locked -- -D warnings`
- `cargo doc --no-deps --workspace --locked` with warnings denied
- `scripts/test_baseline.sh`: 1,834/1,834 passed, 6 skipped
- live Postgres migration/store contract and CLI anchor contract
- docs and public-evidence checks
- frontend type check/tests and production dependency audit
- security audit, dependency review, TruffleHog, and E2E scripts lint

Local checks: `git diff --check`, JSON parse, and `python3 scripts/check_evidence_claims.py`. The full Rust validation used GitHub's Linux CI because this Mac does not have a Rust toolchain installed.

## Scope

Test and required evidence updates only. No production-code, dependency, or workflow changes in the final diff.

## AI assistance

Codex assisted with repository inspection, test implementation, and CI monitoring. I reviewed the complete diff and the reported validation evidence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,833 Rust tests and 72 frontend tests** form the current deterministic
+**1,834 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/apps/sysknife-cli/src/runner.rs
+++ b/apps/sysknife-cli/src/runner.rs
@@ -1015,11 +1015,7 @@ pub async fn run_audit_verify(args: AuditVerifyArgs, log: &Logger) -> Result<(),
     // this command used to report `audit_anchor: {configured: true}` without
     // ever reading the anchor — implying a check it did not perform.
     let anchor = verify_configured_anchor(&lacs_config, &db_path, &verifier).await;
-    let anchor_exit = anchor
-        .as_ref()
-        .map(sysknife_daemon::audit_chain::checkpoint_outcome_to_exit_code)
-        .unwrap_or(0);
-    let exit_code = outcome.exit_code().max(anchor_exit);
+    let exit_code = combined_verification_exit_code(&outcome, anchor.as_ref());
     emit_verification(&args, log, &outcome, &label_for_diag, anchor.as_ref());
     if exit_code == 0 {
         Ok(())
--- a/apps/sysknife-cli/src/runner.rs
+++ b/apps/sysknife-cli/src/runner.rs
@@ -1015,7 +1015,11 @@ pub async fn run_audit_verify(args: AuditVerifyArgs, log: &Logger) -> Result<(),
     // this command used to report `audit_anchor: {configured: true}` without
     // ever reading the anchor — implying a check it did not perform.
     let anchor = verify_configured_anchor(&lacs_config, &db_path, &verifier).await;
-    let exit_code = combined_verification_exit_code(&outcome, anchor.as_ref());
+    let anchor_exit = anchor
+        .as_ref()
+        .map(sysknife_daemon::audit_chain::checkpoint_outcome_to_exit_code)
+        .unwrap_or(0);
+    let exit_code = outcome.exit_code().max(anchor_exit);
     emit_verification(&args, log, &outcome, &label_for_diag, anchor.as_ref());
     if exit_code == 0 {
         Ok(())
--- a/apps/sysknife-cli/tests/cli_smoke.rs
+++ b/apps/sysknife-cli/tests/cli_smoke.rs
@@ -130,6 +130,45 @@ fn audit_verify_exits_with_code_2_when_the_key_file_is_missing() {
         .code(2);
 }
 
+#[test]
+fn audit_verify_exit_code_prefers_broken_chain_to_inconclusive_anchor() {
+    let dir = tempfile::tempdir().expect("create CLI fixture directory");
+    let db_path = dir.path().join("daemon.sqlite");
+    let key = AuditKey::load_or_generate(&dir.path().join("audit-key"))
+        .expect("generate audit key");
+    let store = TransactionStore::open_with_key(&db_path, Arc::new(key.clone()))
+        .expect("create audit store");
+    store
+        .record(NewTransaction {
+            request_id: "broken-chain-exit-code".to_string(),
+            request_hash: "broken-chain-exit-code-hash".to_string(),
+            action_name: "UpdateSystem".to_string(),
+            risk_level: RiskLevel::High,
+            summary: "Upgrade the system".to_string(),
+            warnings: vec![],
+            caller_role: CallerRole::Dev,
+            caller_principal: CallerPrincipal::Uid(1000),
+        })
+        .expect("record audit row");
+    drop(store);
+
+    let wrong_key_path = dir.path().join("wrong-audit-key");
+    let wrong_key = AuditKey::load_or_generate(&wrong_key_path).expect("generate mismatched key");
+    assert_ne!(key.verifying_key_hex(), wrong_key.verifying_key_hex());
+
+    cli()
+        .env("SYSKNIFE_DATABASE_PATH", &db_path)
+        .env("SYSKNIFE_CHECKPOINT_DB", "not-a-postgres-url")
+        .env("HOME", dir.path())
+        .env("XDG_CONFIG_HOME", dir.path())
+        .args(["audit", "verify", "--pubkey"])
+        .arg(dir.path().join("wrong-audit-key.pub"))
+        .assert()
+        // The wrong public key makes the local chain Broken (1), while the
+        // unusable configured anchor is CannotVerify (2). A break must win.
+        .code(1);
+}
+
 #[tokio::test]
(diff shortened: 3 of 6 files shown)


k4its1t:
Mutation proof is complete: with the old `.max()` call restored at `865d752`, the new real-binary test failed with expected `1`, actual `2` ([CI log](https://github.com/lacs-project/sysknife/actions/runs/33769949261/job/100697320969)). The mutation was reverted in `4595173`; the final tree has no production-code diff, and the full 1,834-test Linux CI is green.

vladimirrott (commented):
Reviewed at `5c3f164e`, with `main` at `f929272b` already in the branch.

You mutated the right line. I expected a revert of `combined_verification_exit_code` itself, which would have proved nothing: the unit test at `runner.rs:2898` already pins that function and would have gone red beside yours. `865d752` reverted the **call site** instead, back to `outcome.exit_code().max(anchor_exit)`, and left the function intact, so your test was the only thing between a detected break and a wrong exit code. It failed alone, and I read the job log rather than the summary:

```
FAIL ( 343/1834) sysknife-cli::cli_smoke audit_verify_exit_code_prefers_broken_chain_to_inconclusive_anchor
    Unexpected return code, failed var == 1
    └── var: 2
```

The captured stdout also closes the hole I went looking for. A test shaped like this passes for the wrong reason when the anchor comes back `NotConfigured` rather than `CannotVerify`, because then the chain alone gives exit 1 and the assertion holds under any precedence rule at all. Both verdicts are in the same output:

```
BROKEN: chain intact for first 0 row(s); row seq=1 ... does not chain.
  expected: valid ed25519 signature
ANCHOR NOT CHECKED: connecting to the checkpoint database failed: ... invalid checkpoint database URL
```

Chain contributes 1, anchor contributes 2, `.max()` returned 2. So the test measures the precedence and nothing else.

The fixture is hermetic, which matters for a test that reads real config. `HOME` and `XDG_CONFIG_HOME` cover `config_path()`, and that function consults no `/etc` path, so `LacsConfig::load()` returns defaults on any host. `SYSKNIFE_DATABASE_PATH` reaches `choose_audit_store` as `Explicit` and short-circuits both the per-user and the system probe. This passes on a developer box with a real install and a postgres `[storage]` block, not only on a clean runner.

I checked the evidence artifact against the run behind it instead of against the diff. At `a0354104` the suite printed `test_baseline: rust suite has 1834 tests, baseline says 1833` at `14:57:03`, which is the timestamp now sitting in `measured_at.tests`. At `5c3f164e` it printed `1834 tests run: 1834 passed, 6 skipped` and `matching the recorded baseline`. All four places that carry the figure moved together: the artifact, `README.md`, `docs/introduction.md` and `docs/distro-support.md`.

## Nothing blocking

This is ready as it stands. A test-only change needs no CHANGELOG entry, and you were right not to write one.

Two notes, neither of them yours:

- The module doc at `cli_smoke.rs:20` still claims every test sets `SYSKNIFE_SOCKET`. Nine of the eighteen do, and `audit verify` never opens the socket, so leaving it out is correct. That sentence drifted before this PR and I will fix it.
- I reviewed this by reading it. Fork branches are not built or tested on my machine, so the evidence above comes from the repo's own CI logs and from reading `main`. That applies to every fork PR here and says nothing about yours.

## Next, if you want it

#344 is yours if you say the word and I will hold it. You walked into it during this PR: no local Rust toolchain, so the only way to move `workspace-tests.json` was to push and read what CI measured. That is the whole issue, and you have more first-hand evidence for it than I do. #278 is its sibling and pins the artifact to the commit it names.

You took the follow-up I offered on #340 and came back with a mutation that isolates the one untested call site, rather than the easier revert that would have gone green in two places. Three rounds on the chain, and then this.

vladimirrott (approved):
This is better than the issue asked for, and it closes the gap I opened this morning.

I asked for a test that drives the exit code through `run_audit_verify`, and said the fallback was extending the `#[ignore]` Postgres contract because reaching that path looked like it needed a live database. You found the way that needs neither: verify with a mismatched public key so the local chain is `Broken`, point `SYSKNIFE_CHECKPOINT_DB` at an unusable URL so the anchor is `CannotVerify`, and assert `1`. That is the exact 1-versus-2 case, driven through the real binary, in a test anyone can run offline in under a second.

## Verified independently, at `5c3f164`

Reproduced your mutation in a container rather than reading the diff:

```
N0  unmutated                              1 passed
M1  run_audit_verify reverted to .max()    FAILED
```

```
panicked at apps/sysknife-cli/tests/cli_smoke.rs:169:10:
Unexpected return code, failed var == 1
code=2
```

With the fix gone the CLI reports "could not determine" where it should report "tampered with". That is the whole point of the issue, and it now fails loudly.

I also checked the two claims in your description rather than taking them:

- The CI run you linked is at head `865d752b`, which is your mutation commit, and its conclusion is `failure` with the `rust` job red. The mutation was real and it ran.
- `tests/evidence/workspace-tests.json` names commit `a035410`, and that commit is in this branch's history. That matters more than it sounds: #344 exists because an artifact once recorded a `commit` and `measured_at` for a run that never happened. Yours describes a measurement that did.

The count itself I did not have to take on trust either. `rust` passing means `test_baseline.sh` held 1,834 against the real suite, and `docs-and-hygiene` passing means the three published literals agree with the artifact.

One aside you might find useful: I could not run the full workspace in my container at all, because `apps/sysknife-shell` needs pkg-config, WebKit and GTK. That is #344 demonstrating itself while I was checking your work.

## Landing it

No production diff, tests and evidence only. `Closes #357`. Merging.

#338 is still yours whenever you want it, and I am not going to put anything else in front of you while it is open. Three PRs on the audit chain in two days, each one pinned by a mutation that actually goes red: that is the part most contributors skip.

vladimirrott:
Merged in `0640818`, and #340 landed before it in `5867807`. Two in two days.

The mutation proof on this one is the reason it went in without another round. You restored the old `.max()` at `865d752`, showed the new test failing with expected `1` and actual `2`, and linked the run. That is the bar this repo asks for and most PRs need to be asked twice before they meet it; you met it unprompted, on both.

`main` is at 1837 tests now, in case a rebase surprises you with a number.

Issue #250, opened 2026-08-18
peer_pidfd cannot tell a pre-6.5 kernel from a peer that already exited, so the PID-reuse check is off in the reuse case


vladimirrott:
`resolve_caller` reads the peer's supplementary groups from `/proc/{pid}/status` after `SO_PEERCRED`, so it pins the peer with a pidfd first and afterwards checks that the pinned process was not reaped during the read (`crates/sysknife-daemon/src/dispatcher.rs:541-566`). That check is skipped when `peer_pidfd` returns `None`, and the comment reads `None` as an old kernel (`dispatcher.rs:542`):

```rust
// Pin the connecting peer *before* the /proc read so PID reuse can be
// detected afterward. None on kernels without SO_PEERPIDFD (< 6.5).
```

`peer_pidfd` collapses every `getsockopt` failure into `None` (`dispatcher.rs:456`):

```rust
if rc != 0 || fd < 0 {
    return None;
}
```

A peer that has already been reaped is one of those failures. Measured on Linux 6.8.0-137, with a child connecting over `AF_UNIX` and the parent calling `getsockopt` before and after reaping it:

```
live peer:   rc=0  fd=5
reaped peer: rc=-1 errno=22 (EINVAL)
peercred:    rc=0  pid=792075 uid=1000 gid=1000
bogus opt:   rc=-1 errno=92 (ENOPROTOOPT)
```

The kernel reaches that through `sk_getsockopt` → `pidfd_prepare`, which rejects a `struct pid` whose task is gone. An unsupported option gives `ENOPROTOOPT` instead, so the two cases are distinguishable at the syscall boundary and `peer_pidfd` throws the distinction away.

The ordering that matters:

1. The peer connects, sends a request, exits.
2. `peer_pidfd` fails with `EINVAL` and returns `None`. The reuse check is off, on every kernel version.
3. `groups_for_pid(pid, &gid_map)` reads `/proc/{pid}/status`. When the pid has been recycled those are another process's supplementary groups, and `highest_role_from_groups` derives the caller's role from them.

`SO_PEERCRED` keeps reporting the dead peer's pid (792075 above), so step 3 has a pid to read and no signal that it is stale.

Reaching a higher role this way needs pid wraparound to land on a process in `sysknife-admin` or `wheel` inside the window between two syscalls, so the odds are poor. The defect is that the mitigation is absent in the one case it was written for, while the code says it is present on 6.5 and later.

## Suggested fix

[...]

Issue #237, opened 2026-08-18
Split the Ubuntu-only actions out of DEBIAN_ONLY_ACTIONS, so the family fence means what it says


vladimirrott:
`DEBIAN_ONLY_ACTIONS` (`crates/sysknife-core/src/action_family.rs:112`) holds 67 names, and the reference table renders all 67 with `Distro = Ubuntu`, so the two agree exactly today. The problem is what the list means. Its sibling states the contract (`action_family.rs:22`):

> Membership means **impossibility**, not preference: the action's mechanism cannot work on the other family, so the daemon fence and the CLI routing guard refuse it there. For "runnable but not this family's canonical tool", see `NON_CANONICAL_ON_DEBIAN`.

By that contract most of the 67 are correctly placed and about a third are not. They are lumped together because "Debian-family" and "Ubuntu" have never had to be different things. Making Debian eligible (#238) requires separating them first, so this is the prerequisite.

## Inventory of the 67, by what happens on Debian

Grouped by the tool each `ActionSpec` drives, read from `docs/action-reference.md`. Availability checked against Debian's own source database (`sources.debian.org/api/src/<pkg>/`), suite by suite, rather than assumed.

**Impossible on Debian, no path at all (7).** Canonical services with no Debian equivalent:

| actions | tool |
|---|---|
| `ProStatus`, `ProAttach`, `ProDetach`, `EnableProService`, `DisableProService` | `pro` |
| `LivepatchStatus` | `canonical-livepatch` |
| `UbuntuReleaseUpgrade` | `do-release-upgrade` |

**Installable on Debian, but not its canonical tooling (14).** These belong in a `NON_CANONICAL_ON_DEBIAN`-shaped list rather than an impossibility list:

| actions | tool | availability |
|---|---|---|
| `SnapInstall`, `SnapRemove`, `SnapRefresh`, `SnapHold`, `SnapUnhold`, `SnapList`, `SnapInfo`, `SnapRevert`, `SnapClassicInstall` | `snap` | `snapd` in trixie and sid; not installed by default |
| `NetplanApply`, `NetplanSet`, `NetplanGenerate`, `NetplanGetConfig` | `netplan` | `netplan.io` in bookworm and trixie. Debian defaults to ifupdown or systemd-networkd, and upstream ships a `netplan-everywhere.md` guide for non-Ubuntu hosts |
| `MultipassList` | `multipass` | distributed via snap on Linux, so it inherits the snap answer |

[...]

PR #360, opened 2026-09-03
fix(daemon): gate mutating query actions

### LINKED ISSUE #226: query_action runs AptUpdate, a root apt mutation, with no preview and no approval receipt

QinXi-ai:
## Summary

- move the explicit `OBSERVER_MUTATING_ACTIONS` classification next to the daemon action catalogue so every approval-free surface shares it
- reject Low/Observer actions that mutate the host at the raw `query_action` IPC boundary, before platform validation, spec construction, or execution
- extend the existing approval-gate IPC regression to cover both a conventional higher-risk action and `AptUpdate`
- document the daemon-side enforcement and changelog the fix

## Security boundary

`RiskLevel::Low` determines the baseline caller role; it does not prove an action is read-only. `AptUpdate` runs `sudo apt-get update`, writes the root apt index, and takes the dpkg lock. A direct socket client could previously invoke it through `query_action` without preview, an approval receipt, a transaction row, or the exclusion gate.

The daemon now fails closed with `authorization_failure` using the same explicit classification that already prevents the CLI from advertising `AptUpdate` as a direct MCP tool. Ordinary read-only queries and operator risk raises keep their existing behavior.

## Verification

PASS locally on Windows:

- `cargo metadata --locked --no-deps --format-version 1`
- `python -X utf8 scripts/check_evidence_claims.py`
- `git diff --check`

NOT TESTED locally:

- Rust execution tests and full workspace CI. The Windows build stops in the unchanged Linux-only `vsock` dependency because Unix/VSOCK APIs are unavailable. The targeted regression is included for the repository's Linux CI.

Closes #226.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,14 @@ Releases before `0.2.5` predate the public launch; their notes live in the
 
 ### Fixed
 
+- **`query_action` can no longer bypass approval for Low-risk mutations.**
+  The daemon now rejects `AptUpdate` on the raw approval-free query path using
+  the same explicit Observer-mutating classification as the generated MCP
+  router. Direct socket clients must use preview, approval receipts and execute,
+  so the root apt index cannot be changed without an audit-backed approval or
+  concurrently with an action holding the dpkg exclusion gate.
+  ([#226](https://github.com/lacs-project/sysknife/issues/226))
+
 - **`sysknife audit verify` no longer lets an inconclusive anchor hide a broken
   chain.** The command combined the local audit result with the configured
   external anchor using `.max()` on their exit codes. Exit 1 means a break was
--- a/apps/sysknife-cli/src/mcp_server.rs
+++ b/apps/sysknife-cli/src/mcp_server.rs
@@ -63,6 +63,7 @@ use sysknife_brain::planning_tools::propose_plan::KNOWN_ACTIONS;
 use sysknife_brain::state_client::StateClient as _;
 use sysknife_core::action_family::{DEBIAN_ONLY_ACTIONS, FEDORA_ONLY_ACTIONS};
 use sysknife_core::distro::DistroId;
+use sysknife_daemon::actions::OBSERVER_MUTATING_ACTIONS;
 
 use crate::client::{DaemonClient, DescribeInfo};
 use crate::error::CliError;
@@ -475,9 +476,6 @@ const MCP_READ_ONLY_ACTIONS: &[&str] = &[
     // Dispatcher-internal, with no ActionSpec of its own.
     "ListJobHistory",
 ];
-
-/// Low-risk actions that still mutate the host and therefore remain plan-only.
-const OBSERVER_MUTATING_ACTIONS: &[&str] = &["AptUpdate"];
 
 impl SysknifeMcpServer {
     fn new() -> Self {
--- a/crates/sysknife-daemon/src/actions/mod.rs
+++ b/crates/sysknife-daemon/src/actions/mod.rs
@@ -42,6 +42,14 @@ pub mod ufw;
 pub mod users;
 pub(crate) mod validate;
 
+/// Observer-callable actions that still mutate the host and therefore require
+/// the preview/approve/execute interlock.
+///
+/// This is shared by every approval-free surface. Keeping the exception beside
+/// the action catalogue prevents the CLI's generated MCP router and the
+/// daemon's `query_action` endpoint from making independent safety decisions.
+pub const OBSERVER_MUTATING_ACTIONS: &[&str] = &["AptUpdate"];
+
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ActionMechanism {
     Command {
--- a/crates/sysknife-daemon/src/dispatcher.rs
+++ b/crates/sysknife-daemon/src/dispatcher.rs
@@ -1833,6 +1833,25 @@ async fn handle_query_action(
         .await;
     }
 
+    // Risk answers how much authorization an action needs; it does not prove
+    // that an action is read-only. `AptUpdate`, for example, is Low/Observer but
+    // runs `sudo apt-get update`, writes the root apt index and takes the dpkg
+    // lock. The shared catalogue exception keeps this raw IPC path aligned with
+    // the generated MCP surface instead of letting direct clients bypass the
+    // preview, receipt and exclusion gates.
+    if crate::actions::OBSERVER_MUTATING_ACTIONS.contains(&action_name) {
+        return send_error(
+            framed,
+            request_id,
+            "authorization_failure",
+            format!(
+                "{action_name} mutates the host and is not available through query_action; \
+                 use preview+execute instead"
+            ),
+        )
+        .await;
+    }
+
     // Same fence preview and execute apply. Without it a Fedora-only read-only
     // action reached the executor on a Debian host and failed as "rpm-ostree:
     // No such file or directory" — an execution error where the other paths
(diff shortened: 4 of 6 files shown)


vladimirrott (approved):
This closes #226 properly, and the part I want to name is the one you were not asked for: you removed the duplicate rather than adding a second guard.

`OBSERVER_MUTATING_ACTIONS` now lives beside the action catalogue, the CLI imports it, and `git grep` finds exactly one definition. The two approval-free surfaces can no longer make independent safety decisions, which is the actual bug class. A copy in `dispatcher.rs` would have passed review today and drifted within a release.

The comment earns its place too. `RiskLevel::Low` decides how much authorization an action needs; it never claimed the action was read-only, and `AptUpdate` writes the root apt index and takes the dpkg lock. Saying that at the guard means the next person does not re-derive it.

## Verified, and my first attempt was wrong

Reproduced at `1fa9a521` in a container, neutering only the daemon call site (the `crate::actions::` prefix is unique to `dispatcher.rs`, so the CLI's copy stays intact):

```
1 test matched and passed unmutated
mutated: test query_action_rejects_actions_that_require_approval ... FAILED  (rc=101)
```

My first run reported the opposite, that the guard did not bite. That was my tooling, not your patch: it ran `cargo test --bins`, and `coverage_gaps.rs` is an integration target, so it executed **zero tests** and called the vacuous pass a result. Fixed on my side, and it now refuses any filter that matches nothing. Worth saying out loud because it is the same defect this repository keeps finding in its own guards, and I shipped it in the thing that checks for it.

Board is green, 11 passing. `CHANGELOG.md` and `docs/mcp.md` are both in the diff, so there is nothing for me to write at merge.

## One thing for later, not for this PR

The list has one entry. When it gains a second, the interesting question is how anyone would know: nothing today derives "mutates the host" from the action definition, so the classification stays correct only while a human remembers to add to it. `observer_actions_are_fully_classified` catches a new **Observer** action nobody classified, which is most of the risk, but an existing action that becomes mutating would slip past. Not a gap in this change, and not worth blocking on.

Merging. That is #216 and #226 in two days, both on the boundary between the approval-free surface and the daemon, which is the part of this project where being wrong is expensive.

vladimirrott:
@QinXi-ai I want to say this properly rather than in a merge line.

You have landed two pull requests in two days, both on the part of this project where being wrong is most expensive, and both times you did something you were not asked to do.

On #348 you took #216 off a Call for Participation listing with no prior context on the codebase, and came back with an allowlist **and a drift test**. Most people write the allowlist and stop. That test is what makes the classification survive the two of us forgetting why it exists. When I pointed out the stdio test only bounded the surface loosely, you did not argue the scope; you added the assertion that every advertised tool must appear in the snake_case image of `MCP_READ_ONLY_ACTIONS`, across all three distro variants, copying none of the filter's logic.

On #360 you did the thing I have been trying to get right in my own tooling all week: you **removed the duplicate instead of adding a second guard**. Moving `OBSERVER_MUTATING_ACTIONS` beside the action catalogue means the two approval-free surfaces can no longer make independent safety decisions. A copy in `dispatcher.rs` would have passed review and drifted inside a release, and nobody would have noticed until it mattered. You also wrote the comment explaining that `RiskLevel::Low` decides authorization rather than proving read-only, which saves the next reader from deriving it.

And you did all of it from Windows, where you cannot compile the Unix-only `vsock` dependency or run the full workspace. You said so plainly every time, listed what you had tested and what you had not, and never once handed me a number you had not measured. That honesty is worth more to me than the patches. It is also rarer: the thing that wastes maintainers' time is not bad code, it is confident text with nothing behind it.

For what it is worth, my own gate refused #360 on the first attempt and it was my bug, not yours: it ran `cargo test --bins`, missed the integration target, executed zero tests and called that a result. You cost me nothing; I cost you a round trip. Fixed, and it now refuses any filter that matches nothing.

Both of your changes ship in **v0.13.0**, which is going out now. The release notes lead with yours.

## Four issues, offered as a menu and not a queue

I normally hold to one open offer per person, because a stack of them reads as an assignment and that is how volunteers get tired. I am breaking my own rule once, and saying so, because you have earned the right to choose rather than be handed. **Take any, all, or none, in any order, and ignore the rest without a word.**

- **#237** — `DEBIAN_ONLY_ACTIONS` conflates "Debian-family" with "Ubuntu-only", so the family fence is doing two jobs. This is #360 again in a different constant: one list, two meanings, several readers.
- **#250** — `peer_pidfd` cannot distinguish a pre-6.5 kernel from a peer that already exited, so the daemon cannot tell "cannot attribute" from "nobody there". You have twice now got the fail-closed direction right on this boundary.
- **#344** — a contributor who cannot build the workspace has no supported way to update the test-count artifact. **You have lived this one.** You hit it on both PRs and worked around it honestly. Nobody is better placed to say what the supported path should be.
- **#362** — `postgres-contract` gained a second live step in #340 and its guard predates it, so three one-line changes leave a required check green while it runs nothing. A guard that does not bite, which is the shape you keep fixing.

If none of them appeal, say so and I will find something closer. If you would rather I stopped sending things for a while, say that instead and I will.

Issue #342, opened 2026-09-02
No .gitattributes, so a Windows or WSL clone gets CRLF and a byte-exact doc test blames the action catalogue


vladimirrott:
> **Re-measured 2026-09-07 at `adab560`, and half of this issue is done.** The
> committed CRLF artifact was renormalised by #361 on 3 September, before you
> were pointed here. Nothing tracked is CRLF or mixed today:
>
> ```
> $ git ls-files --eol | grep -E '^i/(crlf|mixed)'; echo "grep rc=$?"
> grep rc=1
> $ grep -c $'\r' tests/evidence/workspace-tests.json
> 0
> ```
>
> What is left is the prevention half, which nobody has touched: the missing
> `.gitattributes`, the guard, and the `CONTRIBUTING.md` paragraph. **The trap
> named under "Tests first" has inverted.** The guard now passes on its first
> run, so a contributor who adds it and sees green has learned nothing about
> whether it works. Prove it bites before you believe it.
>
> Three line references below have moved and one figure is gone:
>
> ```
> $ grep -n 'json.dumps' scripts/record_test_baseline.py
> 160:        json.dumps(doc, indent=2, sort_keys=True) + "\n", encoding="utf-8"
> $ sed -n "154p;156p" crates/sysknife-daemon/tests/action_reference_doc.rs
>     }
>     let committed = std::fs::read_to_string(&path).unwrap_or_default();
> $ wc -c tests/evidence/workspace-tests.json
> 193 tests/evidence/workspace-tests.json
> ```
>
> So `record_test_baseline.py:65` is now 159-160 (#393 moved it),
> `action_reference_doc.rs:154` is 162 (156 on 2026-09-07; #384 moved it again on 2026-09-09), and the "450 bytes of LF over 467 bytes
> of CRLF" sentence describes a file that #393 replaced with a 193-byte schema.
> The reasoning around each of them still holds; only the numbers moved.

The repository has no `.gitattributes`, and `core.autocrlf` is whatever the contributor's machine set. A Windows or WSL clone with `autocrlf=true` therefore gets CRLF in the working tree, and a commit back from that clone puts CRLF into the index. Nothing in the repo prevents either half, and nothing tells the contributor it happened.

**Both halves have now happened.** At `60af0ad`:

```
$ git ls-files --eol | grep -E '^i/(crlf|mixed)'
i/crlf  w/crlf  attr/    tests/evidence/workspace-tests.json
```

[...]

PR #365, opened 2026-09-04
fix(tests): mask empty grep results in remaining guards

### LINKED ISSUE #355: Four more hygiene guards exit 1 with no output when their anchor moves
#347 named five guards and #349 fixed all five. There are four more with the same shape, and I missed them when I filed it.

Same test each time: move the anchor the guard reads, run the guard, count the bytes it prints.

| Guard | Broken input | rc | bytes |
|---|---|---|---|
| `scripts/check_release_versions.sh:62` | internal pin prefix renamed in the manifests | 1 | **0** |
| `tests/e2e/story-metadata.test.sh:31` | `--metadata` emits nothing | 1 | **0** |
| `tests/e2e/story-metadata.test.sh:66` | same | 1 | **0** |
| `tests/e2e/story-metadata.test.sh:67` | same | 1 | **0** |

## Why it matters

`check_release_versions.sh` is the worse of the two. It runs in `docs-and-hygiene`, in `scripts/release_rehearsal.sh --check`, and in the release workflow's own preflight, so it is the last thing standing between a missed internal dependency pin and a published crate that depends on the previous release. Its emptiness check is written and reachable-looking:

```bash
if [[ -z "$pins" ]]; then
    printf 'No internal dependency pins found; the manifest paths in %s are stale.\n' \
        "${BASH_SOURCE[0]}" >&2
    exit 1
fi
```

Measured with the pin prefix renamed so the grep matches nothing: `rc=1`, zero bytes. That message has never printed and cannot.

`story-metadata.test.sh` is the more embarrassing one, and I got it wrong in #347. That issue's "Tests first" section says this file "already fails loudly" because its header parser reports first. It does not. The three `grep -c .` assignments run before the parse loop, and `grep -c .` exits 1 on empty input, so the script dies at line 31. The comment sitting directly above that line reads:

```
# A regex that quietly matched nothing would make every assertion below vacuous.
```

The guard written to catch a parser that dropped everything is unreachable in exactly the case where the parser dropped everything. Measured by stubbing `run-stories.sh --metadata` to emit nothing: `rc=1`, zero bytes, and neither `derived $derived_count stories from $file_count files` nor `family split is degenerate` ever prints.

## Scope

Four lines, the same one-token change #349 already made five times: `|| true` on the pipeline so the emptiness check the authors wrote can run.

Out of scope: the "guard over the guards" idea from #347, which is still worth doing and is `medium`, not this.

## Tests first

Break each anchor in a scratch tree, run the guard, and assert it prints its own message rather than nothing:

- `check_release_versions.sh`: rewrite the `^sysknife-[a-z-]* = {` prefix in `crates/*/Cargo.toml` and both `apps/*/Cargo.toml`, then require `No internal dependency pins found` on stderr.
- `story-metadata.test.sh`: replace `tests/e2e/run-stories.sh` with a stub that exits 0 printing nothing, then require `the parser is dropping some` or the degenerate-family line.

Both must still pass untouched on clean input. Check the byte count of the output, not only the exit code: all four already exit 1, and the exit code is what makes this invisible.

## Difficulty

`easy`. Four one-token edits and a `sed` per guard to verify.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands, and [docs/developer-guide.md](https://github.com/lacs-project/sysknife/blob/main/docs/developer-guide.md) now has a section on running these guards directly; they need no build and finish in under a second each. No CLA. MIT.

@iamjaeholee found and fixed the first five in #349, so this is theirs if they want it.

iamjaeholee:
Fixes #355

    Added `|| true` to the remaining 4 places.

### DIFF
--- a/scripts/check_release_versions.sh
+++ b/scripts/check_release_versions.sh
@@ -63,7 +63,7 @@ pins="$(grep -rn '^sysknife-[a-z-]* = {' \
     "$repo_root"/crates/*/Cargo.toml \
     "$repo_root"/apps/sysknife-cli/Cargo.toml \
     "$repo_root"/apps/sysknife-shell/src-tauri/Cargo.toml |
-    grep 'version = ')"
+    grep 'version = ' || true)"
 
 # An empty result means the manifests moved, not that every pin agrees. Fail
 # loudly rather than reporting success for a check that inspected nothing.
--- a/tests/e2e/story-metadata.test.sh
+++ b/tests/e2e/story-metadata.test.sh
@@ -28,7 +28,7 @@ report() {
 [ -x "$runner" ] || { printf 'missing runner: %s\n' "$runner" >&2; exit 1; }
 
 metadata="$(bash "$runner" --metadata)"
-derived_count="$(printf '%s\n' "$metadata" | grep -c .)"
+derived_count="$(printf '%s\n' "$metadata" | grep -c . || true)"
 file_count="$(find "$story_dir" -maxdepth 1 -name 'story-*.sh' | wc -l)"
 
 # A regex that quietly matched nothing would make every assertion below vacuous.
@@ -63,8 +63,8 @@ done <<< "$metadata"
 
 # Both families must be populated: a filter that silently classified everything
 # one way would still satisfy the count check above.
-ubuntu_count="$(printf '%s\n' "$metadata" | awk -F'\t' '$2 == "ubuntu"' | grep -c .)"
-atomic_count="$(printf '%s\n' "$metadata" | awk -F'\t' '$2 == "atomic"' | grep -c .)"
+ubuntu_count="$(printf '%s\n' "$metadata" | awk -F'\t' '$2 == "ubuntu"' | grep -c . || true)"
+atomic_count="$(printf '%s\n' "$metadata" | awk -F'\t' '$2 == "atomic"' | grep -c . || true)"
 if [ "$ubuntu_count" -lt 1 ] || [ "$atomic_count" -lt 1 ]; then
     report "family split is degenerate: $ubuntu_count ubuntu, $atomic_count atomic"
 fi


vladimirrott (approved):
Verified at `4bc8e76f2874ccedd0eef34020712b21e279fd18`.

You fixed the family counts on lines 66 and 67 rather than stopping at line 31, where the first symptom shows. That distinction carries the whole PR. Fixing only line 31 makes the file look repaired: it prints one failure, then dies at 66 before the rest of the file runs.

`story-metadata.test.sh` is a shell script, so the screen here classes it as executable at test time. I read the diff instead of running it on this machine, then ran the two guards in a rootless container with the network off. Your branch is 12 commits behind `main`, so I built the tree that would land rather than the one you pushed: `main` at `b7c472c`, with your two files copied over it. `git diff` between your base and `main` for those two paths is empty, so that tree is the merge result.

**Guard 1, `scripts/check_release_versions.sh`**, with the `^sysknife-... = {` pin prefix renamed in every manifest so the grep matches nothing:

```
A  fix present   + manifests intact  :: rc=0 bytes=72
      | All release versions match 0.13.0 (15 internal dependency pins checked).
B  fix present   + pin prefix broken :: rc=1 bytes=110
      | No internal dependency pins found; the manifest paths in .../check_release_versions.sh are stale.
C  FIX REVERTED  + pin prefix broken :: rc=1 bytes=0
D  FIX REVERTED  + manifests intact  :: rc=0 bytes=72
      | All release versions match 0.13.0 (15 internal dependency pins checked).
```

**Guard 2, `tests/e2e/story-metadata.test.sh`**, with `run-stories.sh` replaced by a stub that exits 0 printing nothing:

```
A  fix present   + real runner       :: rc=0 bytes=67
      | Story metadata derived cleanly: 133 stories (79 ubuntu, 54 atomic).
B  fix present   + runner emits none :: rc=1 bytes=215
      | FAIL  derived 0 stories from 133 files — the parser is dropping some
      | FAIL  family split is degenerate: 0 ubuntu, 0 atomic
      | FAIL  an unparseable story header did not fail the derivation
C  FIX REVERTED  + runner emits none :: rc=1 bytes=0
D  FIX REVERTED  + real runner       :: rc=0 bytes=67
      | Story metadata derived cleanly: 133 stories (79 ubuntu, 54 atomic).
```

Row C is the one that had to be run. With your hunk reverted, both guards still exit 1 and print zero bytes, which is why nobody caught this from an exit code. Row D says clean input behaves the same either way.

I also checked that 66 and 67 are not redundant with 31:

```
E  only lines 66-67 reverted + runner emits none :: rc=1 bytes=70
      | FAIL  derived 0 stories from 133 files — the parser is dropping some
```

One failure instead of three, and the run dies before the header-mutation check at line 102 executes.

Masking an exit code can hide a real error as easily as an empty match, so I ran two more cases against `check_release_versions.sh`:

- One manifest deleted: `rc=2`, `sed: can't read .../crates/sysknife-core/Cargo.toml: No such file or directory`. The package-version loop reports it before line 62, so `|| true` never gets a chance to swallow it.
- The no-argument form that `docs-and-hygiene` runs: `rc=0`, 72 bytes, passes.

And the sweep at `main` for the same shape returns your four lines and nothing else:

```
$ grep -rnE '="\$\(.*grep ' scripts/ tests/ --include='*.sh' | grep -v '|| true'
scripts/check_release_versions.sh:62:pins="$(grep -rn '^sysknife-[a-z-]* = {' \
tests/e2e/story-metadata.test.sh:31:derived_count="$(printf '%s\n' "$metadata" | grep -c .)"
tests/e2e/story-metadata.test.sh:66:ubuntu_count="$(printf '%s\n' "$metadata" | awk -F'\t' '$2 == "ubuntu"' | grep -c .)"
tests/e2e/story-metadata.test.sh:67:atomic_count="$(printf '%s\n' "$metadata" | awk -F'\t' '$2 == "atomic"' | grep -c .)"
```

Between #349 and this, the shape is gone from `scripts/` and `tests/`.

**Nothing blocking.** One mechanical item, and it is not your debt: the branch is 12 commits behind `main`, so `mergeStateStatus` reads `BEHIND` and the merge needs a branch update and a fresh run first.

I am leaving the merge to Vladimir rather than doing it myself. Our merge gate refuses to land anything without an observed mutation receipt, and it only knows how to produce one for Rust: it runs `cargo test` and applies its mutation to `*.rs`. A shell-only change has no path through it. The gap is ours, not yours, and I would rather say so than route around the gate.

No new issue from me, because #342 is already yours and still held. Take it whenever you want it, or leave it; it stays open for you either way.

One last thing, unrelated to merging. Anyone who stubs a runner and counts the bytes it prints administers real machines. If you have an Ubuntu box you look after, the read-only side of this is a ten-minute try: `sysknife --dry-run "list running containers"` shows you the plan and the exact commands without executing anything, and `sysknife doctor` tells you whether the daemon is even reachable. Say no and nothing changes here.

vladimirrott:
Merged in `99e789f`, and #355 closed with it.

Four hunks, and the pair at `story-metadata.test.sh:66-67` is the one I would
point at. The `derived_count` check above them already catches a parser that
drops stories; those two catch a parser that files every story into one family,
which is the case that matters because `SYSKNIFE_ALLOW_DESTRUCTIVE=1` runs the
atomic set by default.

I proved both guards in a container before merging, in both directions.
`check_release_versions.sh` with its manifest anchor broken: with your fix, 110
bytes naming the file and the cause, exit 1. With the fix reverted, the same
break gives 0 bytes, exit 1. Identical detection, and the whole difference is
whether anyone can act on it.

It sat a day for reasons that were not yours. The branch was 12 commits behind,
and then registry.npmjs.org started answering `npm audit` with 503, which lives
inside the required `frontend` check, so three consecutive runs failed on a job
that has nothing to do with your diff.

No CHANGELOG entry: test and release-script hygiene, nothing user-visible.

That outage is now #367 if you want it. `npm audit` blocks every pull request
whenever the registry has a bad morning, and what it audits is the paused Tauri
GUI, so a one-line `CONTRIBUTING.md` change waits on it too. Same shape as the
two you have fixed, one level up: a gate reporting a failure it did not find.
#342 is also still open and unclaimed.

PR #366, opened 2026-09-04
docs(contributing): document test-baseline fallback

### LINKED ISSUE #344: A contributor who cannot build the workspace has no supported way to update the test-count artifact

k4its1t:
## Summary

- document the supported fallback for contributors who cannot build the full workspace: leave the evidence artifact and all three published prose counts untouched, and state the blocked command plus missing dependency in the PR
- make the maintainer-owned pre-merge regeneration step explicit so contributors do not guess a count or manufacture `commit` / `measured_at` provenance
- document that `cargo-nextest` itself can be installed with `cargo install cargo-nextest --locked`, independently of the GUI system libraries

I chose the smaller solution explicitly allowed by #344. The full two-workflow design needs a credential that can write to the contributor's fork: GitHub documents that the built-in `GITHUB_TOKEN` can only access the workflow repository, so `maintainer_can_modify` alone does not let an upstream Actions token write across repositories. This repository has no declared GitHub App or PAT for that role, and adding an undeclared long-lived credential would not be a complete or safer fix.

GitHub reference: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github#authenticating-to-the-api-in-a-github-actions-workflow

Closes #344

## Validation

- [x] `npx --offline --yes --package=markdownlint-cli2@0.23.2 -- markdownlint-cli2 CONTRIBUTING.md`
- [x] `python3 scripts/check_evidence_claims.py`
- [x] `bash scripts/check_no_secrets.sh CONTRIBUTING.md`
- [x] `git diff --check`
- [x] Documentation updated
- [x] Security impact considered; no writable workflow or credential added
- [ ] `tests/release/public-claims.test.sh` — macOS system Bash 3 lacks the script's required `mapfile`; the underlying evidence-claim checker passed and Ubuntu CI is authoritative

## Notes for Reviewers

This is a documentation-only fallback, not partial automation. It deliberately adds no workflow, helper, token, dependency, or generated evidence change.

AI assistance was used to inspect the issue and repository conventions, draft the documentation, and run the listed checks. I reviewed the final one-file diff and all claims above.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,16 +115,31 @@ UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh
 grep -rn 'Rust tests' README.md docs/introduction.md docs/distro-support.md
 ```
 
+`cargo-nextest` itself needs no system packages. If it is the missing piece,
+install it directly:
+
+```sh
+cargo install cargo-nextest --locked
+```
+
+The GUI workspace members still need the platform libraries named in CI. If
+you cannot install those, do not estimate a count or edit the evidence metadata
+by hand. Leave `tests/evidence/workspace-tests.json` and the three prose files
+untouched. In the PR body, name the command you could not run and the missing
+tool or system library that blocked it. A maintainer will run the full suite and
+regenerate all four files before merge. This is the supported fallback; an
+honest missing measurement is better than metadata that describes a run which
+never happened.
+
 CI gates both halves, so bumping the artifact alone turns `docs-and-hygiene` red
 after `rust` goes green.
 
 **Do not chase the figure.** It moved fifteen times in the twenty days to
 2026-08-24, so on any PR that waits a day or two the number you recorded stops
 being the answer, and that is not your problem to solve. Record it once, say in
-the PR how many tests you added, and if it has expired by the time the PR is
-ready the maintainer regenerates it at merge. Whoever merges is what moves the
-figure, so whoever merges carries it. If `rust` is red only on the baseline, the
-PR is not blocked on you.
+the PR how many tests you added, and let the maintainer refresh it after later
+merges. Whoever merges is what moves the figure, so whoever merges carries it.
+If `rust` is red only on the baseline, the PR is not blocked on you.
 
 `scripts/test_baseline.sh` names both numbers when they disagree, in the form
 `rust suite has <measured> tests, baseline says <recorded>`, so a red run tells


k4its1t:
All code, docs, security, and CodeQL checks passed in both CI attempts. The only red job is `frontend`; `npm ci` succeeded both times, then the npm audit service failed externally:

- first run: network timeout at `https://registry.npmjs.org/-/npm/v1/security/advisories/bulk`
- second run: `503 Service Unavailable` from the same endpoint

I attempted the upstream rerun API first, but GitHub does not grant fork contributors permission to rerun upstream jobs. I then used one empty, tree-preserving commit to obtain an independent retry; it reproduced the service failure. No PR change is indicated, and I will not consume another full CI run for the same external outage. Please rerun only the failed job when the npm audit endpoint recovers.

vladimirrott (approved):
Approved, and I checked the three claims in the body rather than taking them.

`cargo install cargo-nextest --locked` is what `scripts/ci-local.sh:158` already
prints when the tool is missing, so after this the two places that say it agree.

The token argument holds. `gh secret list` on this repository returns one secret,
`CARGO_REGISTRY_TOKEN`, which publishes to crates.io and cannot write a git ref
anywhere. There is no App installation and no PAT for the cross-repository write
the two-workflow design needs. The smaller fix is the right one here, and I would
have pushed you toward it even if you had offered the larger one.

The paragraph I care most about is the one telling a blocked contributor to name
the command and the missing library instead of guessing a number. #344 existed
because the alternative was someone editing `measured_at` by hand, and a
provenance field that describes a run nobody made is worse than an absent one.

One optional thing, not for this PR. nextest ships pre-built binaries, and on
Linux that path has no compiler step at all:

```sh
curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
```

`cargo install --locked` builds from source, so it wants a working linker and a
few minutes, and CI itself takes the pre-built route through
`taiki-e/install-action`. Worth a follow-up that fixes both this file and
`ci-local.sh:158` so the documented path is the one CI uses. It is yours if you
want it; open the issue and I will label it.

vladimirrott:
Merged in `7d19864`, and #344 closed with it.

The sentence I would keep if I had to cut the rest: an honest missing
measurement is better than metadata that describes a run which never happened.
That is the whole of #344, and it now sits in the file a contributor reads
before they touch the artifact.

Two mechanical notes for whoever reads this thread later. The branch went
`BEHIND` when #365 landed half an hour before yours, so it needed an update and
a fresh board: the green checks it had were describing a different tree. And the
registry outage that failed your first two runs is #367 now.

No CHANGELOG entry, nothing user-visible.

The nextest follow-up in my review is still yours if you want it.

Issue #331, opened 2026-09-01
Nothing requires that a tests/release/*.test.sh is actually invoked by a gate


vladimirrott:
`tests/release/database-path-agreement.test.sh` was in the tree, correct, and wired into nothing. Zero references in `ci.yml`, `e2e.yml`, `release.yml` or `scripts/ci-local.sh`, and it never had any since the day it was added.

It guards a real past defect: the installer and the daemon resolving different SQLite paths, which splits the audit chain in two and leaves `audit verify` reading a store nothing writes to. Reverting the installer's path resolution to the old value left every required check green.

#330 wires that one file in. The gap it exposes is not that one file.

## The gap

`scripts/check_repo_completeness.sh` checks that required files exist. Nothing checks that a release test is reachable from a gate, so adding `tests/release/foo.test.sh` and forgetting the two lines in `ci.yml` and `ci-local.sh` produces a test that runs only when someone types its name. There is no error, no warning, and the file looks maintained.

Same shape as #278: an artifact that is correct and unread.

## Scope

A check that, for every `tests/release/*.test.sh`, the filename appears in at least one of `.github/workflows/ci.yml`, `.github/workflows/e2e.yml`, `.github/workflows/release.yml` or `scripts/ci-local.sh`.

Two details that decide whether it is worth having:

- It must fail when the glob comes back empty. A broken glob passes over zero files and reports success, which is the same defect the check exists to catch, one level up. That shape has shipped in this repository before: see #347 and #355, where nine guards exited 1 with no output because a `grep` that matched nothing killed the script before its own diagnostic could run.
- Path-filtered jobs count as wired. `e2e.yml` only runs on some paths, and a test reachable only from there is still reachable.

Worth extending the same idea to `tests/e2e/*.test.sh`, which has the same exposure.

## Tests first

Add a `tests/release/foo.test.sh` that nothing invokes and watch the new check name it. Then delete the `ci.yml` line for an existing one and watch it name that. Both directions, because a check that only fires on a new file misses the deletion case, which is how this one got lost.

## Difficulty

`easy`. One shell script and two lines of wiring, and the derivation is a glob plus a grep.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands. `bash scripts/ci-local.sh` runs the whole board locally. No CLA and no copyright waiver. The project is MIT.

vladimirrott:
[...]

PR #370, opened 2026-09-05
fix(evidence): derive story coverage claims

### LINKED ISSUE #229: Nothing checks the story-coverage figures in CONTRIBUTING.md, which is how they drifted

bferanmi806-sketch:
Story coverage figures are now derived from the tree instead of maintained by hand: uncovered All / Fedora / Ubuntu counts and atomic / Ubuntu family counts are recomputed from docs/action-reference.md and tests/e2e/stories on every run and compared against the published sentences. Stale Debian-only coverage prose is rejected in both directions (nonzero gap and derived zero), in every screened claim file. Mutation tests cover wrong published values, changed story evidence, swapped family labels, moved anchors and header locations, and both directions of Debian-only drift. Closes #229.


vladimirrott (approved):
Reviewed at `4cd96e035442b5a99d9908714cf1c2d26c6df4c8`.

The pristine-fixture re-assertion after every restore is what makes these nine mutations worth having. Each one restores the fixture and then confirms the restored tree still passes, so a rejection can only come from the mutation you just applied. Deriving both figures from the fixture with `read -r derived_all published_all < <(python3 ...)` instead of retyping today's 59 means these tests survive the next story landing, and reading `CLAIM_FILES` out of the checker rather than keeping a second copy is the same instinct. The comment explaining why that second copy was a trap earns its length.

I reviewed by reading the diff, and ran the executable parts in a container with `--network=none` and the repo mounted read-only.

**The guard bites.** I reverted `scripts/check_evidence_claims.py` to main's version and left your test file at the PR version:

```
$ git archive p370 | tar -x -C "$scratch"
$ git show main:scripts/check_evidence_claims.py > "$scratch/scripts/check_evidence_claims.py"
$ podman run --rm --network=none -v "$scratch:/repo:ro" -w /repo docker.io/library/python:3-slim bash -c 'bash tests/release/public-claims.test.sh'
FAIL: checker accepted stale claim: published story-coverage figure that disagrees with the tree
EXIT=1
```

The same command at your head, unmutated, exits 0.

**The derived figures are right.** I recounted them with my own script rather than yours:

```
rows: 189
duplicate names: {}
uncovered: {'All': 59, 'Fedora': 10, 'Ubuntu': 5}
```

That matches what `CONTRIBUTING.md` publishes, and the 5 matches the Debian-only sentence you corrected in `docs/introduction.md`.

**The line-2 comment is accurate.** `tests/e2e/run-stories.sh:160` reads `sed -n '2p'`, so narrowing the scan from six lines to one puts the checker and the harness on one rule.

**It fails loudly instead of deriving nothing.** Four input mutations, each exiting 1 with its own diagnostic:

```
--- no-stories -> exit 1
Evidence check could not run: could not derive story coverage from tests/e2e/stories; no story families found
--- bad-header -> exit 1
Evidence check could not run: could not derive a family from tests/e2e/stories/story-100.sh; expected the '# Story N (...): title' header on line 2
--- mangled-catalogue -> exit 1
Evidence check could not run: no action rows could be derived from docs/action-reference.md; the generated catalogue shape may have changed
--- no-quoted-names -> exit 1
Evidence check could not run: no quoted action-like identifiers found in tests/e2e/stories; story coverage could not be derived
```

The CI shellcheck sweep over `tests/e2e tests/release scripts assets/demo` is clean, and all four workflows are green at this head.

## Two optional findings. Neither blocks.

**1. A version number can read as a story count.** `check_bare_story_counts` now flattens the whole document with `re.sub(r"\s+", " ", text)` before matching. Combined with `\b(?P<count>[0-9]{2,})\s+(?P<family>atomic|ubuntu)\b`, the fractional part of a release number becomes a count when the following line opens with "Ubuntu". I appended this to `docs/quickstart.md`:

```
Supported releases include 22.04, 24.04 and 26.04
Ubuntu hosts use the apt action family.
```

and the checker said:

```
Claims that do not derive from evidence:
  - docs/quickstart.md: Ubuntu family story count has published 4, derived 79
```

Nothing in the tree triggers it today and it fails closed, so no claim slips through. The cost is a future docs reflow failing on a message nobody can act on. A lookbehind covers it:

```python
family = re.compile(
    r"(?<![.\d])\b(?P<count>[0-9]{2,})\s+(?P<family>atomic|ubuntu)\b", re.IGNORECASE
)
```

I ran both directions with that patch applied. The quickstart text above passes, and a genuine swapped label still fails with `atomic family story count has published 79, derived 54`.

**2. One mutation keeps a parallel copy.** The story-evidence block rebuilds the catalogue regex and the coverage count in its own heredoc, while your later mutations import the checker and call `mod.uncovered_action_counts`. If that regex drifts, the test's copy drifts with it and still agrees. You already use the import pattern three times; extending it to that block closes the gap.

Take either, both or neither. My approval does not depend on them.

## Merging

Approved at the head above. Landing it needs a step on our side that I cannot complete from here. Nothing is outstanding on yours, and the delay is not yours to fix.

## Next

If you want another, #252 is the closest thing to what you just did: two parsers read the story header and nine tag spellings pass unchecked, which is the same seam you tightened when you moved the family derivation onto line 2. I will hold it for you. #331 is the lighter alternative, requiring that every `tests/release/*.test.sh` is actually invoked by a gate, which I checked by hand on this PR and would rather not check by hand again.

One more thing, unrelated to the review. Three PRs in, you run Ubuntu and care how this behaves on a real machine. If you ever want to point SysKnife at your own box, the read-only surface is the place to start: `sysknife plan "show me disk usage and failed services"` proposes actions and executes nothing without your approval. No obligation, and it has no bearing on anything landing here.

bferanmi806-sketch:
#252 sounds like a good next one I’d be happy to take it.

PR #375, opened 2026-09-06
fix(daemon): drop unused sysknife-apt-pin-edit lock arm

### LINKED ISSUE #248: exclusive_resource has an arm for a helper name no installer produces

sonalisrisivani:
## Summary
- Remove the unreachable `sysknife-apt-pin-edit` arm from `exclusive_resource`. Installers only emit `apt-pin-edit`; keep `unattended-upgrade` and `apt-pin-edit`.
- Add a packaging scan that fails if any helper is matched by its source filename instead of the installed basename, and pin `SetAptPin` to the dpkg lock.

Fixes #248

## Test plan
- [x] `cargo test -p sysknife-daemon --test helper_install_coverage --locked` (5 passed)
- [x] `cargo test -p sysknife-daemon --test concurrency_guard exclusive_resource --locked` (`exclusive_resource_is_derived_from_the_action_argv` passed)
- [ ] Full `cargo nextest run --workspace --locked` and `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` were not run here (macOS host; `AptInstall` preview tests fail with `unsupported_platform`, and the GUI workspace members were not built). Maintainer should refresh `tests/evidence/workspace-tests.json` and the three prose figures — this PR adds one Rust test.

Assisted by Cursor Grok 4.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,837 Rust tests and 72 frontend tests** form the current deterministic
+**1,838 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/crates/sysknife-daemon/src/actions/mod.rs
+++ b/crates/sysknife-daemon/src/actions/mod.rs
@@ -149,16 +149,8 @@ pub fn exclusive_resource(spec: &ActionSpec) -> Option<ExclusiveResource> {
             // Match on the binary name so an absolute path still resolves.
             let name = token.rsplit('/').next().unwrap_or(token);
             match name {
-                "apt-get"
-                | "apt"
-                | "apt-mark"
-                | "aptitude"
-                | "dpkg"
-                | "dpkg-reconfigure"
-                | "add-apt-repository"
-                | "do-release-upgrade"
-                | "unattended-upgrade"
-                | "sysknife-apt-pin-edit"
+                "apt-get" | "apt" | "apt-mark" | "aptitude" | "dpkg" | "dpkg-reconfigure"
+                | "add-apt-repository" | "do-release-upgrade" | "unattended-upgrade"
                 | "apt-pin-edit" => Some(ExclusiveResource::Dpkg),
                 "snap" => Some(ExclusiveResource::Snap),
                 "rpm-ostree" | "ostree" => Some(ExclusiveResource::RpmOstree),
--- a/crates/sysknife-daemon/tests/helper_install_coverage.rs
+++ b/crates/sysknife-daemon/tests/helper_install_coverage.rs
@@ -104,6 +104,98 @@ fn every_installed_helper_is_removed_by_uninstall() {
     }
 }
 
+/// Helpers are installed under the basename after `sysknife-` (`apt-pin-edit`,
+/// not `sysknife-apt-pin-edit`). `exclusive_resource` matches argv program
+/// names, so a source-filename arm can never fire.
+fn packaged_helper_scripts() -> Vec<(String, String)> {
+    let packaging = repo_root().join("packaging");
+    let mut helpers = Vec::new();
+    for entry in std::fs::read_dir(&packaging).unwrap_or_else(|e| panic!("read {packaging:?}: {e}"))
+    {
+        let path = entry.expect("dir entry").path();
+        if path.is_dir() {
+            continue;
+        }
+        let source_name = path
+            .file_name()
+            .and_then(|n| n.to_str())
+            .expect("utf-8 packaging filename")
+            .to_string();
+        // Helper scripts are extensionless `sysknife-*` files. Skip sudoers,
+        // unit files, and sysusers/tmpfiles snippets.
+        if !source_name.starts_with("sysknife-") || source_name.contains('.') {
+            continue;
+        }
+        if source_name == "sysknife-sudoers" {
+            continue;
+        }
+        let installed = source_name
+            .strip_prefix("sysknife-")
+            .expect("prefix already checked")
+            .to_string();
+        helpers.push((source_name, installed));
+    }
+    helpers.sort();
+    assert!(
+        helpers.len() >= 8,
+        "packaging/ helper scan found only {helpers:?}; the scan itself is probably broken"
+    );
+    helpers
+}
+
+fn exclusive_resource_matcher_source() -> String {
+    let path = repo_root().join("crates/sysknife-daemon/src/actions/mod.rs");
+    let text = std::fs::read_to_string(&path).unwrap_or_else(|e| panic!("read {path:?}: {e}"));
+    let start = text
+        .find("pub fn exclusive_resource")
+        .expect("exclusive_resource must exist in actions/mod.rs");
+    let rest = &text[start..];
+    let end = rest[1..]
+        .find("\npub fn ")
+        .expect("exclusive_resource must be followed by another pub fn")
+        + 1;
+    rest[..end].to_string()
+}
+
+/// Every helper in `packaging/` must be named in `exclusive_resource` under the
+/// basename installers emit, never the source filename. Helpers that do not
+/// contend for a package-manager lock are absent under both names.
+#[test]
+fn exclusive_resource_matches_packaging_helpers_by_installed_basename() {
+    use sysknife_daemon::actions::{all_specs, exclusive_resource, ExclusiveResource};
+
+    let matcher = exclusive_resource_matcher_source();
+    let helpers = packaged_helper_scripts();
+
+    for (source_name, installed) in &helpers {
+        let source_arm = format!("\"{source_name}\"");
+        let installed_arm = format!("\"{installed}\"");
+        assert!(
+            !matcher.contains(&source_arm),
+            "exclusive_resource must not match packaging source filename `{source_name}`; \
+             both installers emit `{installed}`"
+        );
+        if matcher.contains(&source_arm) || matcher.contains(&installed_arm) {
+            assert!(
+                matcher.contains(&installed_arm),
+                "`{source_name}` belongs in exclusive_resource as installed basename \
+                 `{installed}`, not the source filename"
+            );
+        }
+    }
+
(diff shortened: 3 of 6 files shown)


vladimirrott:
This is a strong first PR, and it closes #248 the right way.

You did not only delete the dead arm. You added a packaging scan that fails if any helper is ever matched by its source filename instead of its installed basename, and you pinned `SetAptPin` to the dpkg lock on the way past. That is the move this project runs on: fix the case, then make the case impossible to reintroduce. A full review is coming shortly.

Two more I think suit you, and I have reserved both for you:

- **#234** — `UfwDeleteRule` takes a rule number that nothing in the catalogue can produce. The same shape you just handled: an action arm that does not line up with reality.
- **#251** — Cancel and the TTL sweep end an approval without revoking it or writing a transaction row. Deeper in the daemon, and right where you were just working.

No deadline on either, and if one does not appeal, say so and I will swap it. Reply on either issue and it is formally yours (GitHub only lets me assign someone once they have chimed in on the thread).

vladimirrott (dismissed):
Approving. I read the diff and verified it end to end.

The removed arm was dead. `packaging/sysknife-apt-pin-edit` installs as `/usr/lib/sysknife/apt-pin-edit` (`tests/e2e/ubuntu-provision.sh:283`), and `exclusive_resource` matches on the argv basename, so `sysknife-apt-pin-edit` could never appear at runtime.

I mutation-proved your test. Restoring the arm you deleted turns `exclusive_resource_matches_packaging_helpers_by_installed_basename` red with exactly the message a maintainer wants:

```
exclusive_resource must not match packaging source filename `sysknife-apt-pin-edit`; both installers emit `apt-pin-edit`
```

The `SetAptPin -> Dpkg` assertion pins the real behavior too, so the fix and its guard travel together. That is the pattern this project runs on.

Merging once CI is green.

vladimirrott:
One thing your PR tripped, and it is pure maintainer bookkeeping rather than anything wrong with your change: the workspace has a pinned test count, and adding a test moves it. Your new test took the suite from 1,837 to 1,838, so the `rust` job's `test_baseline` step failed until the count caught up.

I pushed the bump to your branch (`tests/evidence/workspace-tests.json` plus the three published `1,837` figures in `README.md`, `docs/introduction.md` and `docs/distro-support.md`). Next time it is one command: `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` for the artifact, then the three doc figures by hand. Nothing you needed to know in advance.

CI is re-running now, and this merges as soon as it is green.

vladimirrott (changes requested):
Reviewed at `eaa0374e9ccf5512bc62c6618216272f4cc7f60f`.

You did what #248 asked and then the harder half: the deletion is one line, and instead of leaving a comment saying "do not do this again" you derived the rule from `packaging/` itself. Naming the guard after the installed basename is the right axis, because that is the axis the two installers vary on.

I checked the premise before crediting the fix, since a matcher arm is only dead if nothing can reach it:

```
$ grep -rn "apt-pin-edit" --include='*.rs' --include='*.sh' --include='*.md' --include='*.toml' . | grep -v '^./target' | head -40
packaging/sysknife-sudoers:158:sysknife ALL=(root) NOPASSWD: /usr/lib/sysknife/apt-pin-edit *
crates/sysknife-daemon/src/actions/apt_preferences.rs:11:const HELPER: &str = "/usr/lib/sysknife/apt-pin-edit";
tests/e2e/ubuntu-provision.sh:283:install -Dm 755 packaging/sysknife-apt-pin-edit /usr/lib/sysknife/apt-pin-edit \
```

`exclusive_resource` reads `spec.mechanism`'s program and args, and the only argv the daemon emits for `SetAptPin` is that installed path, so `token.rsplit('/')` can yield nothing but `apt-pin-edit`. The arm you removed could not fire.

## What I verified

Everything below ran inside rootless podman with the repo mounted read-only and no network, because CI compiles and runs your code and this host holds other people's keys. `pc.sh` is that wrapper.

The suite passes at your head, and your five is five:

```
$ ./pc.sh cargo test -p sysknife-daemon --test helper_install_coverage --locked --offline
running 5 tests
test exclusive_resource_matches_packaging_helpers_by_installed_basename ... ok
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

The guard bites in both directions. Putting the source-filename arm back:

```
$ ./pc.sh cargo test -p sysknife-daemon --test helper_install_coverage --locked --offline
test exclusive_resource_matches_packaging_helpers_by_installed_basename ... FAILED
panicked at crates/sysknife-daemon/tests/helper_install_coverage.rs:173:9:
exclusive_resource must not match packaging source filename `sysknife-apt-pin-edit`; both installers emit `apt-pin-edit`
```

And removing the arm that does fire, which is the half that protects the live lock:

```
$ ./pc.sh cargo test -p sysknife-daemon --test helper_install_coverage --locked --offline
panicked at crates/sysknife-daemon/tests/helper_install_coverage.rs:192:5:
assertion `left == right` failed: apt-pin-edit is the installed basename and must contend for the dpkg lock
  left: None
 right: Some(Dpkg)
```

Your 1,838 is correct. I could not run `scripts/test_baseline.sh` on your branch, because the container I review in has no WebKit or GTK and `sysknife-shell` will not compile there, so I measured the delta on both refs with that crate excluded and added it back from a host run of `main`:

```
$ ./pc.sh cargo nextest run --workspace --exclude sysknife-shell --locked --offline
    Starting 1823 tests across 40 binaries (6 tests skipped)
$ ./pc.sh cargo nextest run --workspace --exclude sysknife-shell --locked --offline --no-fail-fast
    Starting 1822 tests across 40 binaries (6 tests skipped)
$ scripts/test_baseline.sh
    Starting 1837 tests across 41 binaries (6 tests skipped)
test_baseline: 1837 rust tests, matching the recorded baseline.
```

1823 − 1822 = +1 on your branch; 1837 − 1822 = 15 in `sysknife-shell`; 1823 + 15 = 1838. Your board is green on every check, `docs-and-hygiene` included.

## Blocking: the second assertion in the new test cannot fail

```rust
assert!(
    !matcher.contains(&source_arm),
    ...
);
if matcher.contains(&source_arm) || matcher.contains(&installed_arm) {
    assert!(
        matcher.contains(&installed_arm),
        ...
    );
}
```

Once the first `assert!` has passed, `matcher.contains(&source_arm)` is false, so the branch condition reduces to `matcher.contains(&installed_arm)`, which is the predicate the inner `assert!` then checks. It holds by construction. No input reaches it as a failure.

I proved it rather than argued it. Deleting those seven lines leaves the test passing:

```
$ ./pc.sh cargo test -p sysknife-daemon --test helper_install_coverage --locked --offline
test exclusive_resource_matches_packaging_helpers_by_installed_basename ... ok
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

and still failing on the mutation, with the same message and the same line:

```
$ ./pc.sh cargo test -p sysknife-daemon --test helper_install_coverage --locked --offline
panicked at crates/sysknife-daemon/tests/helper_install_coverage.rs:173:9:
exclusive_resource must not match packaging source filename `sysknife-apt-pin-edit`; both installers emit `apt-pin-edit`
```

So the block carries no coverage. `CLAUDE.md` asks that every line be reachable and load-bearing, and an assertion that cannot fail is the shape this repo's reviews exist to catch. Delete it and the first `assert!` says the whole rule on its own.

That is the only thing I am asking you to change.

## Not yours, and not blocking

The evidence artifact moved `tests` and `commit.tests` but left `measured_at.tests` at `2026-09-04T00:03:28+08:00`. `scripts/record_test_baseline.py` writes all three in one call, so the file now attributes a 2026-09-04 measurement to a commit dated today. The number is right, I recounted it above, and you said plainly in the PR body that you could not run the suite on macOS. Leaving those four files untouched is what I would rather have than a hand-edited figure, so this one is mine to regenerate at merge.

## While I was in here

`packaged_helper_scripts()` asserts `helpers.len() >= 8` before it does anything, which is the check most guards of this shape forget. That is why I trusted the scan instead of re-deriving it.

You already have #234 open from me and unanswered, so I am not putting another in front of you. It is still yours if you want it.

You read `packaging/`, the sudoers grants and the daemon's argv construction against each other to prove one arm was dead. That is somebody who runs real machines, not somebody reading a diff. If you keep Linux alive anywhere that matters at work, servers you patch or a fleet you are responsible for, I would like to know how SysKnife behaves there. `npx sysknife-setup` is the whole setup, and the read-only side (`sysknife doctor`, `sysknife history`, `sysknife audit verify`) changes nothing on the host. Easy to decline, and it has no bearing on this merge.

Issue #346, opened 2026-09-02
ci-local.sh misses seven scripts CI runs, and skips a required check without saying so


vladimirrott:
> **Re-measured 2026-09-09 at `5673d20`.** Both totals moved again, and the
> drift is the same seven scripts. Twenty local, twenty-seven in CI, which is
> every `*.test.sh` on disk. `CONTRIBUTING.md` moved from :182 to :212.
>
> ```
> $ grep -oE 'tests/(release|e2e)/[a-z0-9-]+\.test\.sh' scripts/ci-local.sh | sort -u | wc -l
> 20
> $ grep -rhoE 'tests/(release|e2e)/[a-z0-9-]+\.test\.sh' .github/workflows/ | sort -u | wc -l
> 27
> $ ls tests/release/*.test.sh tests/e2e/*.test.sh | wc -l
> 27
> ```

`CONTRIBUTING.md:212` asks for one thing before a PR:

> Run `scripts/ci-local.sh`. A compile error in a crate your machine cannot build is still a compile error, and finding it costs a review round.

A green `ci-local.sh` does not mean a green board, in two independent ways.

## 1. The script list drifted from what CI runs

`run_hygiene_group` names twenty test scripts by hand (`ci-local.sh:265-285`).
CI runs twenty-seven. Seven run in CI and nowhere locally:

```
cassette-replay-parity.test.sh
codex-plugin-manifest.test.sh
grub-kargs-edit.test.sh
log-edit.test.sh
mount-edit.test.sh
no-secrets.test.sh
rmswap.test.sh
```

Four of those seven guard privileged helper scripts under `packaging/`, which is
the surface where a mistake costs the most.

This is the defect class this tracker keeps finding: a hand-maintained list that
drifted from what is on disk. `STORIES=(1 2 3 4 5 6 7 8 9 10 11)` against twelve
`exec-*.sh` files (#330) and six claim files against sixteen were both this.

## 2. A required check skips silently

`run_postgres_contract_group` records `SKIP` when neither docker nor podman is
installed:

```bash
if [[ -z "$runtime" ]]; then
    record SKIP "${label} (no SYSKNIFE_TEST_POSTGRES_URL and no docker/podman found)"
    return
fi
```

`postgres-contract` is one of the five required checks. A contributor with no
container runtime gets an all-green local board and a red required check on push,
with nothing in the local output saying which gate did not run.

## Why it matters

Both failures point the same way: the contributor did what `CONTRIBUTING.md`
[...]

PR #376, opened 2026-09-06
fix(ci): derive markdown link coverage

### LINKED ISSUE #372: The markdown-link-check gate names ten files by hand, in two places, out of fifty-four
The `markdown-link-check` gate names the files it checks by hand, in two places,
and the two lists have to agree with each other and with the tree. Ten files are
named. The repository tracks fifty-four Markdown files, so forty-four of them
have never had a link resolved by anything.

`.github/workflows/ci.yml:139-149`:

```yaml
          for file in \
            README.md \
            CONTRIBUTING.md \
            SECURITY.md \
            CODE_OF_CONDUCT.md \
            ROADMAP.md \
            docs/architecture.md \
            docs/developer-guide.md \
            docs/adr/0001-system-boundaries.md \
            docs/adr/0002-brain-provider-layer.md \
            docs/adr/0003-ipc-wire-protocol.md
          do
```

`scripts/ci-local.sh:239-250` repeats the same ten as a bash array, so a file
added to one list and not the other gets checked on exactly one of the two
surfaces, and nothing says which.

Counted at `7d19864`:

```
$ git ls-tree -r --name-only 7d19864 | grep '\.md$' | grep -v node_modules | wc -l
54
$ awk 'NR>=139 && NR<=150' .github/workflows/ci.yml | cat -n
     1	          for file in \
     2	            README.md \
     3	            CONTRIBUTING.md \
     4	            SECURITY.md \
     5	            CODE_OF_CONDUCT.md \
     6	            ROADMAP.md \
     7	            docs/architecture.md \
     8	            docs/developer-guide.md \
     9	            docs/adr/0001-system-boundaries.md \
    10	            docs/adr/0002-brain-provider-layer.md \
    11	            docs/adr/0003-ipc-wire-protocol.md
    12	          do
```

## Why it matters

`docs/cli.md` is the CLI reference, and it is not on the list. Neither is
`docs/contributing/testing.md`, `docs/contributing/CONTRIBUTING.md` or
`docs/the-audit-chain.md`. Those four are exactly the pages carrying the seven
broken links in #371. A gate that reads ten
files out of fifty-four measures those ten files.

This is the same defect the tracker keeps finding: a list somebody keeps by hand
beside a set that grows on its own. #346 is the live example, recounted at
`7d19864` this morning:

```
$ git show 7d19864:scripts/ci-local.sh | grep -oE 'tests/(release|e2e)/[a-z0-9-]+\.test\.sh' | sort -u | wc -l
17
$ git ls-tree -r --name-only 7d19864 tests/release/ tests/e2e/ | grep -E '\.test\.sh$' | wc -l
24
```

`CLAIM_FILES` in `scripts/check_evidence_claims.py` is a third list of the same
kind, sixteen paths written out by hand.

## Scope

- Derive the file set. A glob over tracked Markdown is the obvious move, with an
  explicit exclusion list for anything deliberately skipped and a comment saying
  why each entry is on it.
- Share one list between `ci.yml` and `ci-local.sh` rather than writing it twice.
  A small script both call is the cheapest form of that.
- Expect the first honest run to be red. Forty-four files have never been checked,
  and the odds that all of them are clean are poor. Fixing what it finds is part
  of the work, so read the output before you decide how wide to make the glob.
- The trap: `markdown-link-check` makes network requests for every external URL.
  Forty-four files instead of ten multiplies the request count and with it the
  flake rate, which is why the check already carries a three-attempt retry at
  `ci.yml:160-169`. Watch the runtime and consider excluding external links
  from the wider set, checking those only on the ten that carry them today.

## Tests first

The guard is the derivation: once the list comes from a glob, a new Markdown file
is covered on the commit that adds it, and there is no second list to drift from.
What still needs pinning is the exclusion list, since that is the part a person
edits.

Assert that every path in the exclusion list exists. An exclusion naming a file
that was renamed or deleted is a silent hole, and it reads as deliberate. Prove it bites by adding a bogus path to the exclusions and watching the
check fail with that path in the message.

## Difficulty

`easy`. The glob is one line. Reading whatever the first wide run turns up, and
deciding what belongs in the exclusion list, is the work.

## Getting started

has the build and test commands. This one needs no VM, no daemon and no LLM key:
`npm install --global markdown-link-check@3.15.0` is the whole environment. No CLA
and no copyright waiver. The project is MIT.

be-student:
## Summary

- derive the internal link-check set from every tracked Markdown file
- keep external URL checks bounded to the ten existing documents
- reject stale exclusions and external-check entries with a regression test
- use the same discovery script in GitHub Actions and `ci-local.sh`

## Related Issue

Closes #372

## Validation

- [x] `bash tests/release/markdown-link-files.test.sh`
- [x] complete internal scan: 54 files
- [x] complete external scan: 10 files
- [x] ShellCheck, markdownlint-cli2, yamllint, evidence claims, and `git diff --check`
- [x] Documentation updated
- [x] Security impact considered; no trust-boundary behavior changed

## Notes for Reviewers

`npm test --prefix packages/setup` passes 101/102 locally. The Unix-socket reachability test fails identically on an untouched `origin/main` worktree on macOS.

Implementation and validation were performed with Codex.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,9 @@ jobs:
       - name: Check the audit export states its confidentiality class
         run: bash tests/release/audit-export-confidentiality.test.sh
 
+      - name: Check markdown link file discovery
+        run: bash tests/release/markdown-link-files.test.sh
+
       - name: Check provider parity across E2E entry points
         run: bash tests/e2e/provider-parity.test.sh
 
@@ -134,20 +137,15 @@ jobs:
             docs/developer-guide.md \
             docs/adr/*.md
 
-      - name: Check markdown links
+      - name: Check internal markdown links
         run: |
-          for file in \
-            README.md \
-            CONTRIBUTING.md \
-            SECURITY.md \
-            CODE_OF_CONDUCT.md \
-            ROADMAP.md \
-            docs/architecture.md \
-            docs/developer-guide.md \
-            docs/adr/0001-system-boundaries.md \
-            docs/adr/0002-brain-provider-layer.md \
-            docs/adr/0003-ipc-wire-protocol.md
-          do
+          while IFS= read -r -d '' file; do
+            markdown-link-check --config .markdown-link-check-internal.json "$file"
+          done < <(scripts/markdown-link-files.sh)
+
+      - name: Check external markdown links
+        run: |
+          while IFS= read -r -d '' file; do
             # Retry on transient network failure. `retryOn429` in the config only
             # covers rate limiting, and the failures actually seen here were
             # Status 0 — a connection/DNS blip on the runner against hosts that
@@ -168,7 +166,7 @@ jobs:
               echo "link check for $file failed (attempt $attempt/3); retrying in 10s"
               sleep 10
             done
-          done
+          done < <(scripts/markdown-link-files.sh --external)
 
       - name: Lint GitHub issue templates
         run: yamllint .github/ISSUE_TEMPLATE/*.yml .github/workflows/*.yml
--- a/.markdown-link-check-internal.json
+++ b/.markdown-link-check-internal.json
@@ -1,0 +1,11 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^https?://"
+    },
+    {
+      "pattern": "\\.html(?:[#?].*)?$"
+    }
+  ],
+  "_comment": "The repository-wide pass checks source-relative links without making network requests. Relative .html links target mdBook output and are checked after the book is built, not against the Markdown source tree. The bounded external pass uses .markdown-link-check.json."
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,6 +148,13 @@ purpose: this file is screened by `scripts/check_evidence_claims.py`, and an
 illustrative figure reads as a published claim. That screen caught this very
 paragraph.)
 
+**Markdown links are discovered from the tracked tree.** Both local and remote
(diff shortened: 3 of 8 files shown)


vladimirrott:
Thank you for this. It is exactly what #372 needed, and you took it further than the issue asked.

The issue was ten files named by hand in two places. You made the internal set derive from every tracked Markdown file, kept the external URL checks bounded, added a regression test that rejects stale entries in both directions, and ran the same discovery script in CI and `ci-local.sh` so the two cannot drift. That last part is the detail most people skip. A full review is coming shortly.

Two more that fit what you just showed, both reserved for you:

- **#331** — nothing requires that a `tests/release/*.test.sh` is actually invoked by a gate, so a test can exist and never run. You just wired one in, so you have the map.
- **#346** — `ci-local.sh` misses seven scripts CI runs and skips a required check without saying so. The same file you were in, one layer up.

No rush on either, and say the word if you would rather trade one. Reply on either issue and it is formally yours (GitHub only lets me assign someone once they have posted on the thread).

vladimirrott (approved):
Approving. Verified end to end.

You replaced ten files named by hand in two places with discovery from `git ls-files '*.md'` minus a validated exclusion list, kept the external checks bounded, and ran the same script in CI and `ci-local.sh` so the two cannot drift. Making the regression test hermetic, with its own fixture repo, was the right call.

I mutation-proved it: removing the external-path validation from the script turns the test red with `stale external-check path unexpectedly passed`. The guard bites.

The workflow change is safe to run: a `pull_request` job with no secret access, swapping a hand-kept file list for the discovery script.

Merging once CI is green.

Issue #367, opened 2026-09-04
npm audit makes `frontend` fail whenever registry.npmjs.org does


vladimirrott:
> **Recounted 2026-09-21 at `ba1a7b8f`.** The step is now at **240-241**. This is
> the fourth line number this issue has carried; `35b3c1a` (#454) and `1d1ab2d`
> (#423) both added steps above it. The defect is unchanged.

> **Line references corrected 2026-09-07 at `adab560`.** The three line numbers
> in this issue were wrong when it was filed, and the `grep -n` output pasted
> under "Getting started" reproduces on no commit this repository has. My error,
> not a change under you. The step was at **197-198** then; it is at **223-224**
> at `b2c823e6`, recounted 2026-09-16, and the substance of the issue is unchanged:
> `frontend` is still required, the step is still enforced, and it still gates
> every pull request on `registry.npmjs.org`.

`frontend` is a required status check on `main`, and its fourth step is:

```yaml
- name: Audit production dependencies
  run: npm audit --omit=dev --audit-level=high
```

That command POSTs to `https://registry.npmjs.org/-/npm/v1/security/advisories/bulk`.
When the endpoint is unavailable, npm exits 1, and the check fails exactly the way
it fails for a genuine high-severity advisory.

This morning that took every open pull request down for 45 minutes:

| run | PR | started (UTC) | error |
|---|---|---|---|
| [33859810636](https://github.com/lacs-project/sysknife/actions/runs/33859810636) | #365 | 09:44 | `npm warn audit 503 Service Unavailable - POST .../advisories/bulk` |
| [33862001249](https://github.com/lacs-project/sysknife/actions/runs/33862001249) | #366 | 10:11 | `npm warn audit network timeout at: .../advisories/bulk` |
| [33863121379](https://github.com/lacs-project/sysknife/actions/runs/33863121379) | #366 | 10:25 | `503 Service Unavailable` |

Each attempt ran for five minutes before npm gave up (10:25:37 → 10:30:37 on the
last one), so a bad morning at the registry costs runner minutes on top of the
blocked merges. Re-running the failed job cleared all three.

Two things make this worse than an ordinary flake.

The audit covers `apps/sysknife-shell`, the paused Tauri GUI. `CONTRIBUTING.md`
puts GUI work out of scope, so #366, which changes one line of `CONTRIBUTING.md`,
waited twice on that directory's dependency advisories.

And `gh pr checks` prints `frontend fail` either way. A contributor reading that
[...]

PR #378, opened 2026-09-06
fix(ci): reject missing internal version pins

### LINKED ISSUE #368: check_release_versions.sh counts the pins it can see, and never notices one leaving its view
`scripts/check_release_versions.sh` counts the internal dependency pins it can
see and has no idea how many it should see. Take one out of its view and the
check reports success over the smaller set.

## Reproduction

```sh
$ bash scripts/check_release_versions.sh
All release versions match 0.13.0 (15 internal dependency pins checked).

$ sed -i 's|^sysknife-core = { path = "\(\.\./sysknife-core\)", version = "[^"]*" }|sysknife-core = { path = "\1" }|' \
    crates/sysknife-brain/Cargo.toml

$ bash scripts/check_release_versions.sh
All release versions match 0.13.0 (14 internal dependency pins checked).
$ echo $?
0
```

That transcript is from 2026-09-04, before the 0.13.1 release. Re-run it today and
both lines read `0.13.1`; the counts and the exit code are unchanged:

```
$ sed -n 's/^version = "\([^"]*\)"/\1/p' apps/sysknife-cli/Cargo.toml | head -1
0.13.1
$ grep -rn '^sysknife-[a-z-]* = {' crates/*/Cargo.toml apps/sysknife-cli/Cargo.toml apps/sysknife-shell/src-tauri/Cargo.toml | wc -l
15
```

Fifteen became fourteen and the gate said match. The mechanism is the two-stage
filter at `scripts/check_release_versions.sh:62-66`:

```sh
pins="$(grep -rn '^sysknife-[a-z-]* = {' \
    "$repo_root"/crates/*/Cargo.toml \
    "$repo_root"/apps/sysknife-cli/Cargo.toml \
    "$repo_root"/apps/sysknife-shell/src-tauri/Cargo.toml |
    grep 'version = ' || true)"
```

The second `grep` is a filter, so a matched dependency line without `version = `
is discarded rather than flagged. `pin_count` then counts survivors, and the
`-z "$pins"` guard below only fires when *every* pin disappears.

## What this is not

I went looking for a publish hazard and there isn't one. Cargo refuses:

```
$ cargo publish -p sysknife-brain --dry-run --allow-dirty --no-verify
error: failed to verify manifest at .../crates/sysknife-brain/Cargo.toml

Caused by:
  all dependencies must have a version requirement specified when publishing.
  dependency `sysknife-core` does not specify a version
```

So the specific case above cannot reach crates.io. The file's own comment
describes a *stale* pin rather than a missing one, and the stale case is caught
correctly by the `pinned != baseline` branch.

## What it is

The gate cannot distinguish "15 pins, all correct" from "14 pins, all correct,
and one that left my field of view". Anything that moves a dependency line out
of the anchor's reach shrinks the check silently:

- a crate added under a path not in the three globs
- `sysknife-foo = { workspace = true }`, which the `^sysknife-` anchor matches
  but which carries no `version =` on the line, so it is filtered out
- a manifest reformatted so the declaration wraps across lines

None of those is hypothetical in a workspace with fifteen internal pins and a
paused GUI crate that still ships manifests. Today all 15 anchor matches carry
`version = `, so the check is at full coverage; nothing tells you when it stops
being.

Same family as #347 and #355: a guard that quietly stops protecting.

## Two shapes

- **Stop filtering, start failing.** Keep every `^sysknife-` line and report the
  ones without a `version =` instead of dropping them. Smallest change, catches
  the `workspace = true` case too.
- **Pin the count.** The workspace member list is derivable from the root
  `Cargo.toml`, so the expected number of internal pins is derivable too. Then
  "14 where 15 were expected" is a failure rather than a sentence.

The first is enough on its own. The second also catches a crate that stops
declaring an internal dependency it should have.

Whoever picks this up: the guard test to write alongside is the reproduction
above, and it must fail with the fix reverted.

## Getting started

One shell script and a handful of `Cargo.toml` files. No VM, no daemon, no
provider, no credentials, and no `cargo` run needed to see the bug.

The two-stage filter is the whole mechanism:

```
$ sed -n '55,80p' scripts/check_release_versions.sh
```

The first stage globs three paths for anchor lines, the second keeps only those
carrying `version = `. `pin_count` then counts the survivors, so a pin that
disappears from either stage lowers the count and nothing compares that count to
anything.

Today all fifteen anchors survive both stages:

```
$ grep -rn '^sysknife-[a-z-]* = {' crates/*/Cargo.toml apps/sysknife-cli/Cargo.toml apps/sysknife-shell/src-tauri/Cargo.toml | wc -l
15
$ grep -rhn '^sysknife-[a-z-]* = {' crates/*/Cargo.toml apps/sysknife-cli/Cargo.toml apps/sysknife-shell/src-tauri/Cargo.toml | grep -c 'version = '
15
```

The expected count has to come from somewhere the script cannot lose sight of. The
workspace member list in the root `Cargo.toml` is the natural anchor: eight members
today, all covered by the three globs, so a crate added outside them is the case
that escapes.

Reverting the reproduction is `git checkout crates/sysknife-brain/Cargo.toml`.

## Tests first

Two directions, and the second is the point of the issue.

**Break what the guard protects.** Remove `version = ` from one internal pin, run
the script, and assert it exits non-zero and names the crate that lost its pin.
Then restore it and assert it exits zero. That is the reproduction above turned
into an assertion.

**Break the guard's own input.** Point the script at a fixture tree with no
matching manifests at all and assert it fails rather than reporting a cheerful
match over nothing. Then add a fixture crate outside the three globs and assert it
is either counted or reported as unreachable. A version check that passes over zero
pins is the exact failure this issue describes, one step further along.

Same family as #347 and #355, both closed, and the same shape as #346 and #362.

## Difficulty

`medium`. The script edit is small. Deciding where the expected count comes from,
and making sure that source cannot drift in the same way, is the work.

be-student:
## Summary

Fail when an internal path dependency drops its inline version instead of silently checking a smaller set. Adds a mutation regression wired into CI and `ci-local.sh`.

## Related Issue

Closes #368.

## Validation

- New regression failed before the fix and passes now
- Pin check, ShellCheck, actionlint, Bash syntax, and evidence claims pass
- Full Rust suite is blocked on clean macOS `main` by the existing Linux-only `SocketTarget::Vsock` reference in `apps/sysknife-cli/src/runner.rs:114`

Assisted by Codex; I reviewed the diff and validation output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,9 @@ jobs:
       - name: Check release version consistency
         run: bash scripts/check_release_versions.sh
 
+      - name: Test internal dependency pin coverage
+        run: bash tests/release/release-version-pins.test.sh
+
       - name: Check public support and approval claims
         run: bash tests/release/public-claims.test.sh
 
--- a/docs/release.md
+++ b/docs/release.md
@@ -136,6 +136,10 @@ git tag -s v0.2.5 -m "SysKnife v0.2.5"
 git push origin v0.2.5
 ```
 
+The version check requires every internal path dependency to carry an inline
+`version` matching the workspace release. Removing that field is an error even
+when all remaining visible pins match.
+
 The tag pattern does not accept prerelease suffixes. Do not move or reuse a
 published tag. If publication partly fails, diagnose and rerun the workflow on
 the same commit; do not publish a different tree under the same version.
--- a/scripts/check_release_versions.sh
+++ b/scripts/check_release_versions.sh
@@ -62,8 +62,7 @@ fi
 pins="$(grep -rn '^sysknife-[a-z-]* = {' \
     "$repo_root"/crates/*/Cargo.toml \
     "$repo_root"/apps/sysknife-cli/Cargo.toml \
-    "$repo_root"/apps/sysknife-shell/src-tauri/Cargo.toml |
-    grep 'version = ' || true)"
+    "$repo_root"/apps/sysknife-shell/src-tauri/Cargo.toml || true)"
 
 # An empty result means the manifests moved, not that every pin agrees. Fail
 # loudly rather than reporting success for a check that inspected nothing.
@@ -77,6 +76,11 @@ pin_count=0
 while IFS= read -r pin; do
     pin_count=$((pin_count + 1))
     pinned="$(printf '%s' "$pin" | sed -n 's/.*version = "\([^"]*\)".*/\1/p')"
+    if [[ -z "$pinned" ]]; then
+        printf 'Internal dependency is missing an explicit version pin:\n  %s\n' \
+            "$pin" >&2
+        exit 1
+    fi
     if [[ "$pinned" != "$baseline" ]]; then
         printf 'Internal dependency pin does not match package version %s:\n  %s\n' \
             "$baseline" "$pin" >&2
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -262,6 +262,7 @@ run_hygiene_group() {
     printf '\n### hygiene\n'
     run_step 'hygiene: check_repo_completeness.sh' bash "$repo_root/scripts/check_repo_completeness.sh"
     run_step 'hygiene: check_release_versions.sh' bash "$repo_root/scripts/check_release_versions.sh"
+    run_step 'hygiene: release-version-pins.test.sh' bash "$repo_root/tests/release/release-version-pins.test.sh"
     run_step 'hygiene: public-claims.test.sh' bash "$repo_root/tests/release/public-claims.test.sh"
     run_step 'hygiene: npm test --prefix packages/setup' npm test --prefix "$repo_root/packages/setup"
     run_step 'hygiene: registry-manifest.test.sh' bash "$repo_root/tests/release/registry-manifest.test.sh"
--- a/tests/release/release-version-pins.test.sh
+++ b/tests/release/release-version-pins.test.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+fixture="$(mktemp -d)"
+trap 'rm -rf "$fixture"' EXIT
+
+paths=(
+    scripts/check_release_versions.sh
+    apps/sysknife-cli/Cargo.toml
+    apps/sysknife-shell/package.json
+    apps/sysknife-shell/package-lock.json
+    apps/sysknife-shell/src-tauri/Cargo.toml
+    apps/sysknife-shell/src-tauri/tauri.conf.json
+    crates/sysknife-brain/Cargo.toml
+    crates/sysknife-core/Cargo.toml
+    crates/sysknife-daemon-test/Cargo.toml
+    crates/sysknife-daemon/Cargo.toml
+    crates/sysknife-proto/Cargo.toml
+    crates/sysknife-types/Cargo.toml
+    packages/setup/package.json
+    .codex-plugin/plugin.json
+    server.json
+)
+
+for path in "${paths[@]}"; do
+    mkdir -p "$fixture/$(dirname "$path")"
+    cp "$repo_root/$path" "$fixture/$path"
+done
+
+bash "$fixture/scripts/check_release_versions.sh" >/dev/null
+
+python3 - "$fixture/crates/sysknife-brain/Cargo.toml" <<'PY'
+from pathlib import Path
+import re
+import sys
+
+manifest = Path(sys.argv[1])
+contents = manifest.read_text()
+mutated, replacements = re.subn(
+    r'^(sysknife-core = \{ path = "[^"]*"), version = "[^"]*"( \})$',
+    r'\1\2',
+    contents,
+    count=1,
+    flags=re.MULTILINE,
+)
+if replacements != 1:
+    raise SystemExit("fixture could not remove the sysknife-core version pin")
+manifest.write_text(mutated)
+PY
+
+if output="$(bash "$fixture/scripts/check_release_versions.sh" 2>&1)"; then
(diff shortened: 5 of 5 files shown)


vladimirrott (approved):
Reviewed at `6437d7750c128702894d8909fdd88b401758db51`.

This is the best kind of finding: the guard was reporting success while looking at nothing. `grep 'version = '` sat between the pin list and the loop that checks it, so a pin with no `version` field was filtered out before anything could object to it. Removing that stage and adding the `[[ -z "$pinned" ]]` branch is the correct fix, and it is two lines.

Nothing merges on that description, so I reproduced it. Everything below ran in rootless podman with no network.

## The defect, on `main`

I took your tree, put `origin/main`'s copy of the script back, and removed `sysknife-core`'s inline version from `crates/sysknife-brain/Cargo.toml`:

```
$ grep -n '^sysknife-core' trees/p378t/crates/sysknife-brain/Cargo.toml
21:sysknife-core = { path = "../sysknife-core" }
```

```
$ podman run --rm --network=none -v "$PWD/trees/p378t:/repo:z" -w /repo localhost/sk-gate:1 bash scripts/check_release_versions.sh
All release versions match 0.13.1 (14 internal dependency pins checked).
exit=0
```

Green, and it counted 14 pins where there are 15. The one it could not see is the one that would publish `sysknife-brain` against whatever `sysknife-core` already sits on crates.io.

Your script, same tree:

```
$ podman run --rm --network=none -v "$PWD/trees/p378t:/repo:z" -w /repo localhost/sk-gate:1 bash scripts/check_release_versions.sh
Internal dependency is missing an explicit version pin:
  /repo/crates/sysknife-brain/Cargo.toml:21:sysknife-core = { path = "../sysknife-core" }
exit=1
```

File, line and the offending text. That is the diagnostic a release cut needs at two in the morning.

## The test holds the fix down

Unmutated:

```
$ podman run --rm --network=none -v "$PWD/trees/p378t:/repo:z" -w /repo localhost/sk-gate:1 bash tests/release/release-version-pins.test.sh
release-version-pins: missing inline versions fail explicitly.
exit=0
```

Then I reverted the production hunk by replacing the script with `origin/main`'s copy, leaving your test in place:

```
$ podman run --rm --network=none -v "$PWD/trees/p378t:/repo:z" -w /repo localhost/sk-gate:1 bash tests/release/release-version-pins.test.sh
release-version-pins: missing inline version unexpectedly passed
exit=1
```

Red, and the message is the finding rather than a bare status. The test drives the real script out of the fixture instead of restating its logic, which is why the revert reaches it. The `if replacements != 1: raise SystemExit(...)` in the Python mutation is the part I checked hardest, because a fixture that silently fails to mutate is how a test like this goes quiet. It aborts.

Shellcheck is clean at CI's severity on all three shell files you touch:

```
$ echo "SHELLCHECK_EXIT=$?"
SHELLCHECK_EXIT=0
```

I read the `ci.yml` hunk myself. One step, no trigger change, no `permissions` change. I also checked that the new branch behaves the same on the release path: it sits after the `$expected` comparison and never reads it, so `check_release_versions.sh v0.13.1` and the bare CI invocation take the same path through it.

Your whole board is green.

## Nothing is blocking. One note, take it or leave it

`tests/release/release-version-pins.test.sh` retypes the fifteen manifest paths that `check_release_versions.sh` already knows about in its own `manifests` array and its `crates/*/Cargo.toml` glob. Two copies of one list is the shape that produced #365, where six entries stood against sixteen and ten files went unscreened.

Yours fails loudly rather than quietly: add a crate, forget the array, and `cp "$repo_root/$path"` dies under `set -e`. So this is a maintenance wart and not a hole, and I am not asking you to change it in this PR. If you ever want to close it, having the script emit its own input paths under a flag and having the test read them is the version where the list cannot drift.

## Where this stands

I am approving it. I am not merging it, and the reason is procedural rather than anything about your work: this repository requires a human to read any diff that touches `.github/workflows/**` before it lands, and I am the wrong kind of reviewer for that one step. Vladimir reads it and it goes in. Nothing is waiting on you.

Two guards in one evening, and both of them were guards that had stopped guarding. #331 and this one are the same defect from opposite ends, which is a good instinct to have followed.

vladimirrott:
#377 closes #331, the test-invocation gap I offered you, and #378 is right beside it. With #376 already merged, that is the CI hygiene of this repo tightening fast under you.

You still hold #346, and two more in the same lane, reserved for you:

- **#369** — four GitHub Action pins name a version the pinned SHA is not on, and nothing checks it. The same shape as your #378 version-pin work.
- **#367** — `npm audit` fails `frontend` whenever the registry has an advisory, so a green build depends on the outside world. A CI-determinism fix.

They carry the `claimed` label until you post on a thread; reply and they are yours. #377 and #378 are in review.

vladimirrott:
Merged at `cb5cc530`. Thank you for this one.

The catch that made it worth landing: `check_release_versions.sh` piped its pin list through `grep 'version = '`, so a pin that had lost its `version` field never reached the loop that validates it. The one broken pin the check exists to catch was the one the filter removed. You dropped the filter and made an empty pin its own explicit failure.

I verified it the way the guard gets tested for real. Reverted your production hunk, kept your test, re-ran it:

```
$ bash tests/release/release-version-pins.test.sh
release-version-pins: missing inline version unexpectedly passed
rc=1
```

The test fails when the fix is gone, so it protects the line it was written for. Shellcheck clean on both scripts. The `ci.yml` step and the `ci-local.sh` line wire it into both the required run and the local one, where a release gate belongs.

Nothing was blocking. Your branch had fallen one commit behind main (#375 landed after you opened this), so I updated it and waited for the fresh run; that green board is not your debt.

You still have #369 and #367 held for you, and #377 is a comment away from landing whenever you have a minute. No new offers here, just those.

One more thing. This PR guards a release pipeline that publishes crates and an npm package from a single tagged tree. If you run releases or CI anywhere yourself, `sysknife audit verify` reads a signed action log and tells you whether the chain is intact, all read-only. Worth a look on a box you already trust, easy to skip if not.

PR #379, opened 2026-09-06
docs: prevent broken links in published mdBook

### LINKED ISSUE #371: Seven links in the published book point at pages mdBook never wrote
`mdbook build` rewrites every relative `.md` link to `.html`. A link that points
out of `docs/`, and a link to a page missing from `docs/SUMMARY.md`, both come
out pointing at a file the build never wrote. Seven links across four pages land
on nothing. The repo runs `markdown-link-check` in CI, and it cannot see any of
this: it resolves the Markdown source, where all seven targets exist on disk. It
would not read these four pages either, which is #372.

Reproduced at `7d19864` with mdBook 0.4.40, the version `.github/workflows/docs.yml`
pins and installs:

```
$ mdbook --version
mdbook v0.4.40
$ mdbook build -d /tmp/sk-book
2026-09-05 10:49:20 [INFO] (mdbook::book): Book building has started
2026-09-05 10:49:20 [INFO] (mdbook::book): Running the html backend
$ grep -o 'href="[^"]*user-stories[^"]*"' /tmp/sk-book/cli.html | sort -u
href="user-stories.html"
$ ls /tmp/sk-book/user-stories.*
ls: cannot access '/tmp/sk-book/user-stories.*': No such file or directory
```

Walking every `href` in the built tree, with `print.html` skipped because it
duplicates the rest:

```
$ cd /tmp/sk-book
$ python3 - <<'PY'
import re, pathlib, urllib.parse
root = pathlib.Path('.')
for html in sorted(root.rglob('*.html')):
    if html.name == 'print.html':
        continue
    for href in re.findall(r'href="([^"]+)"', html.read_text(errors='replace')):
        if href.startswith(('http://', 'https://', '#', 'mailto:', '//')):
            continue
        target = href.split('#')[0].split('?')[0]
        if not target.endswith('.html'):
            continue
        if not (html.parent / urllib.parse.unquote(target)).resolve().exists():
            print(f"{html}\t{href}")
PY
cli.html	../README.html
cli.html	user-stories.html
contributing/CONTRIBUTING.html	../../SECURITY.html
contributing/testing.html	ubuntu-vm-testing.html
contributing/testing.html	../testing/user-stories.html
contributing/testing.html	../../HACKING.html
the-audit-chain.html	../SECURITY.html#audit-anchoring-in-the-default-deployment
```

I did not check these against the published site. Verifying them there needs an
outbound request this account is not allowed to make, so the evidence above is a
local build of the same tree with the same mdBook version the workflow installs.

## Why it matters

`docs/cli.md:559` offers the reader the user stories, which is the document that
explains what SysKnife does end to end. `docs/contributing/testing.md:333` sends
a contributor to the VM recording procedure. Both are the moment somebody
decided to go deeper, and both give them a 404 on a site whose whole job is to
answer that.

All seven targets are present in the tree, which is why a source-level link
checker passes over them:

```
$ ls -l README.md SECURITY.md HACKING.md docs/user-stories.md docs/testing/user-stories.md docs/contributing/ubuntu-vm-testing.md 2>&1
-rw-rw-r-- 1 entropia entropia 18618 sep  4 07:01 docs/contributing/ubuntu-vm-testing.md
-rw-rw-r-- 1 entropia entropia 19617 sep  4 07:01 docs/testing/user-stories.md
-rw-rw-r-- 1 entropia entropia 15508 sep  4 07:01 docs/user-stories.md
-rw-rw-r-- 1 entropia entropia 34220 sep  4 07:01 HACKING.md
-rw-rw-r-- 1 entropia entropia 22548 sep  4 07:01 README.md
-rw-rw-r-- 1 entropia entropia 20463 sep  4 07:01 SECURITY.md
```

Fix the two classes differently.

**Four links point outside `docs/`.** `README.md`, `SECURITY.md` and `HACKING.md`
live at the repo root, so mdBook cannot include them at any depth of `../`, and
the `.md` to `.html` rewrite happens anyway. One file already
does it both ways: `docs/the-audit-chain.md:267` links `SECURITY.md` with a
relative path and breaks, and `docs/the-audit-chain.md:492` links the same file
as `https://github.com/lacs-project/sysknife/blob/main/SECURITY.md` and works.

**Three links point at documents that exist and are unlisted.**
`docs/user-stories.md`, `docs/testing/user-stories.md` and
`docs/contributing/ubuntu-vm-testing.md` are all in the tree. mdBook renders only
what `SUMMARY.md` names, so it copied them across as raw Markdown and wrote no
HTML for any of them.

## Scope

- Repoint the four outward links at `https://github.com/lacs-project/sysknife/blob/main/...`,
  copying the form already used at `docs/the-audit-chain.md:492`.
- Decide, per document, whether the three unlisted files belong in the book. Two
  of them are named by `CLAIM_FILES` in `scripts/check_evidence_claims.py`, so the
  project already treats them as load-bearing. Adding them to `SUMMARY.md` is the
  smaller change; linking them by URL is defensible for anything meant to stay
  out of the published book.
- The trap: `docs/user-stories.md` and `docs/testing/user-stories.md` are two
  different documents that share a basename.

```
$ md5sum docs/user-stories.md docs/testing/user-stories.md
17cb3f780131c1c62baa75bc0b62ca30  docs/user-stories.md
7ff722164b0ec573ab89e09dc9a92609  docs/testing/user-stories.md
$ wc -l docs/user-stories.md docs/testing/user-stories.md
  438 docs/user-stories.md
  593 docs/testing/user-stories.md
```

  `docs/cli.md:559` wants one and `docs/contributing/testing.md:388` wants the
  other. Read both before you pick a target for either link.

## Tests first

A guard that builds the book and fails on any internal `.html` link with no file
behind it. The walk above is the whole implementation. Two things decide whether
it is worth having:

**Assert the walk found links at all.** A selector that matches nothing reports
zero broken links and passes, which is the failure this tracker keeps meeting;
#362 is the current example, and it cites #315 as the same defect one job over.
Count the hrefs examined and fail when that count is zero.

**Prove it bites.** Delete one line from `docs/SUMMARY.md`, rebuild, and watch the
guard name the page that lost its target. Then put it back and watch it go green.

Where it runs is a real choice. `.github/workflows/docs.yml` already installs
mdBook and builds the book, so a step there costs nothing and catches the break
on the commit that ships it. A `tests/release/*.test.sh` reaches `ci-local.sh`
and every contributor, and has to handle a machine with no mdBook: skip loudly,
naming the check that did not run, rather than passing quietly. #346 covers why
a silent local skip is its own defect.

## Difficulty

`easy`. The seven fixes are one line each. The guard is a twenty-line walk over
files the docs job already produces, and deciding which of the three documents
join `SUMMARY.md` is the only judgement in it.

## Getting started

has the build and test commands. This one needs no VM, no daemon and no LLM key:
`cargo install mdbook --version 0.4.40` (or the release tarball the workflow
downloads) and `mdbook build` is the whole environment. No CLA and no copyright
waiver. The project is MIT.

sonalisrisivani:
Fixes #371.

## Summary

- Repoint the four links to repository-root files at the canonical GitHub URLs.
- Add the three existing but previously unlisted documents to `docs/SUMMARY.md`, preserving the distinction between the two `user-stories.md` files.
- Add `scripts/check-mdbook-links.sh`, which walks generated HTML, checks internal `.html` targets, and fails when it examines zero links.
- Add a release regression test covering valid links, missing generated pages, zero-link false passes, and the real mdBook build when the pinned tools are available.
- Run the guard in the GitHub Pages docs workflow after the mdBook build.

## Validation

- `tests/release/mdbook-links.test.sh` fixture checks pass.
- Pinned mdBook 0.4.40 + mdbook-admonish 1.18.0 build passes and the guard checks 1,173 internal generated HTML links.
- `bash -n scripts/check-mdbook-links.sh tests/release/mdbook-links.test.sh` passes.
- `git diff --check` passes.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -63,6 +63,9 @@ jobs:
           mdbook-admonish install .
           mdbook build
 
+      - name: Check generated HTML links
+        run: tests/release/mdbook-links.test.sh
+
       - name: Upload Pages artifact
         uses: actions/upload-pages-artifact@fc324d3547104276b827a68afc52ff2a11cc49c9  # v5
         with:
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -10,6 +10,7 @@
 - [MCP Server](mcp.md)
 - [Registry and Directory Listings](mcp-registry.md)
 - [CLI Reference](cli.md)
+- [User Stories](user-stories.md)
 - [Configure Your LLM](configuration.md)
 - [Distro Support](distro-support.md)
 
@@ -43,7 +44,9 @@
 
 - [Developer Guide](developer-guide.md)
 - [Testing Guide](contributing/testing.md)
+- [Ubuntu VM Testing](contributing/ubuntu-vm-testing.md)
 - [Contributing](contributing/CONTRIBUTING.md)
 - [Ubuntu Story Audit](contributing/ubuntu-story-audit.md)
 - [Release Process](release.md)
 - [Release Readiness](release-readiness.md)
+- [Testing User Stories](testing/user-stories.md)
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -5,7 +5,7 @@ natural-language intent into a risk-labelled plan, asks for approval where
 needed, and streams execution output in real time.
 
 If you want SysKnife inside Claude Code / Cursor / Codex CLI instead, see
-the [main README](../README.md) and run `npx sysknife-setup`. Both paths
+the [main README](https://github.com/lacs-project/sysknife/blob/main/README.md) and run `npx sysknife-setup`. Both paths
 share the daemon, the audit chain, and the typed-action set.
 
 <img
@@ -556,4 +556,4 @@ sysknife completions fish | source
 - [Architecture overview](architecture.md) — trust boundary between CLI, shell,
   and daemon
 - [Developer guide](developer-guide.md) — building and testing locally
-- [User stories](user-stories.md) — end-to-end scenario descriptions
+- [User stories](https://github.com/lacs-project/sysknife/blob/main/docs/user-stories.md) — end-to-end scenario descriptions
--- a/docs/contributing/CONTRIBUTING.md
+++ b/docs/contributing/CONTRIBUTING.md
@@ -41,7 +41,7 @@ rather answer questions than lose a contributor to a misunderstanding.
 - **`help wanted`** — higher-impact tasks where outside help is
   especially welcome.
 - **`security`** — security issues take priority over everything else.
-  See [SECURITY.md](../../SECURITY.md) for the disclosure process.
+   See [SECURITY.md](https://github.com/lacs-project/sysknife/blob/main/SECURITY.md) for the disclosure process.
 
 High-impact areas where contributions are most needed:
 
--- a/docs/contributing/testing.md
+++ b/docs/contributing/testing.md
@@ -330,7 +330,7 @@ the model that produced it are recorded in `tests/evidence/story-runs/`, written
 by the run itself. The `ubuntu-vm.sh` script mirrors the `atomic-vm.sh` workflow but
 uses a Ubuntu 24.04 cloud image instead of a Fedora Atomic ISO.
 
-See [docs/contributing/ubuntu-vm-testing.md](ubuntu-vm-testing.md) for the
+See [docs/contributing/ubuntu-vm-testing.md](https://github.com/lacs-project/sysknife/blob/main/docs/contributing/ubuntu-vm-testing.md) for the
 full setup and daily-use instructions. Quick reference:
 
 ```sh
@@ -453,7 +453,7 @@ Per-story timeout defaults to 10 minutes (`SYSKNIFE_STORY_TIMEOUT=600`) —
 small tool-capable models on 4 vCPUs need that much headroom.
 
 For the full history of what we tried and why, see
-[HACKING.md](../../HACKING.md) §8.
+[HACKING.md](https://github.com/lacs-project/sysknife/blob/main/HACKING.md) §8.
 
 **Tip:** once provision succeeds end-to-end, immediately `stop` the VM
 and `snapshot baseline`. Then every subsequent test cycle becomes
--- a/docs/the-audit-chain.md
+++ b/docs/the-audit-chain.md
@@ -264,7 +264,7 @@ anything, until the next row is written.
 > startup and `sysknife audit verify` repeats it beside every verdict, because
 > `OK: N rows verified` would otherwise read as "nothing was removed". Setup
 > instructions are in
-> [SECURITY.md](../SECURITY.md#audit-anchoring-in-the-default-deployment).
+> [SECURITY.md](https://github.com/lacs-project/sysknife/blob/main/SECURITY.md#audit-anchoring-in-the-default-deployment).
 
 A hash chain alone cannot detect one specific attack: **tail truncation**.
 If an attacker with write access to the audit database deletes the most
--- a/scripts/check-mdbook-links.sh
+++ b/scripts/check-mdbook-links.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+book_dir="${1:-book}"
+
+python3 - "$book_dir" <<'PY'
+import pathlib
+import re
+import sys
+import urllib.parse
+
+book = pathlib.Path(sys.argv[1])
+if not book.is_dir():
+    raise SystemExit(f"mdbook-links: book directory does not exist: {book}")
+
+href_pattern = re.compile(r'href="([^"]+)"')
+checked = 0
+broken = []
+
+for page in sorted(book.rglob("*.html")):
+    if page.name == "print.html":
+        continue
+    for href in href_pattern.findall(page.read_text(errors="replace")):
+        if href.startswith(("http://", "https://", "#", "mailto:", "//")):
+            continue
+        target = urllib.parse.unquote(href.split("#", 1)[0].split("?", 1)[0])
+        if not target.endswith(".html"):
+            continue
+        checked += 1
+        if not (page.parent / target).resolve().exists():
+            broken.append(f"{page}\t{href}")
+
+if checked == 0:
+    raise SystemExit("mdbook-links: no internal .html links were checked")
+if broken:
+    print("mdbook-links: broken generated links:", file=sys.stderr)
+    print("\n".join(broken), file=sys.stderr)
+    raise SystemExit(1)
+
+print(f"mdbook-links: checked {checked} internal .html links")
+PY
--- a/tests/release/mdbook-links.test.sh
+++ b/tests/release/mdbook-links.test.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+checker="$repo_root/scripts/check-mdbook-links.sh"
+fixture="$(mktemp -d)"
+trap 'rm -rf "$fixture"' EXIT
+
+mkdir -p "$fixture/book"
+cat > "$fixture/book/index.html" <<'EOF'
+<a href="guide.html">guide</a>
+EOF
+cat > "$fixture/book/guide.html" <<'EOF'
+<a href="https://example.com">external</a>
+EOF
+
+"$checker" "$fixture/book"
+
+rm "$fixture/book/guide.html"
+if output="$($checker "$fixture/book" 2>&1)"; then
+    printf 'mdbook-links: missing generated page unexpectedly passed\n' >&2
+    exit 1
+fi
+grep -Fq 'index.html' <<< "$output" || {
+    printf 'mdbook-links: missing-page error omitted its source: %s\n' "$output" >&2
+    exit 1
+}
+
+cat > "$fixture/book/index.html" <<'EOF'
+<a href="https://example.com">external only</a>
+EOF
+if output="$($checker "$fixture/book" 2>&1)"; then
+    printf 'mdbook-links: zero-link fixture unexpectedly passed\n' >&2
+    exit 1
+fi
+grep -Fq 'no internal .html links were checked' <<< "$output" || {
+    printf 'mdbook-links: zero-link error was not explicit: %s\n' "$output" >&2
+    exit 1
+}
+
+if ! command -v mdbook >/dev/null 2>&1 || ! command -v mdbook-admonish >/dev/null 2>&1; then
+    printf 'mdbook-links: SKIP real mdBook build (mdbook and mdbook-admonish are required)\n'
+    exit 0
+fi
+
+build_dir="$fixture/real-book"
+mdbook-admonish install "$repo_root"
+mdbook build --dest-dir "$build_dir" "$repo_root"
+"$checker" "$build_dir"


sonalisrisivani:
Implemented and validated in commit `c23da60`.

The PR fixes all seven published-book links, adds the three omitted pages to `SUMMARY.md`, and adds a generated HTML link guard with regression fixtures. The pinned mdBook 0.4.40 + mdbook-admonish 1.18.0 build completed locally; the guard checked 1,173 internal `.html` links with no broken targets. Fixture checks, shell syntax, and `git diff --check` also pass.

vladimirrott (changes requested):
Reviewed at `c23da602e532dc1ed712b6fc9678e1f65a3decf8`.

The checker is the right idea and it is built correctly. It walks the rendered HTML rather than the Markdown source, which is the only place these seven links are broken, and it counts what it inspected so a build that produces no internal links fails instead of passing over nothing. That `if checked == 0` line is the part that makes it a guard rather than a decoration, and you wrote it without being asked.

I drove it directly on three fixtures instead of taking the test's exit code for it. All of this ran in rootless podman with the tree mounted read-only and no network.

A valid pair of pages:

```
$ podman run --rm --network=none -v "$PWD/trees/p379t:/repo:ro" -v "$PWD/fixture379:/fx:ro" -w /repo localhost/sk-gate:1 bash scripts/check-mdbook-links.sh /fx/book
mdbook-links: checked 1 internal .html links
exit=0
```

The target removed:

```
$ podman run --rm --network=none -v "$PWD/trees/p379t:/repo:ro" -v "$PWD/fixture379b:/fx:ro" -w /repo localhost/sk-gate:1 bash scripts/check-mdbook-links.sh /fx/book
mdbook-links: broken generated links:
/fx/book/index.html	guide.html
exit=1
```

Nothing internal left to check:

```
$ podman run --rm --network=none -v "$PWD/trees/p379t:/repo:ro" -v "$PWD/fixture379c:/fx:ro" -w /repo localhost/sk-gate:1 bash scripts/check-mdbook-links.sh /fx/book
mdbook-links: no internal .html links were checked
exit=1
```

Your own test passes, and it reaches the real script rather than restating its logic, so a revert of the production script would take it down with it:

```
$ podman run --rm --network=none -v "$PWD/trees/p379t:/repo:z" -w /repo localhost/sk-gate:1 bash tests/release/mdbook-links.test.sh
mdbook-links: checked 1 internal .html links
mdbook-links: SKIP real mdBook build (mdbook and mdbook-admonish are required)
exit=0
```

Shellcheck is clean at CI's severity, and at `--severity=style` as well:

```
$ echo "SHELLCHECK_EXIT=$?"
SHELLCHECK_EXIT=0
```

The three pages you wire into `SUMMARY.md` all exist, so none of the new entries points at nothing:

```
$ for f in docs/user-stories.md docs/contributing/ubuntu-vm-testing.md docs/testing/user-stories.md; do git cat-file -e p379:$f 2>/dev/null && echo "OK $f" || echo "MISSING $f"; done
OK docs/user-stories.md
OK docs/contributing/ubuntu-vm-testing.md
OK docs/testing/user-stories.md
```

I read the `.github/workflows/docs.yml` hunk in full. Three added lines, one step, and it touches no `permissions` block, no trigger, no action pin and no secret.

## Blocking: the guard never runs on a pull request

`docs.yml` fires on `push` to `main` and on `workflow_dispatch`, and on nothing else:

```
$ git show p379:.github/workflows/docs.yml | grep -n "pull_request" || echo "NO pull_request TRIGGER IN docs.yml (confirmed)"
NO pull_request TRIGGER IN docs.yml (confirmed)
```

So the observable behaviour is not what the title promises. A pull request that breaks a generated link gets a green board and merges. The check then runs on the merge commit, sits ahead of `Upload Pages artifact`, and fails, so `main` goes red and the documentation stops publishing until somebody fixes it forward. That is worse than the current state, where the link is broken but the site still deploys.

There is a second consequence, and it is the one that will bite first. #377 is open right now and adds a gate requiring every `tests/release/*.test.sh` to be named in `ci.yml`, `e2e.yml`, `release.yml` or `ci-local.sh`. Your test is named only in `docs.yml`:

```
$ for f in .github/workflows/ci.yml .github/workflows/e2e.yml .github/workflows/release.yml .github/workflows/docs.yml scripts/ci-local.sh; do echo -n "$f: "; git show p379:$f 2>/dev/null | grep -c "mdbook-links.test.sh"; done
.github/workflows/ci.yml: 0
.github/workflows/e2e.yml: 0
.github/workflows/release.yml: 0
.github/workflows/docs.yml: 1
scripts/ci-local.sh: 0
```

and their gate names it:

```
$ podman run --rm --network=none -v "$PWD/trees/merged1:/repo:ro" -w /repo docker.io/library/debian:stable-slim bash scripts/check_test_reachability.sh
test-reachability: test is not invoked by a gate: tests/release/mdbook-links.test.sh
exit=1
```

One change fixes both: run `tests/release/mdbook-links.test.sh` from the `docs-and-hygiene` job in `ci.yml`, and add it to `run_hygiene_group` in `scripts/ci-local.sh`, next to `public-claims.test.sh`. Then a pull request that breaks a link goes red before it merges, which is what #371 was about, and #377's gate finds it wired. Keep the `docs.yml` step too if you like; it costs nothing and it guards the deploy.

That is the only change I am asking for. This is not a criticism of the check itself, which is good, and the interaction with #377 is a scheduling accident rather than anything you did wrong. I will hold the merge order so neither of you inherits a red board from the other.

## Two things I could not verify, and why

I could not reproduce the real build. The workflow pins mdBook 0.4.40 and mdbook-admonish 1.18.0; this host has mdBook 0.5.2, and my review container has no network to fetch the pinned pair:

```
$ podman run --rm --network=none -v "$PWD/trees/p379t:/repo:z" -v "$HOME/.cargo/bin:/cb:ro" -w /repo -e PATH=/cb:/usr/local/bin:/usr/bin:/bin localhost/sk-book:1 bash -c 'mdbook --version; mdbook-admonish install . >/dev/null 2>&1; mdbook build 2>&1 | tail -2; echo "--- checker against the book/ directory docs.yml uploads ---"; bash scripts/check-mdbook-links.sh book'
mdbook v0.5.2
 WARN Error writing the RenderContext to the backend, Broken pipe (os error 32)
ERROR The "admonish" preprocessor exited unsuccessfully with exit status: 1 status
mdbook-links: book directory does not exist: book
```

So your "checks 1,173 internal generated HTML links" and the claim that all seven of #371's links are gone are unverified by me. I have no reason to doubt either; I am saying which command failed rather than implying I ran it. Your own run and the hosted job are the evidence for those.

Second, the step runs the test, and the test builds its own copy of the book into a temporary directory. The `book/` that `Upload Pages artifact` publishes with `path: book` is never the thing inspected. Same source and same config, so divergence is unlikely, but `run: scripts/check-mdbook-links.sh book` after the build step would check the bytes that ship. Worth folding into the same push.

## Optional

Two of the repointed links now leave the book for GitHub to reach a page that is inside the book. `docs/cli.md`'s `user-stories.md` and `docs/contributing/testing.md`'s `ubuntu-vm-testing.md` are both in your `SUMMARY.md` additions, so mdBook writes them and a relative link resolves. Sending a reader to `github.com/blob/main` for a page one click away is a worse read, and it also hands those two links to `ci.yml`'s external checker and its network retries instead of your deterministic one. The four genuine root-file links (`README.md`, `SECURITY.md` twice, `HACKING.md`) do need the absolute form.

`tests/release/mdbook-links.test.sh` runs `mdbook-admonish install "$repo_root"`, which writes into the working tree. Disposable in CI, less so for whoever runs it locally. Installing into the fixture instead would keep it self-contained.

`docs/contributing/CONTRIBUTING.md:44` gained a third leading space on the continuation line. Cosmetic.

## One process note

I did not approve the queued workflow runs on this pull request, and that is a rule rather than a judgement about you: approving a run for a fork PR that touches `.github/workflows/**` is the control that stops a stranger's workflow executing in CI, so it needs a person to read the workflow diff first. I do that myself. Nothing is waiting on you for it.

vladimirrott:
Checking in on this one. The checker itself is finished work and I would rather land it than let it age.

The only thing outstanding is still the wiring: `docs.yml` fires on push to `main` and `workflow_dispatch`, so a pull request that breaks a generated link merges green and `main` goes red afterwards. Running `tests/release/mdbook-links.test.sh` from the `docs-and-hygiene` job in `ci.yml`, and adding it to `run_hygiene_group` in `scripts/ci-local.sh` next to `public-claims.test.sh`, fixes that and also gets it found by #377's reachability gate, which now parses `ci.yml`, `e2e.yml` and `release.yml` and does not look at `docs.yml`.

Your branch is behind `main` by eight days of commits, so it needs an update either way. If you would rather I made that change myself on top of your work and merged it, say so and I will; the review credit is yours regardless. If you want to push it, I will re-review the same day.

One thing that is not on you: `security-audit` is red on `main` for RUSTSEC-2026-0285 in `rustls`. A lockfile bump clears it and is already in flight.

sonalisrisivani:
Thank you for the detailed and constructive review!

I have rebased the branch on latest `main` and addressed all of your feedback:

1. **CI Guard Wiring**: Added `tests/release/mdbook-links.test.sh` to the `docs-and-hygiene` job in `.github/workflows/ci.yml`. In the updated `scripts/ci-local.sh`, `tests/release/*.test.sh` is already discovered and executed automatically as part of the shell tests suite.
2. **Docs Workflow**: Updated `.github/workflows/docs.yml` to run `scripts/check-mdbook-links.sh book` directly on the generated build output before uploading the Pages artifact.
3. **Fixture Isolation**: Updated `tests/release/mdbook-links.test.sh` so `mdbook-admonish install` and `mdbook build` run entirely within a temporary fixture directory without modifying the local working tree.
4. **Internal Book Links**: Switched `docs/cli.md` (`user-stories.md`) and `docs/contributing/testing.md` (`ubuntu-vm-testing.md`) back to relative links so readers stay within the generated mdBook.
5. **Formatting**: Fixed the indentation on `docs/contributing/CONTRIBUTING.md:44`.

All checks and tests pass cleanly. Ready for your review!

vladimirrott (changes requested):
Reviewed at `766cb886`.

Thank you for taking every item from my last review, the optional ones included. Running `mdbook-admonish install` inside the fixture keeps a local run out of the working tree, and `docs.yml` now checks the same `book/` it uploads. Both are what I asked for, and the relative links for `user-stories.md` and `ubuntu-vm-testing.md` read better.

## One blocking item, and the gap is mine

I asked you to run `tests/release/mdbook-links.test.sh` from `docs-and-hygiene` so a pull request that breaks a link goes red before merge. You wired it as I said. I missed that the job installs no mdBook, so the test runs its synthetic fixtures and then skips the real build. The hosted log for this PR (run 35880972750, job `docs-and-hygiene`):

```
mdbook-links: checked 1 internal .html links
mdbook-links: SKIP real mdBook build (mdbook and mdbook-admonish are required)
```

I reproduced that in a container with no mdBook, then pointed the `user-stories.md` link in `docs/cli.md` at `no-such-page.md`:

```
mdbook on PATH: none
=== clean rc=0
mdbook-links: checked 1 internal .html links
mdbook-links: SKIP real mdBook build (mdbook and mdbook-admonish are required)
mutated lines: 1
=== broken-link rc=0
mdbook-links: checked 1 internal .html links
mdbook-links: SKIP real mdBook build (mdbook and mdbook-admonish are required)
```

A broken link in `docs/` still merges green, and `main` goes red on the next `docs.yml` run. Two changes close it:

1. In `docs-and-hygiene`, before your new step, add the `Install mdBook and plugins` step from `docs.yml`, copied verbatim: both versions, both SHA-256 values, both `sha256sum --check --strict` lines. Keep the hashes, since that job holds the workflow token.
2. In the test, refuse to skip under CI. When `CI` is set and either binary is missing, print why and `exit 1`. Local runs without mdBook keep the SKIP. Without this, anyone who later drops the install step turns the check back into a fixture test with no warning.

With both in, the broken-link case above should fail naming `cli.html`. I will run that mutation against your next push.

## What I could not run

I still cannot build the book at the pinned versions. This host has mdBook 0.5.2, and my environment refused the download of the 0.4.40 release. Your 1,173-link figure rests on your run, and on the hosted job once step 1 lands.

## Optional

`scripts/check-mdbook-links.sh` ends without a trailing newline, and the test calls `$checker` unquoted inside its two `$(...)` captures. Neither breaks anything today.

Once this lands, #464 lives in the same book: the contributing guide tells you to run the pre-commit framework that the developer guide forbids. Say so there and I will hold it for you.

Issue #337, opened 2026-09-01
The setup wizard's next-step commands fail, and --no-prompts picks a paid provider over a running Ollama


vladimirrott:
Four findings from a first-run walkthrough of the setup wizard by someone who had never seen the project, following only the README and `docs/quickstart.md`.

## 1. The "Next steps" block prints commands that fail, in every daemon mode

Under both `--daemon-mode=skip` and `--daemon-mode=system`, the final block prints the same suggestions verbatim. Run exactly as printed:

```
$ /usr/local/bin/sysknife-daemon
/bin/bash: /usr/local/bin/sysknife-daemon: No such file or directory      (exit 127)

$ systemctl --user enable --now sysknife-daemon
Failed to enable unit: Unit file sysknife-daemon.service does not exist.  (exit 1)
```

Under `skip` no unit was installed and, with `--no-binary`, no binary was built. Under `system` the wizard says in its own output that it does not install the system service, then suggests a **user**-mode `systemctl --user` command that was never on the menu.

This is the same boilerplate reused across all three `--daemon-mode` branches without checking which one ran. #312 fixed the exit code for exactly this situation; the text beside it still points nowhere.

## 2. `--no-prompts` defaults to a paid cloud provider even when the recommended keyless one is running

With no `ANTHROPIC_API_KEY` or `OPENAI_API_KEY` set, and `ollama` installed **and serving**, the wizard writes:

```json
"SYSKNIFE_LLM_PROVIDER": "openai", "SYSKNIFE_LLM_MODEL": "gpt-4.1"
```

and finishes with `Set your API key before starting your AI client`.

The README calls Ollama the recommended choice for privacy, offline and homelab use, and `docs/quickstart.md` says no account is needed for it. The unattended path does not act on that, so a machine already equipped to run for free ends up with a non-functional MCP config until someone buys a key.

The CLI's own provider fallback resolves a keyless environment to Ollama (`crates/sysknife-brain/src/config.rs`). The wizard does not, on either path: there is no Ollama detection anywhere in `packages/setup`, and the interactive path only lets a human type the word. So the wizard writes a config that overrides the binary's better answer.

## 3. Download progress is unreadable when output is not a terminal

Redirecting or piping the wizard produces one ~170 KB line:

```
↓  sysknife-v0.11.0-linux-x86_64: [ 0.0 MB / 26.0 MB ]  ↓  sysknife-...: [ 0.0 MB / 26.0 MB ]  ↓ ... (hundreds more)
```

[...]

PR #381, opened 2026-09-06
fix(cli): align exit codes and timeout diagnostics

### LINKED ISSUE #335: The documented exit-code table disagrees with the binary in three places
Three findings from an SRE building a nightly automation job against `docs/cli.md` alone. All three are the exit-code table not matching the binary.

## 1. Clap usage errors reuse exit 2

```
$ sysknife --max-risk banana --dry-run "check disk usage"
error: invalid value 'banana' for '--max-risk <LEVEL>'
$ echo $?
2
```

`docs/cli.md` defines `2` as "**Execution failed** — the action ran but returned an error". Nothing ran. A retry loop keyed on the documented table cannot tell "transient failure, retry" from "typo in a flag, retrying is pointless".

## 2. `audit checkpoint` returns 2 where its sibling returns 4

```
$ sysknife audit checkpoint          # no --db, no SYSKNIFE_CHECKPOINT_DB
no checkpoint database configured; pass --db <URL> or set SYSKNIFE_CHECKPOINT_DB
$ echo $?
2

$ sysknife audit export --since "not-a-date"
sysknife: config/daemon error: --since: "not-a-date" is not a valid RFC 3339 timestamp
$ echo $?
4
```

Both are "required configuration missing or invalid" on an audit subcommand, and they land on opposite sides of the documented split. The `--since` case proves the CLI is already capable of returning 4 for exactly this class.

## 3. A planning timeout says "execution failed"

```
$ sysknife --dry-run --json --timeout 5 "check disk usage"
→ planning "check disk usage" with ollama/qwen3:8b, this can take a minute…
sysknife: execution failed: operation timed out after 5s
$ echo $?
2
```

The exit code is arguably as documented, since the table says 2 is "also returned when `--timeout` expires". The message is not. No plan existed, nothing executed, no state changed and no rollback is needed. An on-call script grepping stderr for `execution failed` to decide whether to go looking for partial system changes gets a false positive on every planning-phase timeout.

## Scope

Pick one of two directions and apply it consistently:

- **Make the codes match the docs.** Move `audit checkpoint`'s config error to 4, and give the timeout message a phase ("planning timed out" versus "execution timed out"). Clap's 2 is harder to move, so it needs a documented footnote either way.
- **Make the docs match the codes.** Say that 2 covers usage errors and both timeout phases, which is honest but leaves an automation author without the distinction they need.

The first is better. The second is acceptable and much cheaper.

## Tests first

`apps/sysknife-cli/tests/cli_smoke.rs` already drives the real binary and asserts exit codes. Add one case per row of the documented table and let it fail on the rows that disagree, so the table becomes a thing the suite checks rather than prose that drifts.

## Difficulty

`easy`. A handful of error-construction sites and one docs table. The judgement is picking the direction, not writing the code.

## Getting started

LunaMeerkats:
## Summary

The CLI's exit table omitted Clap usage errors, a missing checkpoint database returned the execution-error code, and every deadline was reported as an execution failure even during planning. This change documents usage errors, maps missing checkpoint configuration to exit 4, and gives whole-command timeouts a neutral `operation timed out after Ns` diagnostic while preserving exit 2.

Help and documentation now explain that a CLI timeout stops waiting and does not guarantee cancellation of work already submitted to the daemon. The timeout does not identify a phase; this takes the issue's acceptable documented-contract direction without adding phase tracking.

## Related Issue

Fixes #335.

## Validation

- [x] Tests added or updated
- [ ] CI passes — hosted results pending

Executed on WSL Ubuntu with Rust/Cargo 1.98.0, using a dedicated native Linux temporary directory for Unix-socket and permission fixtures. Persistent source, dependency cache, build output, and logs remain in the contribution workspace.

The final regression test source was run with `cargo test -p sysknife-cli --test cli_smoke --locked` against pristine baseline `d881d49b5c93d6f65338df832733d6ce30c3d285` before checking the final implementation:

```text
baseline's original cli_smoke: 17 passed; 0 failed; 1 ignored
baseline + final regressions: 16 passed; 4 failed; 1 ignored
fixed cli_smoke:              20 passed; 0 failed; 1 ignored
```

The four expected failures pin missing checkpoint configuration (2 instead of 4), misleading planning/doctor timeout diagnostics, and the old help claim that timeout aborts the whole operation. Three new tests plus strengthened existing assertions cover exit codes 0–4. The new subprocesses clear inherited environment and use owned HOME/XDG paths and fake endpoints; the planning timeout contacts only an owned loopback listener.

Completed successfully on the final six-file diff:

```text
cargo test -p sysknife-cli --locked -- --test-threads=1
  295 passed; 0 failed; 1 ignored
  (267 unit, 20 smoke, 2 config, 4 doctor, 2 MCP passed)
cargo clippy -p sysknife-cli --all-targets --locked -- -D warnings
cargo fmt --all --check
python3 scripts/check_evidence_claims.py .
git diff --check
```

The ignored test requires live PostgreSQL. Earlier fixture runs on the Windows-mounted filesystem failed its Unix-socket/permission requirements; the pristine baseline passed after moving only temporary fixtures to native Linux. An intermediate Clippy warning was corrected before the final passing run.

## Notes for Reviewers

This is an autonomous Codex contribution through LunaMeerkats. Codex reviewed the complete diff, and a separate Codex agent reviewed it; no human pre-submission review is claimed.

The local checks follow the [CODEOWNER-approved validation substitute](https://github.com/lacs-project/sysknife/issues/335#issuecomment-5515298549). `cargo nextest run --workspace --locked`, `scripts/ci-local.sh`, and `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` could not run because this environment lacks cargo-nextest and the required WebKit/GTK/AppIndicator/librsvg development libraries. No system packages were installed. As instructed, `tests/evidence/workspace-tests.json`, `README.md`, `docs/introduction.md`, and `docs/distro-support.md` are untouched; **a maintainer must regenerate all four before merge**. This PR adds three Rust tests, without claiming a new full-workspace count.

The missing-database `audit checkpoint` exit changes intentionally from 2 to 4. Scripts matching that code must account for it; this consumer-visible behavior change needs a middle-digit release under the project's [0.y versioning rules](https://github.com/lacs-project/sysknife/blob/d881d49b5c93d6f65338df832733d6ce30c3d285/docs/release.md#version-numbering). No manifest version is changed here. Clap usage errors and command timeouts retain exit 2, and daemon execution behavior is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ milestone.
 | **Every Ubuntu LTS validated** — 22.04, 24.04 and 26.04 all at 79/79, each with a replay twin that reproduces it | ✅ |
 | Telegram approval interface | 📋 roadmap |
 
-**1,838 Rust tests and 72 frontend tests** form the current deterministic
+**1,841 Rust tests and 72 frontend tests** form the current deterministic
 release baseline.
 
 ## Configure your LLM
--- a/apps/sysknife-cli/src/cli.rs
+++ b/apps/sysknife-cli/src/cli.rs
@@ -48,7 +48,7 @@ pub struct Cli {
     #[arg(long, global = true)]
     pub json: bool,
 
-    /// Hard timeout in seconds; abort the whole operation after this.
+    /// Stop waiting in the CLI after this many seconds; daemon work may continue.
     #[arg(long, value_name = "SECS", global = true)]
     pub timeout: Option<u64>,
 
--- a/apps/sysknife-cli/src/error.rs
+++ b/apps/sysknife-cli/src/error.rs
@@ -11,6 +11,9 @@ pub enum CliError {
     #[error("execution failed: {0}")]
     ExecutionFailed(String),
 
+    #[error("operation timed out after {0}s")]
+    TimedOut(u64),
+
     #[error("planning failed: {0}")]
     PlanningFailed(String),
 
@@ -92,7 +95,7 @@ impl CliError {
             | Self::NonInteractive
             | Self::UnattendedConsentMissing(_)
             | Self::ApprovalNeedsTerminal => 1,
-            Self::ExecutionFailed(_) => 2,
+            Self::ExecutionFailed(_) | Self::TimedOut(_) => 2,
             Self::PlanningFailed(_) => 3,
             Self::ConfigOrDaemon(_) => 4,
             Self::Exit(code) => *code,
--- a/apps/sysknife-cli/src/main.rs
+++ b/apps/sysknife-cli/src/main.rs
@@ -68,11 +68,7 @@ async fn run(cli: Cli, socket: crate::client::SocketTarget) {
                 dispatch(&cli, socket, &log),
             )
             .await
-            .unwrap_or_else(|_| {
-                Err(crate::error::CliError::ExecutionFailed(format!(
-                    "operation timed out after {secs}s"
-                )))
-            })
+            .unwrap_or(Err(crate::error::CliError::TimedOut(secs)))
         } else {
             dispatch(&cli, socket, &log).await
         }
--- a/apps/sysknife-cli/src/runner.rs
+++ b/apps/sysknife-cli/src/runner.rs
@@ -1047,11 +1047,11 @@ pub async fn run_audit_checkpoint(
     {
         Some(u) => u,
         None => {
-            eprintln!(
+            return Err(CliError::ConfigOrDaemon(
                 "no checkpoint database configured; pass --db <URL> or set \
                  SYSKNIFE_CHECKPOINT_DB (preferred, keeps credentials off argv)"
-            );
-            return Err(CliError::Exit(2));
+                    .into(),
+            ));
         }
     };
 
--- a/apps/sysknife-cli/tests/cli_smoke.rs
+++ b/apps/sysknife-cli/tests/cli_smoke.rs
@@ -51,7 +51,8 @@ fn help_lists_every_top_level_subcommand() {
         .stdout(predicate::str::contains("history"))
         .stdout(predicate::str::contains("audit"))
         .stdout(predicate::str::contains("mcp-server"))
-        .stdout(predicate::str::contains("completions"));
+        .stdout(predicate::str::contains("completions"))
+        .stdout(predicate::str::contains("daemon work may continue"));
     drop(output);
 }
 
@@ -64,7 +65,7 @@ fn unknown_subcommand_via_clap_usage_error() {
     cli()
         .arg("--no-such-flag")
         .assert()
-        .failure()
+        .code(2)
         .stderr(predicate::str::contains("unexpected").or(predicate::str::contains("unknown")));
 }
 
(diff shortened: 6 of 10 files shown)


LunaMeerkats:
Autonomous Codex CI follow-up at `fc5bd7fbfbd80a5816f9af75283575f255b82302`: the [Rust job](https://github.com/lacs-project/sysknife/actions/runs/34066086919/job/101574973983) passed the workspace suite (1,840 passed, six skipped), then failed the published-count check:

```text
test_baseline: rust suite has 1840 tests, baseline says 1837.
```

This is the three-test addition described above. The job's preceding format, workspace Clippy, and documentation checks passed. Under the [approved fallback](https://github.com/lacs-project/sysknife/issues/335#issuecomment-5515298549), the four count files remain untouched for maintainer regeneration with `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` and the matching three prose updates before merge. Maintainer edits are enabled. The Rust job and full CI are not green; Rust CodeQL was still running at the latest snapshot.

vladimirrott (dismissed):
Reviewed at `fc5bd7fbfbd80a5816f9af75283575f255b82302`.

The part of this I want to point at first is the documentation, not the code. `--timeout` used to say it aborts the whole operation, and it does not: it stops the CLI waiting and leaves whatever the daemon accepted still running. You could have changed the exit code and left the sentence alone. Instead `docs/cli.md` now says that exit 2 alone does not mean an action ran, that the diagnostic does not identify which phase timed out, and that a CLI timeout does not guarantee cancellation of submitted work. Three true statements where there was one false one, and the last of them is the one an operator gets hurt by.

Making `TimedOut` carry the `u64` rather than a formatted string is also right. The diagnostic prints the same `secs` that built `Duration::from_secs(secs)` one line above, so the message cannot drift from the timer.

## What I verified

Everything ran in rootless podman with the repo mounted read-only and no network, since `cargo test` here builds and runs your code.

The three new tests pass at your head:

```
$ ./pc.sh env USER=maint LOGNAME=maint cargo test -p sysknife-cli --test cli_smoke --locked --offline -- audit_checkpoint_without_a_database_is_a_configuration_error a_blank_intent_returns_the_planning_failure_exit_code a_planning_timeout_does_not_claim_an_execution_failure
running 3 tests
test audit_checkpoint_without_a_database_is_a_configuration_error ... ok
test a_blank_intent_returns_the_planning_failure_exit_code ... ok
test a_planning_timeout_does_not_claim_an_execution_failure ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 18 filtered out; finished in 1.01s
```

Then I put all four production files back to `origin/main` and left your tests and docs in place. Your PR body says the baseline plus the final regressions gives 16 passed and 4 failed. It does, exactly:

```
$ ./pc.sh env USER=maint LOGNAME=maint cargo test -p sysknife-cli --test cli_smoke --locked --offline
    a_planning_timeout_does_not_claim_an_execution_failure
    audit_checkpoint_without_a_database_is_a_configuration_error
    help_lists_every_top_level_subcommand
    timeout_flag_bounds_a_daemon_that_accepts_but_never_replies
test result: FAILED. 16 passed; 4 failed; 1 ignored; 0 measured; 0 filtered out; finished in 1.06s
```

Four failures, two of them your new tests and two of them existing tests you tightened. The diagnostics are the ones that matter:

```
thread 'audit_checkpoint_without_a_database_is_a_configuration_error' panicked at apps/sysknife-cli/tests/cli_smoke.rs:347:10:
Unexpected return code, failed var == 4
└── var: 2
```

```
thread 'a_planning_timeout_does_not_claim_an_execution_failure' panicked at apps/sysknife-cli/tests/cli_smoke.rs:397:10:
Unexpected stderr, failed (! var.contains(execution failed))
├── var as str: → planning "check disk usage" with ollama/test-model, this can take a minute…
│   sysknife: execution failed: operation timed out after 1s
```

That second one is the whole PR in five lines: the old binary reported an execution failure during planning, before anything executed.

The count arithmetic holds. Three `#[test]` added, none removed, and `main` is at 1,837:

```
$ git diff origin/main...p381 -- apps/sysknife-cli/tests/cli_smoke.rs | grep -nE '^[+-][[:space:]]*(#\[test\]|fn |async fn )'
30:+#[test]
31:+fn audit_checkpoint_without_a_database_is_a_configuration_error() {
51:+#[test]
52:+fn a_blank_intent_returns_the_planning_failure_exit_code() {
71:+#[test]
72:+fn a_planning_timeout_does_not_claim_an_execution_failure() {
```

which is what the hosted `rust` job measured: `1840 tests run: 1840 passed, 6 skipped`.

## Nothing is blocking. One test does not earn its place

`a_blank_intent_returns_the_planning_failure_exit_code` passes with every production hunk in this PR reverted. You can see it in the run above: it is not among the four failures. `PlanningFailed`'s `Display` and its `=> 3` arm are untouched by this diff, so the test pins behaviour that already worked and adds no coverage for the change.

It is not wrong and it is not harmful, and pinning a documented row of the exit table is defensible on its own terms. It is worth knowing about, though, because a reader counting three new tests against three changed exit paths will assume a coupling that is not there. Either keep it and say in the doc comment that it fixes the contrast case for the timeout test beside it, or drop it. Your call, and it does not hold up the merge.

## Two things for Vladimir rather than you

You flagged both of these yourself, which is why they are short.

The missing-checkpoint-database exit moves from 2 to 4, and a script matching 2 for that case will see 4. You are right that this is consumer-visible. `maintainer-repo release-check` currently reads the last digit for what is on `main`, and this changes that answer; the digit is a human decision under `docs/release.md`, so it is his.

The four published-count files are untouched, which is what he asked for on #335 in as many words: leave them alone rather than hand-edit, and he regenerates and pushes to your branch. So `rust` is red on the baseline check and that is the intended state of this PR, not a defect in it:

```
rust	Test workspace	test_baseline: rust suite has 1840 tests, baseline says 1837.
```

Every other check is green.

## Where this stands

Approved. The remaining step is his regeneration of the four count files, then it lands. Nothing is waiting on you.

You already have #337 open from me and unanswered, so I am not putting another issue in front of you; it stays yours if you want it.

One thing I appreciated: you asked about the validation constraint on #335 before writing the patch, rather than shipping a PR body that quietly omitted what you could not run. The reduced set you agreed to caught everything it was supposed to, and the red-before figure in your PR body reproduced on my machine to the test. That is a better provenance story than a green board with no account of how it got there.

LunaMeerkats:
Autonomous Codex follow-up: I am retaining the approved `fc5bd7f` head while the four count files are regenerated. The blank-intent test pins the existing planning-failure exit 3 as contrast coverage for timeout exit 2; it is not part of the red regression evidence. I have recorded the optional doc-comment suggestion and that distinction here.

I also re-triaged the #337 invitation against the latest scope correction. The remaining wizard guidance, redirected progress, and unattended prompt text form a bounded follow-up; provider selection remains separate. I will post the scoped claim there after the existing COSMICA integration checkpoint.

vladimirrott:
#381 aligns the exit codes and the timeout diagnostics, which is the follow-through on #335. Clean CLI work.

Two more that fit, reserved for you (and #337 is already assigned to you, since you had posted there):

- **#337** — the setup wizard's next-step commands fail and `--no-prompts` picks the wrong path. CLI ergonomics, right where you have been.
- **#342** — no `.gitattributes`, so a Windows or WSL clone gets CRLF and a byte-exact doc test blames the wrong thing. Small hygiene fix.

#342 carries the `claimed` label until you post on the thread; reply and it is yours. #381 is in review.

vladimirrott (approved):
Approving at `96dba92`, which carries one commit of mine on top of yours; details below.

The change I most wanted to see is `run_audit_checkpoint`. It printed its own diagnostic with `eprintln!` and then returned `CliError::Exit(2)`, hardcoding an execution-failure code for a configuration problem and printing the message outside the normal error path. You replaced both with `ConfigOrDaemon`, so the missing-database case exits 4 and prints through the same path as every other config error. That is the content of "align exit codes", and it is the part a reader skims past.

`TimedOut(u64)` earns its place for the same reason. The old code built an `ExecutionFailed` with a timeout string inside it, so `sysknife --timeout 1 …` told an operator an action had run and failed. It had not run.

What I verified, at `fc5bd7fb` and again after the rebase:

- `cargo clippy -p sysknife-cli --all-targets -- -D warnings`, clean.
- `cargo test -p sysknife-cli --test cli_smoke`: 20 passed, 1 ignored (live Postgres).
- Reverted your four production files to `main`, kept your tests, re-ran them:

```
failures:
    a_planning_timeout_does_not_claim_an_execution_failure
    audit_checkpoint_without_a_database_is_a_configuration_error
    timeout_flag_bounds_a_daemon_that_accepts_but_never_replies

test result: FAILED. 0 passed; 3 failed
```

The reverted binary prints `sysknife: execution failed: operation timed out after 1s`, which is the exact string your `.not()` assertion exists to catch. Your tests fail without your fix, which is the property that matters.

The listener in `a_planning_timeout_does_not_claim_an_execution_failure` is the nicest part of the diff. Binding a real socket that accepts and never answers produces a true planning-phase timeout with no provider and no daemon, and the closing `accept()` proves the CLI reached it.

**The commit I pushed.** Your three tests moved the workspace count from 1,838 to 1,841, and this repository publishes that number in `tests/evidence/workspace-tests.json` and in three prose claims (`README.md`, `docs/introduction.md`, `docs/distro-support.md`). Bumping the artifact alone turns `rust` green and `docs-and-hygiene` red, so all four move together or none do. I measured the merged tree, recorded 1,841, and updated the three literals in `96dba92`. Nothing of yours changed. That coupling is invisible to contributors, which is my problem rather than a mistake on your part; #278 tracks it.

That push dismissed my earlier approval, hence this one.

Nothing new to offer you. The wizard half of #337 is assigned to you already, and I would rather you have room for it than a queue.

One aside, since your diff is about what an exit code tells an operator. If you run anything unattended, `sysknife --json` emits one NDJSON object per event, and the codes above are the contract a wrapper script reads. All read-only under `--dry-run`. Worth ten minutes on a box you already trust, easy to skip.

PR #383, opened 2026-09-07
fix(ci): guard each live Postgres contract invocation

### LINKED ISSUE #362: postgres-contract's second live step is unguarded: three one-line changes make it green and run nothing
The `postgres-contract` job gained a second live step in #340:

```yaml
      - name: Run live CLI anchor exit-code contract
        run: >-
          cargo test -p sysknife-cli --test cli_smoke --locked
          audit_verify_exits_with_code_1_when_anchor_is_truncated -- --ignored --exact
```

The step is correct. `tests/release/postgres-contract-guard.test.sh` predates it
and looks for its two tokens anywhere in the job text, so nothing holds the new
step in place. I reviewed #340 and did not extend the guard to cover it, which is
my miss, not the author's. Three one-line mutations now leave that required check
green while it runs nothing, and the guard stays green through all three.

Measured at `60af0adb`, with `SYSKNIFE_TEST_POSTGRES_URL` and
`SYSKNIFE_REQUIRE_POSTGRES` set as the job sets them.

## 1. Drop `--ignored` and the step passes

```
$ cargo test -p sysknife-cli --test cli_smoke --locked \
    audit_verify_exits_with_code_1_when_anchor_is_truncated -- --exact
running 1 test
test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured; 17 filtered out
rc=0
```

The step configured as it stands today, on this machine with no database:

```
running 1 test
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 17 filtered out; finished in 30.00s
rc=101
```

This is #315 again, one job over. Note that the false green still prints
`running 1 test`, so that line does not separate the two. `0 ignored` does.

## 2. Misspell the test name by one character and the step passes

```
$ cargo test -p sysknife-cli --test cli_smoke --locked \
    audit_verify_exits_with_code_1_when_anchor_is_truncatd -- --ignored --exact
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 18 filtered out
rc=0
```

`--exact` matching nothing exits 0. This is the only name-filtered test
invocation on the whole board, and nothing checks that the name resolves.

For contrast, `cargo nextest` fails closed on an empty selection:

```
$ scripts/test_baseline.sh -E 'test(=frame_limit_is_declared_once)'
    Starting 0 tests across 41 binaries (1843 tests skipped)
error: no tests to run
rc=4
```

The `postgres-contract` job uses `cargo test`, which does not.

## 3. The guard sees none of it

I applied each mutation to a `git archive` copy of `60af0adb` and ran the guard:

| Mutation to `ci.yml` | Guard verdict |
|---|---|
| `--ignored` removed from the CLI step | `guard passed`, rc=0 |
| test name misspelled by one character | `guard passed`, rc=0 |
| the whole CLI step deleted | `guard passed`, rc=0 |

`SYSKNIFE_REQUIRE_POSTGRES` sits on the job's `env:` block and `--include-ignored`
sits on the store step, so both tokens are found no matter what happens to the
second step.

## 4. `ci-local.sh` never runs it at all

```
$ grep -c 'cli_smoke' scripts/ci-local.sh
0
```

Both branches of `run_postgres_contract_group` (`ci-local.sh:352` and `:395`) run
`--test postgres_store` and stop there. A contributor with podman installed runs
five of the six live tests and cannot reproduce the sixth, the one that pins the
anchor exit code. The half of #346 about `ci-local.sh` diverging from CI applies
here too, on a different set of commands.

## Fix

Read each cargo invocation separately rather than scanning the job for tokens,
and derive the set of live contracts from the Rust sources: a test file that
names `SYSKNIFE_REQUIRE_POSTGRES` is a live-Postgres contract, so some step has
to run it. Then per invocation, require an ignore flag, and require any
name filter to resolve to an `#[ignore]` test in the file that step targets.

Patch below. It changes `scripts/ci-local.sh` (adds the CLI step to both
branches) and rewrites the middle of
`tests/release/postgres-contract-guard.test.sh`. No workflow change: `ci.yml` is
already correct, and the point is to hold it there.

Applied to `60af0adb`:

```
$ git apply --check -v postgres-contract-second-step.patch
Checking patch scripts/ci-local.sh...
Checking patch tests/release/postgres-contract-guard.test.sh...
$ bash tests/release/postgres-contract-guard.test.sh
postgres-contract guard passed: 2 live contract file(s) (cli_smoke postgres_store), each run with an ignore flag by ci.yml and by both ci-local.sh branches.
rc=0
```

With only the guard half applied, before the `ci-local.sh` half, it names the
real defect and nothing else:

```
FAIL  scripts/ci-local.sh runs --test cli_smoke in 0 branch(es); both the exported-URL path and the container-start path must run it
1 postgres-contract guard failure(s).
rc=1
```

### Mutation matrix, eleven mutations, eleven reds

I applied each to a `git archive` copy with the patch on top, ran the guard, then
restored the file. Clean tree green before and after.

| # | Mutation | Guard now says |
|---|---|---|
| N1 | `ci.yml`: `--ignored` off the CLI step | runs cargo test without an ignore flag |
| N2 | `ci.yml`: test name misspelled | filters on a name that is not an `#[ignore]` test |
| N3 | `ci.yml`: CLI step deleted | no step against `--test cli_smoke` |
| N4 | `ci.yml`: `--include-ignored` off the store step | runs cargo test without an ignore flag |
| N5 | `ci.yml`: `SYSKNIFE_REQUIRE_POSTGRES` removed | job is missing the token |
| N6 | `ci.yml`: job renamed | no `postgres-contract` job to read |
| N7 | `ci-local.sh`: `--ignored` off both CLI branches | names both invocations |
| N8 | `ci-local.sh`: one CLI branch deleted | runs `--test cli_smoke` in 1 branch |
| N9 | `ci-local.sh`: `--include-ignored` off the store branches | names both invocations |
| I1 | env var renamed in both Rust files | the contract set cannot be derived |
| I2 | every `#[ignore]` stripped from `postgres_store.rs` | names the file |
| I3 | `cli_smoke.rs` deleted while `ci.yml` targets it | no such test target exists to run |

N1, N2 and N3 are the three the current guard passes.

I1 is the check on the guard's own input. The marker grep carries `|| true` so an
empty result reaches the diagnostic instead of dying at the assignment under
`set -o pipefail`, which is the shape #347 and #355 are about.

### Other checks run on the patched tree

```
$ find tests/e2e tests/release scripts assets/demo -type f -name '*.sh' -print0 \
    | xargs -0 shellcheck --severity=warning
rc=0
```

That is `e2e.yml`'s `scripts-lint` command verbatim, shellcheck 0.11.0. All
twenty-four scripts under `tests/release` and `tests/e2e` pass on the patched
tree, rc=0 each. The live half cannot run here, so the two
`cargo test` steps themselves are unproven against a real database; no container
runtime works on this host (`unshare -U` returns `Operation not permitted` under
`kernel.apparmor_restrict_unprivileged_userns=1`) and
`/usr/lib/postgresql/17` holds client tools only.

## Difficulty

`medium`. The patch is written and mutation-proven, so most of the work is
reading it and deciding whether deriving the contract set from
`SYSKNIFE_REQUIRE_POSTGRES` is the invariant you want. One judgement call is
open: the guard requires both `ci-local.sh` branches to run every contract, which
keeps the literal duplication the current guard already demands. Factoring both
branches into one helper would read better and would need the branch-count check
replaced with something else.

## Getting started

Related: #346 (`ci-local.sh` diverging from CI), #331 (a guard that every release
test is invoked by a gate), #313 and #315 (the two earlier false greens on this
same job). #340 introduced the step and the test; both are fine as written.

<details>
<summary><code>postgres-contract-second-step.patch</code></summary>

```diff
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -342,7 +342,7 @@
 
 run_postgres_contract_group() {
     printf '\n### postgres-contract (optional)\n'
-    local label="postgres-contract: cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored"
+    local label="postgres-contract: live Postgres contract (store + CLI anchor exit code)"
 
     if [[ "$run_postgres" != true ]]; then
         record SKIP "${label} (--no-postgres)"
@@ -351,7 +351,12 @@
 
     if [[ -n "${SYSKNIFE_TEST_POSTGRES_URL:-}" ]]; then
         SYSKNIFE_REQUIRE_POSTGRES=1 \
-            run_step "$label" cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
+            run_step "postgres-contract: cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored" \
+            cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
+        SYSKNIFE_REQUIRE_POSTGRES=1 \
+            run_step "postgres-contract: cargo test -p sysknife-cli --test cli_smoke --locked audit_verify_exits_with_code_1_when_anchor_is_truncated -- --ignored --exact" \
+            cargo test -p sysknife-cli --test cli_smoke --locked \
+            audit_verify_exits_with_code_1_when_anchor_is_truncated -- --ignored --exact
         return
     fi
 
@@ -392,9 +397,16 @@
         sleep "$POSTGRES_HEALTH_INTERVAL_SECS"
     done
 
-    SYSKNIFE_TEST_POSTGRES_URL="postgres://sysknife:sysknife@127.0.0.1:${POSTGRES_HOST_PORT}/sysknife_test?sslmode=disable" \
+    local container_url="postgres://sysknife:sysknife@127.0.0.1:${POSTGRES_HOST_PORT}/sysknife_test?sslmode=disable"
+    SYSKNIFE_TEST_POSTGRES_URL="$container_url" \
+    SYSKNIFE_REQUIRE_POSTGRES=1 \
+        run_step "postgres-contract: cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored" \
+        cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
+    SYSKNIFE_TEST_POSTGRES_URL="$container_url" \
     SYSKNIFE_REQUIRE_POSTGRES=1 \
-        run_step "$label" cargo test -p sysknife-daemon --test postgres_store --locked -- --include-ignored
+        run_step "postgres-contract: cargo test -p sysknife-cli --test cli_smoke --locked audit_verify_exits_with_code_1_when_anchor_is_truncated -- --ignored --exact" \
+        cargo test -p sysknife-cli --test cli_smoke --locked \
+        audit_verify_exits_with_code_1_when_anchor_is_truncated -- --ignored --exact
 
     "$runtime" rm -f "$POSTGRES_CONTAINER_NAME" >/dev/null 2>&1 || true
 }
--- a/tests/release/postgres-contract-guard.test.sh
+++ b/tests/release/postgres-contract-guard.test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #
-# postgres-contract-guard.test.sh — both halves of the live Postgres contract
-# must stay on the job that claims to run it.
+# postgres-contract-guard.test.sh — every live Postgres contract must stay on
+# the job that claims to run it, and on the local mirror of that job.
 #
 # #313 closed the case where postgres-contract reported success with no
 # database configured. The job now sets SYSKNIFE_REQUIRE_POSTGRES and runs
@@ -10,24 +10,32 @@
 # "6 passed; 5 ignored" and exits 0 — #294's shape with a different cause.
 # See issue #315.
 #
-# This check reads the job and fails when either token is missing. It does
-# not assert a test count: that would break the next time anyone adds a
-# store test and teach whoever hit it to edit the number rather than read
-# the check.
-#
-# Two design rules, both from tests/e2e/provider-parity.test.sh:
-#
-#   * The tokens are derived from postgres_store.rs, never restated as a
-#     count, so renaming the env var or dropping #[ignore] fails here
-#     until the job catches up.
+# The job later grew a second live step (#359, the CLI anchor exit-code
+# contract) and this check did not see it, because it looked for its two
+# tokens anywhere in the job rather than on each command. Three mutations of
+# that step left the guard green: dropping --ignored (cargo reports
+# "1 ignored", exit 0), misspelling the test name by one character (cargo
+# reports "running 0 tests", exit 0), and deleting the step outright. So the
+# check now reads each cargo invocation separately.
+#
+# Design rules, all three inherited from tests/e2e/provider-parity.test.sh:
+#
+#   * Everything is derived from the Rust sources, never restated as a
+#     count or a list. A test file that names SYSKNIFE_REQUIRE_POSTGRES is
+#     a live-Postgres contract file, so some step has to run it; renaming
+#     the env var or dropping #[ignore] fails here until the job catches up.
 #   * It covers every entry point, not just CI. ci-local.sh carries the
-#     pair in two branches (URL already exported, and the path that starts
-#     a container first). A guard that watches CI and ignores the local
-#     mirror leaves half the surface open.
+#     contract in two branches (URL already exported, and the path that
+#     starts a container first). A guard that watches CI and ignores the
+#     local mirror leaves half the surface open.
+#   * No test count is asserted. That would break the next time anyone adds
+#     a store test and would teach whoever hit it to edit the number rather
+#     than read the check.
 #
 # A pattern that matched nothing would make every assertion below vacuously
-# true, which is the exact failure this suite exists to stop. Missing files
-# and an empty job extract fail rather than pass over nothing to read.
+# true, which is the exact failure this suite exists to stop. Missing files,
+# an empty job extract and an empty marker set all fail rather than pass over
+# nothing to read.
 #
 # Host-side only: greps files, needs no VM, no daemon, no network.
 # Wired into docs-and-hygiene and scripts/ci-local.sh; a test nothing
@@ -37,41 +45,136 @@
 repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 ci_yml="$repo_root/.github/workflows/ci.yml"
 ci_local="$repo_root/scripts/ci-local.sh"
-store="$repo_root/crates/sysknife-daemon/tests/postgres_store.rs"
 
-for f in "$ci_yml" "$ci_local" "$store"; do
+for f in "$ci_yml" "$ci_local"; do
     [ -f "$f" ] || { printf 'FAIL: missing file: %s\n' "$f" >&2; exit 1; }
 done
 
-# Fail-closed env var, taken from the store test the job runs.
-# sort -u consumes every grep hit so a later head cannot SIGPIPE the pipeline
-# under `set -o pipefail`.
-require_token="$(grep -o 'SYSKNIFE_REQUIRE_POSTGRES' "$store" | sort -u || true)"
-if [ -z "$require_token" ]; then
-    printf 'FAIL: %s no longer names SYSKNIFE_REQUIRE_POSTGRES; the job token cannot be derived\n' \
-        "$store" >&2
-    exit 1
-fi
+failures=0
+report() {
+    printf 'FAIL  %s\n' "$1" >&2
+    failures=$((failures + 1))
+}
 
-# Live tests carry #[ignore]; without --include-ignored cargo never runs them
-# and still exits 0. Demand the attribute rather than trusting a count.
-if ! grep -Eq '#\[ignore' "$store"; then
-    printf 'FAIL: %s has no #[ignore] tests; --include-ignored would not run the live contract\n' \
-        "$store" >&2
+# --------------------------------------------------------------------------
+# The marker set: which test files are live Postgres contracts.
+# --------------------------------------------------------------------------
+# The fail-closed env var is the marker. A file that reads it refuses to pass
+# quietly without a database (#313), which is precisely what makes it a
+# contract someone has to run. Deriving the set means a new live test file is
+# covered the day it lands, with no list here to forget to update.
+#
+# `|| true` keeps an empty result reachable: grep exiting 1 on no match would
+# otherwise kill the script at the assignment under `set -o pipefail`, and the
+# diagnostic below would never print. That shape cost five guards their error
+# messages (#347).
+require_token='SYSKNIFE_REQUIRE_POSTGRES'
+marker_files="$(
+    grep -rl --include='*.rs' -F "$require_token" \
+        "$repo_root/crates" "$repo_root/apps" 2>/dev/null \
+        | grep '/tests/' | sort || true
+)"
+if [ -z "$marker_files" ]; then
+    printf 'FAIL: no test file under crates/ or apps/ names %s; the contract set cannot be derived\n' \
+        "$require_token" >&2
     exit 1
 fi
-# cargo's flag for #[ignore] tests. Named here once; the rust file proves
-# those tests are ignored, the job must still pass the flag.
-ignore_token='--include-ignored'
 
-tokens=("$require_token" "$ignore_token")
+# stem -> path, and the ordered list of stems.
+declare -A marker_path=()
+marker_stems=()
+while IFS= read -r path; do
+    [ -n "$path" ] || continue
+    stem="$(basename "$path" .rs)"
+    marker_path["$stem"]="$path"
+    marker_stems+=("$stem")
+    # Live tests carry #[ignore]; without an ignore flag cargo never runs them
+    # and still exits 0. Demand the attribute rather than trusting a count.
+    if ! grep -Eq '#\[ignore' "$path"; then
+        report "${path#"$repo_root"/} names $require_token but has no #[ignore] test; no ignore flag would run its contract"
+    fi
+done <<<"$marker_files"
+
+# cargo's two flags for #[ignore] tests. `--ignored` runs only them,
+# `--include-ignored` runs them alongside the rest; either is acceptable, and
+# neither is optional.
+ignore_flag_re='(--include-ignored|--ignored)'
+
+# --------------------------------------------------------------------------
+# Shared assertions for one cargo invocation.
+# --------------------------------------------------------------------------
+# `where` is a human-readable location for the diagnostic; `cmd` is the whole
+# invocation on one line with runs of whitespace squashed.
+check_invocation() {
+    local where="$1" cmd="$2"
+
+    if ! [[ "$cmd" =~ $ignore_flag_re ]]; then
+        report "$where runs cargo test without --ignored or --include-ignored, so its #[ignore] contract would be skipped and still exit 0"
+    fi
+
+    local target=""
+    if [[ "$cmd" =~ --test[[:space:]]+([A-Za-z0-9_]+) ]]; then
+        target="${BASH_REMATCH[1]}"
+    else
+        report "$where runs cargo test with no --test target, so which contract it covers cannot be derived"
+        return
+    fi
+
+    # Any bare word left after the options and before the `--` separator is a
+    # test-name filter. `cargo test NAME -- --exact` matching nothing prints
+    # "running 0 tests" and exits 0, so a filter that no longer resolves is a
+    # green step that tests nothing.
+    local head="${cmd%% -- *}"
+    local skip=0 tok filters=()
+    for tok in $head; do
+        if [ "$skip" = 1 ]; then skip=0; continue; fi
+        case "$tok" in
+            cargo | test) ;;
+            # A shell line continuation survives the whitespace squash.
+            '\') ;;
+            -p | --package | --test | --features | --manifest-path) skip=1 ;;
+            -* | *=*) ;;
+            *) filters+=("$tok") ;;
+        esac
+    done
 
-failures=0
-report() {
-    printf 'FAIL  %s\n' "$1" >&2
-    failures=$((failures + 1))
+    local file="${marker_path[$target]:-}"
+    if [ -z "$file" ]; then
+        # A --test target with no marker file is either a target that has no
+        # live contract, or one whose file has gone. Distinguish them: cargo
+        # errors on a missing target, but a green guard that names a target
+        # nothing can build is a guard reporting on nothing.
+        if ! compgen -G "$repo_root/*/*/tests/$target.rs" >/dev/null; then
+            report "$where targets --test $target, and no crates/*/tests/$target.rs or apps/*/tests/$target.rs exists to run"
+        fi
+        # Otherwise the ignore-flag check above still applies; there is no
+        # source of truth here to resolve a name against.
+        return
+    fi
+
+    for tok in ${filters+"${filters[@]}"}; do
+        # The filter has to name an #[ignore]d test in the file this step
+        # targets. Walk up from the fn to the nearest attribute block so a
+        # non-ignored test cannot satisfy an --ignored run.
+        if ! awk -v name="$tok" '
+            $0 ~ ("(^|[^A-Za-z0-9_])fn[[:space:]]+" name "[[:space:]]*\\(") {
+                for (i = NR - 1; i >= 1 && i >= NR - 6; i--) {
+                    if (line[i] ~ /#\[ignore/) { found = 1 }
+                    if (line[i] ~ /^[[:space:]]*}/) { break }
+                }
+                if (found) { exit 0 }
+            }
+            { line[NR] = $0 }
+            END { exit found ? 0 : 1 }
+        ' "$file"; then
+            report "$where filters on '$tok', which is not an #[ignore] test in ${file#"$repo_root"/}; the step would run 0 tests and exit 0"
+        fi
+    done
 }
 
+# --------------------------------------------------------------------------
+# .github/workflows/ci.yml
+# --------------------------------------------------------------------------
 # Top-level GitHub Actions job. Empty extract is a failure: grepping nothing
 # would report every token present.
 extract_job() {
@@ -84,49 +187,105 @@
     ' "$file"
 }
 
+# One record per `- name:` step, folded scalars joined, whitespace squashed, so
+# a flag on one step cannot satisfy an assertion about another. That blindness
+# is what let #359's step regress unnoticed.
+extract_steps() {
+    awk '
+        /^      - name:/ {
+            if (buf != "") printf "%s\x1e", buf
+            buf = $0
+            next
+        }
+        buf != "" { buf = buf " " $0 }
+        END { if (buf != "") printf "%s\x1e", buf }
+    ' <<<"$1" | tr -s '[:space:]' ' ' | tr '\036' '\n'
+}
+
 job="$(extract_job "$ci_yml" "postgres-contract")"
 if [ -z "$job" ]; then
     report "$ci_yml has no postgres-contract job to read"
 else
-    for token in "${tokens[@]}"; do
-        grep -Fq -- "$token" <<<"$job" \
-            || report "$ci_yml postgres-contract job is missing $token"
+    grep -Fq -- "$require_token" <<<"$job" \
+        || report "$ci_yml postgres-contract job is missing $require_token"
+
+    ci_targets=""
+    while IFS= read -r step; do
+        [ -n "$step" ] || continue
+        case "$step" in
+            *"cargo test"*) ;;
+            *) continue ;;
+        esac
+        step_name="$(sed -E 's/^ *- name: *//; s/ run:.*//' <<<"$step")"
+        # Drop the step's name and the `run: >-` preamble; the words in a step
+        # title are not arguments, and reading them as such reports a filter
+        # per word.
+        step="cargo test ${step#*cargo test }"
+        check_invocation "$ci_yml postgres-contract step '$step_name'" "$step"
+        if [[ "$step" =~ --test[[:space:]]+([A-Za-z0-9_]+) ]]; then
+            ci_targets="$ci_targets ${BASH_REMATCH[1]}"
+        fi
+    done < <(extract_steps "$job")
+
+    # Every live-contract file needs a step. Deleting #359's step, which this
+    # check used to allow, fails here.
+    for stem in "${marker_stems[@]}"; do
+        case " $ci_targets " in
+            *" $stem "*) ;;
+            *) report "$ci_yml postgres-contract job runs no step against --test $stem, whose ${marker_path[$stem]##*/} is a live $require_token contract" ;;
+        esac
     done
 fi
 
-# Each cargo invocation of postgres_store, including the env lines that
+# --------------------------------------------------------------------------
+# scripts/ci-local.sh
+# --------------------------------------------------------------------------
+# Each cargo invocation reached through run_step, including the env lines that
 # continue onto it. The label string also names the command, so only
-# statements that actually call run_step count as a branch.
+# statements that actually call run_step count.
 extract_invocations() {
     awk '
         {
             if (buf != "") buf = buf "\n" $0
             else buf = $0
             if ($0 ~ /\\$/) next
-            if (buf ~ /run_step/ && buf ~ /postgres_store/) printf "%s\x1e", buf
+            if (buf ~ /run_step/ && buf ~ /cargo test/) printf "%s\x1e", buf
             buf = ""
         }
     ' "$1"
 }
 
-branch_count=0
+declare -A local_target_count=()
+invocation=0
 while IFS= read -r -d $'\x1e' stmt; do
-    [ -z "$stmt" ] && continue
-    branch_count=$((branch_count + 1))
-    for token in "${tokens[@]}"; do
-        grep -Fq -- "$token" <<<"$stmt" \
-            || report "$ci_local postgres-contract branch $branch_count is missing $token"
-    done
+    [ -n "$stmt" ] || continue
+    invocation=$((invocation + 1))
+    squashed="$(tr -s '[:space:]' ' ' <<<"$stmt")"
+    # The label repeats the command verbatim, so strip it before parsing or the
+    # quoted copy is read as a second invocation.
+    parsed="${squashed#*run_step }"
+    parsed="cargo test ${parsed#*cargo test }"
+    check_invocation "$ci_local invocation $invocation" "$parsed"
+    if [[ "$parsed" =~ --test[[:space:]]+([A-Za-z0-9_]+) ]]; then
+        target="${BASH_REMATCH[1]}"
+        local_target_count["$target"]=$(( ${local_target_count["$target"]:-0} + 1 ))
+    fi
 done < <(extract_invocations "$ci_local")
 
-if [ "$branch_count" -lt 2 ]; then
-    report "$ci_local has $branch_count postgres-contract run_step branch(es); both the exported-URL path and the container-start path must carry the pair"
-fi
+# Both entry paths, for every contract. ci-local.sh reaches the database two
+# ways (URL already exported, container started here) and a contributor who
+# takes either one has to run the same set CI does.
+for stem in "${marker_stems[@]}"; do
+    count="${local_target_count[$stem]:-0}"
+    if [ "$count" -lt 2 ]; then
+        report "$ci_local runs --test $stem in $count branch(es); both the exported-URL path and the container-start path must run it"
+    fi
+done
 
 if [ "$failures" -ne 0 ]; then
     printf '\n%d postgres-contract guard failure(s).\n' "$failures" >&2
     exit 1
 fi
 
-printf 'postgres-contract guard passed: %s and %s present on the CI job and %d ci-local.sh branches.\n' \
-    "$require_token" "$ignore_token" "$branch_count"
+printf 'postgres-contract guard passed: %d live contract file(s) (%s), each run with an ignore flag by ci.yml and by both ci-local.sh branches.\n' \
+    "${#marker_stems[@]}" "${marker_stems[*]}"
```

</details>

QinXi-ai:
The live CLI anchor test could be deleted, misspelled, or run without `--ignored` while the postgres-contract guard stayed green. The guard now derives live test targets from the Rust requirement marker and checks each CI/local invocation, its ignore flag, and any ignored-test name filter. Both local database paths now run the CLI contract alongside the store contract.

This follows the issue's proposed patch, corrects its #340 attribution, and fixes an additional parser gap: the quoted `run_step` label cannot satisfy checks on the command that actually executes. The guard also rejects disabled required-database settings and ambiguous duplicate target names. The local reproduction documentation includes both commands.

- The guard passes before and after mutation testing.
- All 16 mutations were rejected, including the three issue examples, executable-argv changes with intact labels, absent/disabled database requirement settings, missing source targets, and missing Rust markers/ignore attributes.
- ShellCheck 0.11.0 (`--severity=warning`) passed for both changed shell scripts.
- `python -X utf8 scripts/check_evidence_claims.py` and `git diff --check` passed.

The live database tests were NOT_TESTED locally; this host cannot build the Unix-only workspace. No Rust test counts changed. Codex assisted with this change.

Closes #362.

The three original false greens were also reproduced against the unchanged d881d49 guard, with source files restored afterward. The 16-mutation matrix additionally deletes the CLI invocation from just one local path; the new guard rejects it.

Linux CI at commit 190a3ad: all applicable checks passed, including the full Rust job, live postgres-contract, docs-and-hygiene, scripts-lint and CodeQL. CI run: https://github.com/lacs-project/sysknife/actions/runs/34081898044.


vladimirrott (changes requested):
This is the right rewrite. Parsing each invocation instead of grepping the whole job is the correct reading of #362, and the diagnostics are the best part of it: the misspelled-filter case names the token, the file it searched, and what would have happened ("the step would run 0 tests and exit 0"). Most guards say less than that when they fire.

I reproduced your three regressions in a scratch clone at `190a3adf`, measuring the exit code rather than reading the output, since a pipe to `tail` will hand you a `0` from `tail` and I nearly fooled myself with exactly that:

```
S0 pristine                                rc=0
M1 drop --ignored from the CLI step        rc=1
M2 misspell the filter by one character    rc=1
M3 delete the CLI step outright            rc=1
```

All three are green on `main` today. That is the case for this PR on its own.

**One blocking item.** The marker set can shrink without the guard noticing, and whichever file drops out loses its coverage silently:

```
M4 rename the marker in cli_smoke.rs       rc=0   passed: 1 live contract file(s) (postgres_store)
M5 rename the marker in postgres_store.rs  rc=0   passed: 1 live contract file(s) (cli_smoke)
M6 remove it from both                     rc=1   fails closed, correctly
```

M5 is the one that bites. The job still runs `--test postgres_store`, but the test no longer reads a variable CI sets, so its fail-closed check is inert and a missing database skips and exits 0. That is #313's shape returning for that file, and the old guard caught it, because it hardcoded `postgres_store.rs` and failed when the token left it.

The header states the guarantee outright: "renaming the env var or dropping `#[ignore]` fails here until the job catches up." M4 and M5 make that sentence false, and a guard documenting a guarantee it does not provide is what this repo fails CI over everywhere else.

The asymmetry is that `marker -> step` is checked and `step -> marker` is not: `check_invocation` returns quietly when the target has no marker file but the file exists. The other half closes it:

```diff
         if ! compgen -G "$repo_root/*/*/tests/$target.rs" >/dev/null; then
             report "$where targets --test $target, and no crates/*/tests/$target.rs or apps/*/tests/$target.rs exists to run"
+        else
+            report "$where targets --test $target, whose file no longer names $require_token; the contract set narrowed under the job"
         fi
```

I ran that before suggesting it, so it is tested rather than sketched:

```
S0 pristine + fix   rc=0   passed: 2 live contract file(s) (cli_smoke postgres_store)
M4 with fix         rc=1
M5 with fix         rc=1
S0 restored         rc=0
```

No false positives on the current tree. The nearby comment about a target having "no source of truth here to resolve a name against" then wants a small edit, since there would be one.

I also tried dropping a single `#[ignore]` from the store contract and the guard stayed green. That one is not a defect: you demand the attribute rather than a count, on purpose, and my mutation left the other ignored tests in place. Recording it so you know it was checked.

Nothing else blocks. One optional note: the `>= 2` branch count in `ci-local.sh` reads as "both paths" only because there happen to be two today. A third entry path would satisfy it while skipping one. Worth a word in the comment rather than a change.

Your CHANGELOG entry is accurate and I would keep it exactly as written.

No new issue from me while #267, #239 and #238 are still open on your side; I would rather you have room than a queue. And since this diff is about proving a database contract actually ran: if you run Postgres anywhere yourself, `sysknife audit verify` checks a signed action chain against a live database and exits non-zero on a truncated anchor, read-only. It is the same property you are guarding here, from the operator's end.

vladimirrott (approved):
Approved at `bacc07d39e4292b621f0f506970dc80b695f4463`.

The bidirectional check is the part that matters, and you built it where the one-directional check lived rather than bolting a second pass beside it. Renaming the marker in either contract file now fails the guard by name, which is what the old shape could not see:

```
FAIL  /repo/scripts/ci-local.sh invocation 2 targets --test cli_smoke, whose file no longer names SYSKNIFE_REQUIRE_POSTGRES; the contract set narrowed under the job
```

`maintainer screen 383` returns DO NOT EXECUTE, so I ran nothing on the host. Everything below ran in `docker.io/library/debian:stable-slim` under rootless podman with `--network=none` and the tree mounted from `git archive bacc07d3`.

## Mutations

| # | mutation | guard |
|---|---|---|
| M0 | none | pass: `2 live contract file(s) (cli_smoke postgres_store)` |
| M1 | drop `-- --include-ignored` from the ci.yml store step | FAIL, naming the step |
| M2 | replace the job's `SYSKNIFE_REQUIRE_POSTGRES` env key | FAIL: `job is missing SYSKNIFE_REQUIRE_POSTGRES` |
| M3 | point the CLI filter at a name that does not exist | FAIL: `would run 0 tests and exit 0` |
| M4a | rename the marker in `cli_smoke.rs` to `SYSKNIFE_NEEDS_PG` | FAIL ×2, both `ci-local.sh` branches |
| M4b | same rename in `postgres_store.rs` | FAIL ×2 |
| M6 | delete the `#[ignore]` lines from `postgres_store.rs` | FAIL: `has no #[ignore] test` |
| M7 | delete the CLI anchor step from `ci.yml` | FAIL: `runs no step against --test cli_smoke` |
| M9 | delete the anchor invocations from both `ci-local.sh` branches | FAIL: `runs --test cli_smoke in 0 branch(es)` |
| M10 | restore everything | pass |

The command, verbatim:

```
$ bash /home/entropia/.local/state/sysknife-maint/pti.sh docker.io/library/debian:stable-slim p383 bash -c "$(cat /tmp/sk383proof2.sh)"
### M4a. rename the marker in cli_smoke.rs to a token that does NOT contain the old one
old token still present in cli_smoke.rs: 0
new token count: 2
FAIL  /repo/scripts/ci-local.sh invocation 2 targets --test cli_smoke, whose file no longer names SYSKNIFE_REQUIRE_POSTGRES; the contract set narrowed under the job
FAIL  /repo/scripts/ci-local.sh invocation 4 targets --test cli_smoke, whose file no longer names SYSKNIFE_REQUIRE_POSTGRES; the contract set narrowed under the job

3 postgres-contract guard failure(s).
exit=1
```

Two things I got wrong on the way, since they say something about the guard rather than about you.

My first M4 renamed the token to `SYSKNIFE_REQUIRE_POSTGRES_RENAMED`, and the guard read green. `grep -F "$require_token"` still matched the longer name as a substring, so my mutation never applied. That is the mutation-did-not-land trap, and it was mine.

Before that I ran the whole matrix in `docker.io/library/bash:5` and every case printed the same line: `no test file under crates/ or apps/ names SYSKNIFE_REQUIRE_POSTGRES`. Alpine's busybox `grep` has no `--include`, so the marker derivation returned empty and the guard exited before reaching any mutation. Ten identical outputs is what a vacuous run looks like. Your `|| true` comment about #347 is why the diagnostic printed at all instead of the script dying silently at the assignment.

## Not blocking, worth a line each

`bacc07d3` is a merge commit that carries the fix inline, so `git log 190a3adf..bacc07d3` shows main's commits and the merge but not the change I asked for. The review diff is clean; a rebase would make the fix reviewable as one commit if you push again for another reason.

`check_invocation` runs against every `run_step` plus `cargo test` statement anywhere in `ci-local.sh`, not only the Postgres group. Nothing trips it today: I checked, and the only `cargo test` statements in that file are the four in `run_postgres_contract_group`, while every other Rust step is `cargo nextest run`, `clippy`, `fmt`, `doc` or `audit`. A future non-Postgres `cargo test --test X` step would fail the guard. Pre-existing, not yours.

## The merge is blocked by my tooling, not by this PR

All eleven checks are green and nothing is pending. The gate refused:

```
$ maintainer-merge verify 383 bacc07d39e4292b621f0f506970dc80b695f4463 "tests/release/postgres-contract-guard.test.sh" "/audit_verify_exits_with_code_1_when_anchor_is_truncated/d"
maintainer-merge: no single suite in /home/entropia/.local/share/maintainer/profiles/sysknife/verify.d covers every changed path:
    CHANGELOG.md	docs
    docs/developer-guide.md	docs
    scripts/ci-local.sh	shell
    tests/release/postgres-contract-guard.test.sh	shell

  Suites that would cover the rest between them: docs shell.
```

A shell change plus the CHANGELOG entry the repo asks for spans two suites, so no receipt is earnable. Do not split the PR: the CHANGELOG entry belongs with the change. That part is mine to run, and I have the exact commands.

## Next

`#397` is the obvious follow-on for you: `CHAIN_ROW_COLUMNS` is declared twice and only a live Postgres run notices when the copies drift, the same shape as what you fixed here, one layer down. I am not offering it yet because you already hold #362 and #237; when both land I will bring it to you first.

You read CI logs and run Postgres for a living, going by this diff. If you ever want to point SysKnife at one of your own Ubuntu boxes, the read-only surface is the place to start (`sysknife "show me disk usage and failed services"` plans and previews without executing anything). No obligation, and it has nothing to do with this merging.

vladimirrott:
Merged as `f61e76f`, closing #362.

You ran sixteen mutations and said which platform each one ran on, including the cases Git Bash could not reach. A mutation table is worth what its honesty about the gaps is worth, and yours is the one I trust most on this tracker.

Workspace at `5673d20`:

```
```

#238 came your way as #384 merged, so you have that one if you want it. I recounted its arithmetic afterwards, because your own #384 is what invalidated it: 57 Debian-family actions run unchanged rather than 43, and 154 of 189 rather than 140. The body carries the recount and the two scripts behind it, so start from the body rather than from my offer.

Issue #239, opened 2026-08-18
Debian's default firewall is nftables, and the catalogue has no nftables vocabulary


vladimirrott:
> **Cross-references corrected 2026-09-14 at `61b3a878`.** Three issues this body
> pointed at have since closed (#234, #237, #238), and one of them changed the
> premise rather than the detail: Debian is eligible now, so the first paragraph
> below was written in the conditional and should not have stayed there. The
> defect itself is unchanged and still reproduces:
>
> ```
> $ git rev-parse --short=8 HEAD
> 61b3a878
> $ grep -c Ufw crates/sysknife-types/src/lib.rs
> 8
> $ grep -ci nft crates/sysknife-types/src/lib.rs
> 0
> ```
>
> **Scope 1 is taken.** PR #413 by @QinXi-ai is open against the read-only half
> (`GetNftablesRuleset` and a backend reporter). Scope 2 is untouched.

The catalogue has eight `ufw` actions and no nftables ones. On Ubuntu that is the right call, because ufw is the family's canonical frontend. On Debian it is not: Debian's own handbook names nftables as "the default firewall framework in Debian Buster and later", and stock Debian installs no ufw.

Debian eligibility landed on 10 September. #238 is closed and `is_supported` now accepts Debian 12 and later, so SysKnife already runs on hosts where its only firewall vocabulary is absent.

## The two failure shapes

**ufw absent.** `UfwStatus` and the seven mutating ufw actions fail at exec with a command-not-found, which surfaces as `execution_failure` with no explanation of why.

**ufw installed but idle.** Worse, and this repo has already been burned by this exact shape. The `NON_CANONICAL_ON_DEBIAN` doc comment (`crates/sysknife-core/src/action_family.rs:98`) records it:

> an Ubuntu host running firewalld lost firewall management entirely, and `UfwStatus` answered `Status: inactive` — a confident wrong answer instead of a refusal.

Substitute nftables for firewalld and Debian for Ubuntu and it is the same bug. An operator asking "is the firewall on" gets "inactive" from ufw while nftables is filtering traffic.

## Scope

Two independent pieces; either is a reasonable PR.

**1. Read-only truth first.** A `GetNftablesRuleset` action over `sudo nft list ruleset`, and a state read that reports which backend is in force rather than which frontend was queried. This is the piece that stops the confident wrong answer, and it is Low risk with no approval receipt, so it is the smaller start.

[...]

Issue #238, opened 2026-08-18
Make Debian eligible: a version floor of 12, and a reason in is_supported where there is none


vladimirrott:
`docs/distro-support.md:120` already promises this (this issue said line 91 when it was filed; the row has been at 120 the whole time):

| Debian stable/testing | Planned after Ubuntu hardening |

Ubuntu hardening is where it was meant to be: three LTS releases validated at 79/79 on live VMs with committed replay twins. This issue is the follow-through.

## What already exists

Debian is not an unrecognised distro. It is a parsed, first-class variant that is deliberately refused:

- `DistroId::Debian { version: Option<u32> }` (`crates/sysknife-core/src/distro.rs:142`)
- `detect_debian` parses `VERSION_ID` into that field (`distro.rs:453`)
- `family()` already returns `DistroFamily::Debian` for it (`distro.rs:205`)
- a Debian 12 bookworm fixture and its parse and detect tests exist (`distro.rs:643`, `:764`, `:958`)
- Linux Mint and Pop!_OS already fall back to the Debian family via `ID_LIKE`

The single line that refuses it (`distro.rs:252`):

```rust
Self::UbuntuCore { .. } | Self::Debian { .. } | Self::Other { .. } => false,
```

Note what the doc comment above it justifies and what it does not. Pre-20.04 Ubuntu is excluded because it stops receiving security updates. `UbuntuCore` is excluded structurally, having no apt and a read-only root. Plain Fedora stays detectable so the planner can explain the gap. **Debian is given no reason at all.** It sits in that arm because nothing has needed it to be elsewhere.

## Why it is cheap

Of the 67 Debian-family actions, 57 run on Debian unchanged, and all 97 `All`
actions already do. That is 154 of 189, ≈ 81.5 %, with no new action code. #237
did the inventory and the list surgery, and it merged as #384 today, so nothing
blocks this now.

> **Recounted 2026-09-09 at `5673d20`.** This paragraph said 43 actions and 140
> of 189 when it was filed. #237 split the fences, so the 57 now falls out of
> the constants directly: `DEBIAN_ONLY_ACTIONS` (19) plus
> `NON_CANONICAL_ON_FEDORA` (38), leaving behind the 10 in
> `UBUNTU_ONLY_ACTIONS`. The three sets are disjoint and their union is exactly
> the 67.
>
> ```
> $ python3 - <<'PY'
> import re, pathlib, collections
[...]

PR #384, opened 2026-09-07
fix(core): separate Ubuntu action fences from planner defaults

### LINKED ISSUE #237: Split the Ubuntu-only actions out of DEBIAN_ONLY_ACTIONS, so the family fence means what it says

QinXi-ai:
`DEBIAN_ONLY_ACTIONS` mixed Ubuntu identity requirements, Debian mechanisms, and portable planner defaults. This separates the hard fences from preferences and carries the detected distribution ID into planning, so Debian-family membership alone cannot offer or authorize Ubuntu-only actions.

- `UBUNTU_ONLY_ACTIONS` contains Canonical services, release upgrades, PPAs, and the reboot sentinel. PPAs target Ubuntu series; the sentinel remains conservatively Ubuntu-only until a Debian producer/backend is validated.
- The reverse mechanism assertion also exposed portable entries in the old fence (hostnamectl, ufw, AppArmor, fail2ban, Flatpak, distrobox, cloud-init, snap, netplan, Multipass). These are planner preferences rather than execution impossibilities. The Ubuntu and Fedora default catalogues remain unchanged; an operator may execute installed portable tooling on an eligible host.
- Debian-host preferences withhold snap/netplan/Multipass without removing those defaults from Ubuntu. The CLI, daemon and direct MCP routing share the hard classification. Debian host eligibility remains unchanged.
- The generated reference retains its default-catalogue labels and now explains that those labels include preferences. Tests check hard classification in both directions, Debian-vs-Ubuntu planning/prose, routing, and the direct MCP boundary.

Public Rust API change: `DistroHint` gains `id`, and `propose_plan_tool_def` takes the full hint. Under the repository version policy this needs a middle-digit release in the `0.y` series.

Validation on Windows:
- 109 tests passed in a standalone harness compiling the exact distro, action-family, CLI routing and prompt source files with the current DistroHint definition extracted from sysknife-types. This is not a workspace result.
- `cargo fmt --all -- --check`, locked cargo metadata, `python -X utf8 scripts/check_evidence_claims.py`, and `git diff --check` passed.
- Full workspace execution, the mechanism/reference integration tests, and live Debian/Ubuntu validation are NOT_TESTED locally. The same checkout baseline cannot compile the unchanged Unix-only vsock dependency on Windows; Linux CI must validate the integration.

Five Rust tests were added. The four published-count files remain untouched under CONTRIBUTING.md's supported maintainer-regeneration fallback. Codex assisted with this change.

Closes #237.

Linux CI at Rust implementation commit 37e3a12: 1842 tests run, 1842 passed, 6 skipped. This includes the bidirectional mechanism fence, generated action reference, Debian/Ubuntu schema and prompt tests, daemon platform checks and direct MCP routing. Format, Clippy, rustdoc, docs-and-hygiene, live postgres-contract and scripts-lint passed. The rust job fails only because the recorded baseline remains 1837; the supported maintainer-regeneration fallback applies. Run: https://github.com/lacs-project/sysknife/actions/runs/34082386889/job/101620060267.

The subsequent commit 3965354 only adds the compatibility chapter to docs/SUMMARY.md so mdBook publishes the linked page; it changes no Rust source. Its CI rerun is in progress.


vladimirrott:
Three PRs in a day across the daemon (#382), the Postgres contract (#383) and the core planner (#384). That is real range, and they are holding up in review.

You are plainly at home in the core action model, so three that suit you, all reserved for you:

- **#238** — make Debian eligible: a version floor of 12 and a reason in `is_supported`. The natural next step after your Ubuntu-fence work in #384.
- **#239** — Debian's default firewall is nftables, and the catalogue has no action for it. A new action family, your comfort zone.
- **#267** — `open_read_only` writes: `audit verify` and `audit export` migrate the DB when they should only read it. Deeper in the daemon.

GitHub will not let me hard-assign you until you post on a thread, so they carry the `claimed` label for now; reply on any and it is yours. Reviews of your three are in progress.

vladimirrott (changes requested):
The decomposition is right, and it is the part I want to keep. `DEBIAN_ONLY_ACTIONS` was carrying three unrelated meanings at once: mechanisms that genuinely cannot exist on Fedora (apt, dpkg), Canonical identity requirements (PPAs, Pro, Livepatch), and tools that are merely not somebody's default catalogue (snap, netplan, AppArmor, flatpak). Only the first two are fences. Splitting them and adding `id` to the hint so Debian-family membership stops implying Ubuntu is the correct read of #237.

The `CheckPendingReboot` note is the sort of care that is hard to review and easy to skip writing: it depends on Ubuntu's update-notifier sentinel, so a *missing* sentinel on Debian is not evidence that no reboot is pending. Keeping it Ubuntu-only until a Debian producer exists fails closed. Same for refusing to normalise the portable tools into a fence just because they were sitting in one.

**One blocking item, and it is about the tests rather than the change.**

`ubuntu_identity_is_stricter_than_debian_family` cannot fail on account of which list an action is in, because it iterates the lists and then asks a function that reads the same lists:

```rust
if UBUNTU_ONLY_ACTIONS.contains(&action) { matches!(distro, DistroId::Ubuntu { .. }) }
else if DEBIAN_ONLY_ACTIONS.contains(&action) { distro.family() == DistroFamily::Debian }
```

For any action in `UBUNTU_ONLY_ACTIONS` the answer is Ubuntu-yes/Debian-no by construction; for any action in `DEBIAN_ONLY_ACTIONS` it is yes to both, since Ubuntu's family is Debian. Both loops hold whatever the lists contain. I moved `AddPpa` and `RemovePpa` out of `UBUNTU_ONLY_ACTIONS` and back into `DEBIAN_ONLY_ACTIONS`, which is precisely the #237 regression this PR exists to prevent, and verified the move landed before trusting the result:

```
baseline:   UBUNTU_ONLY has AddPpa: True  | DEBIAN_ONLY has AddPpa: False
mutated:    UBUNTU_ONLY has AddPpa: False | DEBIAN_ONLY has AddPpa: True

running 2 tests
test result: ok. 2 passed; 0 failed
```

Two tests ran, so the filter selected something, and both passed. A Debian host regaining PPAs would ship green.

The fix is to anchor some assertions on names instead of on membership, so the two sides stop moving together. I wrote it and ran it both ways rather than sketching it:

```rust
#[test]
fn named_actions_keep_their_distro_boundary() {
    use crate::distro::DistroId;
    let debian = DistroId::Debian { version: Some(13) };
    let ubuntu = DistroId::Ubuntu { major: 24, minor: 4 };
    for action in ["AddPpa", "RemovePpa", "ProAttach", "LivepatchStatus"] {
        assert!(!action_matches_distro(action, &debian), "{action} on Debian");
        assert!(action_matches_distro(action, &ubuntu), "{action} on Ubuntu");
    }
    for action in ["AptInstall", "AptUpdate"] {
        assert!(action_matches_distro(action, &debian), "{action} on Debian");
    }
}
```

```
pristine tree            -> 1 passed
PPAs moved to the fence  -> FAILED: "AddPpa on Debian"
```

`planner_preferences_are_not_execution_fences` does not have this problem, for the record: it compares two different sets, so an action appearing in both a preference list and a fence is caught. It is the identity test specifically.

**Two things that are not review items.**

The published test-count baseline is mine to move, not yours, and you were right to leave it: CONTRIBUTING says so and the coupling reaches three prose files besides the artifact. I will bump it when you push.

The version consequence needs a decision from me before this lands, and I do not want it to look like I am sitting on your work. `DistroHint` gaining a public field and `propose_plan_tool_def` changing signature is a middle-digit move under our policy, so merging this makes the next release 0.14.0 and forecloses the 0.13.2 I currently have queued from the fixes that landed today. That is sequencing on my side, and I will tell you which way it goes rather than leaving the PR quiet.

Your Windows constraint is handled: I ran the fence tests on Linux, and the numbers above are from that. Nothing in your PR needs a Linux machine from you.

vladimirrott:
Coming back on the version question, as promised, and the answer is that it was never yours to wait on.

I said merging this would force 0.14.0 and foreclose the 0.13.2 I had queued. The first half is right and the second half was wrong: `main` already requires a middle-digit bump without your PR. #381 landed earlier today and changed `audit checkpoint` with no database from exit 2 to exit 4. `docs/release.md` is explicit that behaviour counts, not only signatures, and gives almost exactly this shape as its worked example:

> v0.9.0 also made `sysknife-setup` refuse a malformed `.mcp.json` that earlier versions overwrote, so a run that used to succeed now exits 1. No signature changed and it is still a compatibility break.

An exit code is the contract a wrapper script reads, and `docs/cli.md` says so in the table that same PR updated. So the next release is 0.14.0 whatever happens here, and `DistroHint` gaining a public field costs nothing that was not already spent.

I am cutting 0.14.0 from `main` as it stands rather than holding it for this PR, because I would rather two security fixes in the daemon's authorization path ship now than wait on a review round. That means this lands in the release after, which is not a demotion and not a comment on the work.

So the only thing still open on #384 is the one from my earlier review: `ubuntu_identity_is_stricter_than_debian_family` iterates the fence lists and then asks a function that reads those same lists, so moving `AddPpa` back into the Debian fence passes 2 of 2. No rush on it, and the suggested test is in that review, run both ways.

vladimirrott (changes requested):
Reviewed at `7305f5c9ad6dff687f17bb7c06052d487ab0c232`.

The taxonomy is right and the partition is exact. `DEBIAN_ONLY_ACTIONS` was carrying three different meanings, and splitting it into mechanism, Canonical identity and non-default tool is the decomposition that makes each one testable. I checked the split by set arithmetic rather than by reading:

```
$ ex() { git show "$1:crates/sysknife-core/src/action_family.rs" | awk "/^pub const $2: /,/^\];/" | grep -oE '"[A-Za-z0-9_]+"' | tr -d '"' | sort -u; }
$ ex origin/main DEBIAN_ONLY_ACTIONS > /tmp/m_deb.txt
$ ex p384 DEBIAN_ONLY_ACTIONS > /tmp/p_deb.txt
$ ex p384 UBUNTU_ONLY_ACTIONS > /tmp/p_ubu.txt
$ ex p384 NON_CANONICAL_ON_FEDORA > /tmp/p_ncf.txt
$ cat /tmp/p_deb.txt /tmp/p_ubu.txt /tmp/p_ncf.txt | sort -u > /tmp/p_union.txt
$ comm -23 /tmp/m_deb.txt /tmp/p_union.txt   # in M but not in the union
$ comm -13 /tmp/m_deb.txt /tmp/p_union.txt   # in the union but not in M
main DEBIAN_ONLY      |M| = 67
p384 DEBIAN_ONLY      |D| = 19
p384 UBUNTU_ONLY      |U| = 10
p384 NON_CANON_FEDORA |N| = 38
--- D ∪ U ∪ N vs M ---
union size = 67
in M but not in union (LOST):
  (none above = nothing lost)
in union but not in M (NEW):
  (none above = nothing invented)
--- FEDORA_ONLY_ACTIONS unchanged? ---
identical (25 names)
```

|D| + |U| + |N| = 19 + 10 + 38 = 67 = |M|, and D ∪ U ∪ N = M exactly. No action fell out of every list, none appeared from nowhere, and the Fedora fence is byte-identical. That is the hard part of a change like this and you got it right.

You also fixed the self-referential test I blocked on last time. `action_family.rs:311` names the boundaries as literals instead of iterating the production lists, so moving `AddPpa` back into `DEBIAN_ONLY_ACTIONS` now fails the second assertion. Adding `CheckPendingReboot` and the `linuxmint` derivative went past what I asked for.

`maintainer screen 384` returns DO NOT EXECUTE. Everything below ran in `docker.io/library/rust:1-slim` under rootless podman, `--network=none`, tree from `git archive 7305f5c9`, with a target directory isolated per tree.

## Blocking 1: the client now permits what the daemon refuses

The comment three lines above the line you changed states the invariant:

```rust
// A client that refuses what the daemon would run is confusing; a client that
// *permits* what the daemon refuses is worse, so both stay strict together.
if action_requires_distro(action_name) && !distro.is_supported() {
```

`action_requires_distro` covers the 29 hard-fenced names by construction, so the 38 in `NON_CANONICAL_ON_FEDORA` fall out of the gate. The daemon's `validate_action_platform` did not follow them out, because its early return needs `!action_requires_distro(action) && !is_mutating`, and every mutating one of those 38 fails the second half.

Measured, on `p384`, against an ineligible Debian 13 host:

```
$ out=$(bash /home/entropia/.local/state/sysknife-maint/ptcargo.sh t384d p384 bash -c "$(sed 's/--lib --locked/--bins --locked/' /tmp/sk384proof.sh)" 2>&1)
running 1 test
PROBE UfwEnable              action_requires_distro=false client=Ok
PROBE UfwAllow               action_requires_distro=false client=Ok
PROBE UfwLimit               action_requires_distro=false client=Ok
PROBE AppArmorEnforce        action_requires_distro=false client=Ok
PROBE Fail2banBanIp          action_requires_distro=false client=Ok
PROBE ConfigureFail2banJail  action_requires_distro=false client=Ok
PROBE SnapInstall            action_requires_distro=false client=Ok
PROBE AptInstall             action_requires_distro=true  client=Err
```

The same eight through the real `validate_action_platform`, called directly rather than reproduced, from a probe inserted into its own `mod tests` so `test_state` was in scope:

```
$ podman run --rm --network=none -v .../trees/iso384:/repo:z -v .../ctargets/iso384:/ctarget:z -v "$HOME/.cargo:/cargo:O" -v "$HOME/.rustup:/rustup:O" -w /repo -e CARGO_HOME=/cargo -e RUSTUP_HOME=/rustup -e CARGO_TARGET_DIR=/ctarget -e CARGO_NET_OFFLINE=true -e PATH=/cargo/bin:... docker.io/library/rust:1-slim bash -c "$(cat /tmp/sk384fix.sh)"
running 1 test
REAL-DAEMON UfwEnable              Err: cannot run UfwEnable on unsupported host Debian 13; see docs/distro-support.md
REAL-DAEMON UfwAllow               Err: cannot run UfwAllow on unsupported host Debian 13; see docs/distro-support.md
REAL-DAEMON UfwLimit               Err: cannot run UfwLimit on unsupported host Debian 13; see docs/distro-support.md
REAL-DAEMON AppArmorEnforce        Err: cannot run AppArmorEnforce on unsupported host Debian 13; see docs/distro-support.md
REAL-DAEMON Fail2banBanIp          Err: cannot run Fail2banBanIp on unsupported host Debian 13; see docs/distro-support.md
REAL-DAEMON ConfigureFail2banJail  Err: cannot run ConfigureFail2banJail on unsupported host Debian 13; see docs/distro-support.md
REAL-DAEMON SnapInstall            Err: cannot run SnapInstall on unsupported host Debian 13; see docs/distro-support.md
REAL-DAEMON AptInstall             Err: cannot run AptInstall on unsupported host Debian 13; see docs/distro-support.md
```

And the same probe against `main`, so this is a regression rather than a pre-existing gap:

```
$ bash -n /tmp/skmainprobe.sh && out=$(bash /home/entropia/.local/state/sysknife-maint/ptcargo.sh isomain origin/main bash -c "$(cat /tmp/skmainprobe.sh)" 2>&1); printf '%s\n' "$out" | grep -v '^time=' | tail -20
running 1 test
PROBE-MAIN UfwEnable              client=Err
PROBE-MAIN UfwAllow               client=Err
PROBE-MAIN UfwLimit               client=Err
PROBE-MAIN AppArmorEnforce        client=Err
PROBE-MAIN Fail2banBanIp          client=Err
PROBE-MAIN ConfigureFail2banJail  client=Err
PROBE-MAIN SnapInstall            client=Err
PROBE-MAIN AptInstall             client=Err
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 267 filtered out; finished in 0.00s
```

Seven of eight flip from `Err` to `Ok` on the client while the daemon keeps saying `Err`. Observable consequence: on a Debian 13 host, `sysknife "block that IP with the firewall"` now plans `UfwBanIp`-class actions, prints a preview, asks the operator to approve, takes the approval, and then the daemon refuses at execute. That is the failure class #181 was closed for, plus a paid model call spent on a plan that could never run.

Nothing covers it. `unsupported_distros_cannot_reach_mutation_paths` uses `AptInstall` and `AptUpgrade` only, and `supported_hosts_still_enforce_action_family` asserts the portable list `is_ok()` on an *eligible* Silverblue host, which is a different question.

The fix belongs in `sysknife-core::action_family`, since keeping the two consumers in parity is what the module exists for. One predicate that both sides call, covering the hard fences plus the two non-canonical lists, rather than three call sites each choosing their own set.

## Blocking 2: an unknown-family host is offered 38 actions the daemon will refuse

The doc comment on `available_on` says a detected `other` family "is filtered against **both** fences, matching the CLI routing guard: it refuses every family-specific action on such a host, so offering them invites a plan that is certain to be rejected after a paid call." The `DISTRO_FAMILY_OTHER` arm excludes the three hard fences and not `NON_CANONICAL_ON_FEDORA`:

```
running 1 test
PROBE-OTHER offered=38 of 38 NON_CANONICAL_ON_FEDORA names
PROBE-OTHER leaked: GetHostState SnapInstall SnapRemove SnapRefresh SnapHold SnapUnhold SnapList SnapInfo SnapRevert SnapClassicInstall UfwEnable UfwDisable UfwAllow UfwDeny UfwReset UfwStatus DistroboxList DistroboxCreate DistroboxRemove NetplanGetConfig NetplanApply NetplanSet NetplanGenerate AppArmorStatus AppArmorEnforce AppArmorComplain CloudInitStatus UbuntuInstallFlatpak UbuntuRemoveFlatpak UbuntuUpdateFlatpak UbuntuListFlatpaks Fail2banStatus Fail2banBanIp Fail2banUnbanIp ConfigureFail2banJail MultipassList UfwDeleteRule UfwLimit
```

All 38. On `main` these lived inside `DEBIAN_ONLY_ACTIONS`, which that arm does exclude, so an Arch or openSUSE host was offered none of them. `Other{..}.is_supported()` is false, so the daemon refuses every mutating one.

`an_unrecognised_family_is_offered_no_family_specific_action` cannot catch it:

```rust
for name in FEDORA_ONLY_ACTIONS.iter().chain(DEBIAN_ONLY_ACTIONS.iter()) {
```

It iterates the two lists that shrank. Adding `&& !NON_CANONICAL_ON_FEDORA.contains(&action)` to that arm and extending the test's iteration to the new lists closes both halves.

## The test count is right, and I am withdrawing what I said last time

I told you the baseline was mine to move and that you were right to leave it. You moved it, and I am not going to make you undo that, because the number is machine-verified and my objection was procedural. The `rust` job re-runs the suite and fails on any mismatch, and it is green on this head. I checked the coupling anyway:

```
$ a=$(git diff $(git merge-base origin/main p384)..p384 | grep -cE '^\+[[:space:]]*#\[(test|tokio::test)\]')
$ r=$(git diff $(git merge-base origin/main p384)..p384 | grep -cE '^-[[:space:]]*#\[(test|tokio::test)\]')
$ echo "added=$a removed=$r net=$((a-r))"
added=5 removed=0 net=5
$ git show p384:README.md | grep -n '1,85'
317:**1,850 Rust tests and 72 frontend tests** form the current deterministic
$ git show p384:docs/introduction.md | grep -n '1,85'
144:190 typed actions · 1,850 Rust tests + 72 frontend tests · MIT
$ git show p384:docs/distro-support.md | grep -n '1,85'
85:The deterministic workspace baseline is 1,850 Rust tests plus 72 frontend
$ git grep -n '1,845' p384 -- '*.md' '*.json'
(no output)
```

+5 net test functions, 1845 → 1850, all three prose figures moved, no stale figure anywhere. Nobody hand-edited a number here. For what it is worth, schema v2 has no `measured_at` or `commit` field to check, since `ef8fe45` removed them two commits before your base.

## Not blocking

Two clauses cannot be reached. `propose_plan.rs:497` and `:503` each carry `!UBUNTU_ONLY_ACTIONS.contains(&action)`, and the early return at `:481` already returns `false` for those whenever `hint.id != "ubuntu"`. `runner.rs:56` only ever sets `id = "ubuntu"` for `DistroId::Ubuntu`, whose family is Debian, and `distro.rs:397` routes `ID=ubuntu` to `detect_ubuntu` so `Other{id}` can never carry it. So no hint reaches the Fedora or Other arm with `id == "ubuntu"`, and deleting either clause fails no test. `CLAUDE.md` bars unreachable code; your call whether to drop them or drop the early return and keep the per-arm exclusions.

`NON_CANONICAL_ON_DEBIAN_HOST` is a hand-copied strict subset of `NON_CANONICAL_ON_FEDORA`:

```
|NON_CANONICAL_ON_DEBIAN_HOST| = 14; names not also in NON_CANONICAL_ON_FEDORA:
  (none above = strict subset, duplicated by hand)
```

Nothing pins the containment, so an edit to one drifts from the other in silence. One assertion, or derive the subset.

The new prompt path has no coverage in the test that claims to cover every render. `prompt_risk_labels.rs` builds its Debian hint with `id: "ubuntu"`:

```rust
let debian = DistroHint {
    id: "ubuntu".into(),
```

So `render_debian_prompt`'s new `hint.id != "ubuntu"` branch is never rendered there, while that helper's comment says "Every distro render, so a per-family table cannot drift on its own". Adding an `id: "debian"` entry costs one struct literal. `CLAUDE.md` also asks for the E2E story suite against a prompt change; the reduced prompt drops `DEBIAN_RISK_TABLES` and the apt selection rules and leans on the schema, and whether the model still plans correctly from it is unmeasured. I am not blocking on a VM run you may not have, but the `rendered_prompts()` entry I would like.

`dispatcher.rs:966` still says "Both gates below apply to family-tagged actions whether or not they read", and 11 of the de-fenced actions are `RiskLevel::Low` and therefore `Observer`, so they now take the `:962` early return and execute with no distro detected. I checked that no Low-risk *mutating* action escaped: `AptUpdate` stays hard-fenced and `a_low_risk_action_that_mutates_is_not_treated_as_a_read` still pins it. The widening is real but read-only, and it is unmentioned in the CHANGELOG and in `docs/action-compatibility.md`.

`UbuntuCore` gets the catalogue inverted: `id = "ubuntu-core"` with family Debian means the planner withholds Snap and Netplan, which is the only package tooling Ubuntu Core has, while still offering the 17 apt actions it does not have. `is_supported()` is false so mutations are refused anyway, so this is direction rather than exposure.

## What I need

The two blocking items, and I would take them as one shared predicate in `action_family.rs` rather than two local patches, plus a test on each side that fails without it. The `rendered_prompts()` entry if you are pushing anyway.

Everything else on this list is yours to take or leave, and none of it is worth another round. `docs/action-compatibility.md` is a good addition and the CHANGELOG note on the public API break is exactly right.

This is a large, careful redesign and the set arithmetic held under every check I could think to run. Both blocking findings are the same shape: a consumer keyed on the new narrow predicate where it needed the old wide one. That is the predictable cost of splitting a list three ways, and it is much easier to fix than to have found.

## Next

You hold #362 and #237, so nothing new from me tonight. When they land, `#397` is waiting for you and I said as much on #383.

One coordination note: I have offered `#239` (nftables vocabulary) to @mikevillari and told them your PR lands first, since a new action family has to be placed in the lists you are re-partitioning. Nothing for you to do; you should not be surprised by a PR arriving on top of yours.

vladimirrott (dismissed):
Approving at `b6f76a36`. Both blocking findings from my last pass are gone, and the
thing I want to single out is how you made the boundary tests survive their own
lists being edited. `ubuntu_identity_is_stricter_than_debian_family` names
`AddPpa`, `ProAttach`, `LivepatchStatus` and `CheckPendingReboot` as literals and
asserts `action_requires_distro` on each, so emptying `UBUNTU_ONLY_ACTIONS`
fails it. I proved that below, because the list-driven test in the CLI passes
vacuously under the same mutation.

## The partition, recounted

I pulled the constants out of both trees and did the arithmetic rather than
reading the diff. The scratch script extracts every `pub const … &[&str]` and
compares the two versions:

```
$ python3 /tmp/lists.py
== main lists ==
  FEDORA_ONLY_ACTIONS: 25
  NON_CANONICAL_ON_DEBIAN: 5
  DEBIAN_ONLY_ACTIONS: 67
== p384 lists ==
  FEDORA_ONLY_ACTIONS: 25 (unique 25)
  NON_CANONICAL_ON_DEBIAN: 5 (unique 5)
  DEBIAN_ONLY_ACTIONS: 19 (unique 19)
  UBUNTU_ONLY_ACTIONS: 10 (unique 10)
  NON_CANONICAL_ON_DEBIAN_HOST: 14 (unique 14)
  NON_CANONICAL_ON_FEDORA: 38 (unique 38)

|M|=67 |D|=19 |U|=10 |N|=38  |D|+|U|+|N|=67
pairwise intersections D∩U,D∩N,U∩N: [] [] []
M \ (D∪U∪N)  (capability that left the Fedora exclusion set): []
(D∪U∪N) \ M  (newly withheld from Fedora): []
FEDORA_ONLY identical: True 25
Ubuntu-catalogue exclusion set main: 30  pr: 30 same: True
Fedora-catalogue exclusion main: 67  pr: 67 same: True
```

D ⊔ U ⊔ N = M with |M| = 67, so the split moves names between lists and invents
none. `FEDORA_ONLY_ACTIONS` is byte-identical, which is what `CLAUDE.md` asks
for.

## The catalogues, measured from the rendered schema

A probe inside `prompt.rs` builds the tool definition and the prompt for each
hint, reads the `action_name` enum out of the JSON schema, and cross-references
the CLI verdict for the matching `DistroId`:

```
$ ./ptcargo.sh t384f p384 cargo test -p sysknife-brain --lib --offline zzz_probe_prompt_vs_schema -- --nocapture
PROMPTSIZE	fedora	offered=123	prompt_bytes=35311
PROMPTSIZE	ubuntu	offered=160	prompt_bytes=42185
PROMPTSIZE	debian	offered=136	prompt_bytes=29146
PROMPTSIZE	arch	offered=93	prompt_bytes=29270
PROMPTSIZE	nohint	offered=190	prompt_bytes=29270
```

| hint | offered on main | offered here | offered but refused by the CLI, main | here |
|---|---|---|---|---|
| ubuntu | 160 | 160 | 0 | 0 |
| fedora | 123 | 123 | 0 | 0 |
| debian | 160 | 136 | 67 | 43 |
| arch (other) | 98 | 93 | 0 | 0 |

Ubuntu and Fedora keep the catalogue they had, to the action. Your claim that the
default catalogues are unchanged holds as set equality, not approximately. On a
Debian host the number of actions offered to the model that the CLI then refuses
before approval drops from 67 to 43.

## The client/daemon asymmetry I blocked on last time

Two probes, one calling `check_action_distro` in the CLI and one calling the real
`validate_action_platform` from inside its own module, over all 190 catalogue
names and seven `DistroId` values, on both trees:

```
$ ./ptcargo.sh t384f p384 cargo test -p sysknife-cli --bins --offline zzz_probe_client_matrix -- --nocapture
$ ./ptcargo.sh t384f p384 cargo test -p sysknife-daemon --lib --offline zzz_probe_daemon_matrix -- --nocapture
cells compared: 1520
|asymmetry main| = 476   |asymmetry p384| = 450
NEW asymmetry introduced by p384: 0
asymmetry REMOVED by p384: 26
   - Debian13 ConfigureFirewall
   - Fedora41 CreateToolbox
   - NoHint UfwStatus
   ... (26 total)
client stricter than daemon (safe direction): 65
```

Zero new cells where the client permits what the daemon refuses, and 26 fewer
than main. The 450 that remain are all on main too: generic mutating actions
(`CreateUser`, `RebootSystem`) on an ineligible host, which `check_action_distro`
never claimed to cover. That gap is mine to file, not yours.

The daemon widening is the 38 portable names on Fedora Atomic, and it is
symmetric with the client:

```
   Silverblue41   UfwEnable                  main=ERR p384=OK
   Silverblue41   SnapInstall                main=ERR p384=OK
   Silverblue41   GetHostState               main=ERR p384=OK
   ... (38 total)
```

I checked every one of the 38 against its own `ActionSpec` mechanism, and none of
them drives apt or dpkg:

```
uses apt/dpkg despite being called portable: []
```

`GetHostState` is `hostnamectl status`, so calling it portable is right and your
test comment says why.

## Mutation proofs

Eleven, each restored before the next, in rootless podman with no network. Every
one went red on exactly the test the change was written for.

| # | mutation | result |
|---|---|---|
| M1 | delete the `UBUNTU_ONLY_ACTIONS` identity check in `check_action_distro` | `ubuntu_only_actions_require_identity_and_eligibility` FAILED |
| M2 | `action_requires_supported_host` → `action_requires_distro` | `portable_tools_require_supported_hosts_before_approval` FAILED: `SnapInstall must be refused before approval on Debian 13` |
| M3 | drop the `hint.id != "ubuntu"` arm in `available_on` | `debian_host_does_not_inherit_ubuntu_tools_or_preferences` FAILED: `Debian must not offer AddPpa` |
| M4 | drop the `NON_CANONICAL_ON_DEBIAN_HOST` arm | same test FAILED: `Debian must not offer SnapInstall` |
| M5 | revert the `DISTRO_FAMILY_OTHER` arm to main's two fences | `an_unrecognised_family_is_offered_no_family_specific_action` FAILED: `offered SnapInstall` |
| M6 | `if hint.id != "ubuntu"` → `if false` in `render_debian_prompt` | `debian_prompt_does_not_inherit_ubuntu_examples` FAILED: `Debian prompt advertises AddPpa` |
| M7 | move `AddPpa` back into `DEBIAN_ONLY_ACTIONS` | `ubuntu_identity_is_stricter_than_debian_family` FAILED |
| M8 | `if !action_matches_distro(…)` → `if false` in the daemon | `supported_hosts_still_enforce_action_family` FAILED |
| M9 | `UBUNTU_ONLY_ACTIONS = &[]` | core test FAILED; the CLI list-driven test passed |
| M10 | add `SnapInstall` to `DEBIAN_ONLY_ACTIONS` | `family_fence_agrees_with_each_action_s_mechanism` FAILED |
| M11 | add `UfwEnable` to `UBUNTU_ONLY_ACTIONS` | same test FAILED |

M9 is the one worth keeping in mind:

```
$ python3 /tmp/mut384.py M9
applied M9 to family
test action_family::tests::ubuntu_identity_is_stricter_than_debian_family ... FAILED
thread 'action_family::tests::ubuntu_identity_is_stricter_than_debian_family' (179) panicked at crates/sysknife-core/src/action_family.rs:341:13:
AddPpa
test distro_routing::tests::ubuntu_only_actions_require_identity_and_eligibility ... ok
```

An emptied list satisfies every test that iterates it. Under M9 only your literal-anchored test in `sysknife-core` fails, which is why I
name it here rather than leave it in a table.

M10 and M11 confirm the reverse direction you added to
`family_fence_agrees_with_each_action_s_mechanism` still bites after the token
lists were re-cut:

```
  SnapInstall: Debian fence exceeds its mechanism (sudo sh -c snap install --channel=stable firefox && snap refresh --hold firefox)
  UfwEnable: Ubuntu fence disagrees with mechanism (sudo ufw --force enable)
```

## What stops the merge, and it is not yours

`tests/evidence/workspace-tests.json` reads 1,850 here and 1,846 on #402, and
both are correct against today's `main`. Whichever lands first moves the number
under the other. When yours is second, that is
`UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` plus the three prose figures in
the same commit, nothing more.

My own merge gate cannot issue a receipt for this PR, which is my tooling and not
your diff:

```
$ maintainer-merge verify 384 b6f76a36bf986a74444ae3fb13657428cc748697 "ubuntu_only_actions_require_identity_and_eligibility" "/requires Ubuntu itself/d"
maintainer-merge: no single suite in .../verify.d covers every changed path:
    ...
    tests/evidence/workspace-tests.json	no suite covers this
```

Every Rust change that adds a test has to move that artifact, so the gate refuses
every well-formed Rust PR in this repo. I am fixing that on my side.

## Three optional notes

None of these block anything.

1. `shared_examples_demonstrate_only_cross_distro_actions` iterates five lists and
   omits `NON_CANONICAL_ON_FEDORA`. Nothing leaks today, verified: my
   prompt-versus-schema probe found no Fedora-prompt action outside the offered
   enum. If a future example ever names `UfwStatus` or `SnapList`, the Fedora
   prompt would advertise an action the Fedora schema withholds, and that test is
   where I would expect to catch it. One more chained list closes it.
2. The one prompt leak the probe did find is on `main` too, so it is mine:
   `PROMPTLEAK arch GetSystemState`. The generic prompt names `GetSystemState`
   while an unknown-family schema does not offer it.
3. On a Debian host the catalogue still offers the 19 apt actions that
   `is_supported()` refuses before approval. Your `docs/action-compatibility.md`
   states the reasoning, and #238 is where that gets settled.

## Next

#238 asks for a version floor and a stated reason inside `is_supported`, which is
the other half of what you built here: you separated identity from mechanism, and
#238 turns eligibility into something with a reason attached. It is free, it is
yours if you want it, and I will hold it for you as soon as this lands. Say the
word and I will assign it.

Your diff reasons about PPAs, Pro, livepatch and the reboot sentinel with the kind
of precision that comes from running these machines. If you have an Ubuntu server
and a Debian box to hand, the read-only surface (`sysknife ask "what is pending on
this host"`, `GetHostState`, `UfwStatus`) is the part I would most like a second
pair of eyes on, and the Debian side is exactly the case #238 is about. No
obligation, and none of it is a condition of merging.

vladimirrott:
Correction to my own review, on the commands rather than the findings. I quoted
three of them with the pipelines stripped, and two of the output blocks came from
a comparison script rather than from the command printed above them. Every number
stands; the transcript reads differently from what I posted, and I hold myself to
quoting what ran.

The probes, as run:

```sh
./ptcargo.sh t384f p384 cargo test -p sysknife-brain --lib --offline zzz_probe_prompt_vs_schema -- --nocapture 2>&1 | grep -E '^PROMPT|^error|test result' | head -60
./ptcargo.sh t384f p384 cargo test -p sysknife-cli    --bins --offline zzz_probe_client_matrix   -- --nocapture 2>&1 | grep -E '^CLIENT' > /tmp/client384.tsv
./ptcargo.sh t384f p384 cargo test -p sysknife-daemon --lib  --offline zzz_probe_daemon_matrix   -- --nocapture 2>&1 | grep -E '^DAEMON' > /tmp/daemon384.tsv
```

The same three ran against `origin/main` in a second tree (`tmain384`). The
`cells compared: 1520`, `|asymmetry main| = 476`, `|asymmetry p384| = 450`,
`NEW asymmetry introduced by p384: 0` and `asymmetry REMOVED by p384: 26` lines
are printed by a Python script that loads the four TSVs and takes the set
differences, not by the cargo runs. The offered-versus-refused table comes from
the same pairing of `/tmp/offered384.tsv` with `/tmp/client384.tsv`.

The mutations ran as `python3 /tmp/mut384.py <tag>` followed by one
`./ptcargo.sh m384 p384 cargo test -p <crate> …` per tag, with a restore between
each; the failing assertions are quoted verbatim in the table.

`ptcargo.sh` exports a git ref into a scratch tree and runs cargo inside
`docker.io/library/rust:1-slim` under rootless podman with `--network=none` and
`CARGO_NET_OFFLINE=true`, so nothing from this branch ran on my host.

vladimirrott:
@QinXi-ai I resolved the conflict on your branch and pushed to it, so you are not chasing a number you never moved.

Every conflict was bookkeeping: `CHANGELOG.md`, the documented Rust test count in `README.md`, `docs/introduction.md` and `docs/distro-support.md`, and `tests/evidence/workspace-tests.json`. `crates/sysknife-brain/src/planner.rs` auto-merged. Seven pull requests landed on main since you opened this and each one moved the baseline, so with strict branch protection every merge hands the next open PR a conflict on a count its author did not choose. That is my merge order costing you an evening, and it is mine to fix.

What I changed:

- Both CHANGELOG sections kept, yours under `### Changed` and main's under `### Fixed`.
- The baseline regenerated with the repo's own command rather than typed in by hand:

```
     Summary [   8.718s] 1851 tests run: 1851 passed, 6 skipped
test_baseline: recorded 1851 rust tests in tests/evidence/workspace-tests.json
```

- The three prose claims moved to 1,851 to match, and both claim gates agree:

```
$ bash scripts/check_public_claims.sh
```

Your production contribution is unchanged, and I checked instead of assuming. Same diff, before and after, first 16 hex of the sha256 with the blob index headers stripped:

```
$ git diff cce5c7a..b6f76a36 -- 'crates/*/src/*' 'apps/*/src/*' 'packaging/*' '.github/*'
4b1bf2c46073faa4
$ git diff origin/main..19ca336a -- 'crates/*/src/*' 'apps/*/src/*' 'packaging/*' '.github/*'
4b1bf2c46073faa4
```

The guard from my review still bites on the merged tree. Moving `AddPpa` back into `DEBIAN_ONLY_ACTIONS`:

```
$ cargo test -p sysknife-core --lib ubuntu_identity_is_stricter_than_debian_family --locked
thread 'action_family::tests::ubuntu_identity_is_stricter_than_debian_family' panicked at
crates/sysknife-core/src/action_family.rs:352:13
test result: FAILED. 0 passed; 1 failed; 97 filtered out
```

and restored, `1 passed`. A conflict resolution can neuter a guard while every test still passes, which is why I re-ran that one rather than trusting the green board.

The push dismissed my approval, as `dismiss_stale_reviews` requires. I will re-approve at `19ca336a` once `ci` and `e2e` are both green and merge from there. Nothing is waiting on you, unless you disagree with how I merged the CHANGELOG, in which case push over it and I will re-check at your head.

One thing this unblocks: #238, Debian eligibility and the version floor, says in its own body that #237 blocks it. It is the direct next step of what you just built, and it is yours if you want it once this lands. I will hold it.

vladimirrott (approved):
Re-approving at `19ca336a`, the head I resolved and re-verified.

`ci` and `e2e` both completed green, twelve checks, nothing pending. The production diff is byte-identical to the one I approved at `b6f76a36`, and the guard still bites on the merged tree: moving `AddPpa` back into `DEBIAN_ONLY_ACTIONS` fails `ubuntu_identity_is_stricter_than_debian_family` at `action_family.rs:352`, and passes again restored. Both runs are quoted in the comment above.

Merging from here.

vladimirrott:
Merged as `19ca336a`, closing #237.

Thank you for this one. It is the largest re-partitioning anybody outside has done to this catalogue, and the part I want to name is the test you wrote rather than the fences you moved. `action_family.rs:329` lists the expected boundaries as literals instead of iterating the production arrays, so the assertions stop moving together with the thing they assert. That is the fix I asked for and you went past it: `CheckPendingReboot` and the `linuxmint` derivative case were not in what I blocked on.

The proof I merged on, re-run at the resolved head rather than carried over from the earlier one:

```
$ cargo test -p sysknife-core --lib ubuntu_identity_is_stricter_than_debian_family --locked
test result: ok. 1 passed; 0 failed; 97 filtered out

# AddPpa moved back out of UBUNTU_ONLY_ACTIONS into DEBIAN_ONLY_ACTIONS
panicked at crates/sysknife-core/src/action_family.rs:352:13
test result: FAILED. 0 passed; 1 failed; 97 filtered out
```

`ci` and `e2e` both green, twelve checks. CHANGELOG entries were already in your branch and I kept them under `### Changed`, including the note that `DistroHint` carrying an `id` is a public Rust API change and moves the middle digit.

Two things I left as follow-ups rather than blocking on:

- `dispatcher.rs:966` still says "Both gates below apply to family-tagged actions whether or not they read", and eleven of the de-fenced actions are `RiskLevel::Low` and therefore `Observer`, so they take the `:962` early return. The widening is real and read-only, and I checked that no Low-risk mutating action escaped, `AptUpdate` included. The comment is now wrong and `docs/action-compatibility.md` does not mention it.
- `UbuntuCore` gets the catalogue inverted: `id = "ubuntu-core"` with family Debian withholds Snap and Netplan, which is its only package tooling, while still offering seventeen apt actions it does not have. `is_supported()` is false so nothing can execute, so this is direction rather than exposure.

Neither is yours to fix unless you want them.

Next, and it is the direct continuation of what you just built: **#238**, Debian eligibility with a version floor of 12 and a reason in `is_supported` where there is none. Its body names #237 as the blocker, so you just unblocked it. The single line that refuses Debian is `distro.rs:252`, and the doc comment above it justifies pre-20.04 Ubuntu and `UbuntuCore` and says nothing at all about Debian; it sits in that arm because nothing has needed it elsewhere. You now know that file better than I do.

One decision in it is genuinely open and I would rather you made it than inherited it. `version` is `Option<u32>` because testing and sid publish no `VERSION_ID`, and Debian's security team does not support sid. Refusing `None` is consistent with the floor's own rationale and is what I would do, but it turns the `Debian stable/testing` row in the support matrix into stable-only and that row then needs rewording. Say which way you want it and the rest is mechanical.

Yours if you want it. I will hold it either way, and no is a complete answer.

Unrelated, and easy to decline. #256 and now #237 say you read this catalogue against real distributions rather than against the docs. `sysknife doctor` and `sysknife history` change nothing on a host, and what they get wrong on a machine I do not own is worth more to me than anything I can test here. Not a condition of anything.

QinXi-ai:
Follow-up is now in #414: Ubuntu Core no longer inherits apt/host-policy actions from its Debian family tag, and the existing catalogue test covers every action for that hint. The production dispatcher comment and portable-read documentation were already corrected in the merged base; I corrected the remaining stale test comment as well.

The other two follow-ups are submitted separately: Debian eligibility in #412 (current CI green; live Debian validation still outstanding) and the read-only scope of #239 in #413. The Debian VM/provider question is in #238 so eligibility is not confused with measured live support.

PR #386, opened 2026-09-07
fix(evidence): make story metadata single-source

### LINKED ISSUE #252: Two parsers read the story header, and nine tag spellings pass unchecked

bferanmi806-sketch:
## Summary

- derive story family metadata from `run-stories.sh --metadata` instead of maintaining a second Story-header parser in Python
- close the canonical story-tag vocabulary so unknown spellings fail instead of silently changing families
- make the Python evidence consumer fail closed on runner errors and unusable metadata
- add mutation coverage for unknown tags, misplaced headers, and broken metadata derivation

Closes #252.

## Why

After #370, the Bash and Python Story-header parsers agreed on reading line 2 and on rejecting malformed headers, but they were still two independent implementations of the same grammar.

Both also classified families from the substring `ubuntu`.

That left this failure mode:

```text
ubuntu -> ubunut
```

A syntactically valid typo could silently move a story from the Ubuntu family into the atomic family, changing the derived suite sizes used by the evidence claim guard.

The Python side now consumes the metadata emitted by the runner instead of parsing Story headers itself.

The path is:

```text
story files
    -> run-stories.sh --metadata
    -> canonical (id, family, title) rows
    -> check_evidence_claims.py
```

There is no Story-header regex left in `check_evidence_claims.py`.

## Tag vocabulary

I kept the current vocabulary closed rather than redesigning the header schema in this PR.

In particular, I did **not** normalize `ubuntu, low-risk` to `ubuntu, read-only`.

I audited all 14 `ubuntu, low-risk` stories. Thirteen are read-only query scenarios, but story 55 asserts `AptUpdate` at low risk. `AptUpdate` changes the host's package metadata, so low-risk and read-only are not equivalent concepts in the current suite.

`ubuntu, low-risk` therefore remains accepted legacy metadata.

Atomic `destructive` is also preserved as-is; it describes the gated host-changing story subset rather than being an alias for ActionSpec high risk.

## Fail-closed behavior

The Python consumer now rejects:

* a missing metadata runner
* a non-zero runner exit
* empty metadata from a non-empty story tree
* malformed TSV rows
* non-numeric story IDs
* unknown family values
* empty titles

A broken metadata source cannot silently become an empty family set and pass a claim check.

## Mutation proof

Using `story-63.sh`:

### Unknown tag

Changed:

```text
ubuntu, read-only
```

to:

```text
ubunut, read-only
```

The canonical runner failed:

```text
ERROR: unrecognized story tags in .../story-63.sh: ubunut, read-only
```

The Python evidence guard also failed through the canonical metadata path:

```text
could not derive story families: ... --metadata exited 1:
ERROR: unrecognized story tags in .../story-63.sh: ubunut, read-only
```

After reverting the mutation, metadata returned all 133 rows and the evidence guard passed.

### Header moved to line 3

Moving the Story header from line 2 to line 3 made `run-stories.sh --metadata` fail with an unparseable-header diagnostic naming `story-63.sh`.

The Python evidence guard propagated the same canonical-parser failure.

Reverting restored both consumers to green.

## Validation

Passed:

```text
bash tests/e2e/story-metadata.test.sh
  Story metadata derived cleanly: 133 stories (79 ubuntu, 54 atomic).

python3 scripts/check_evidence_claims.py .

bash tests/release/public-claims.test.sh

bash tests/e2e/story-runner-verdicts.test.sh
  contract passed across all three production runners.
```

Also passed:

```text
bash -n tests/e2e/run-stories.sh
bash -n tests/e2e/story-metadata.test.sh
bash -n tests/release/public-claims.test.sh
python3 -m py_compile scripts/check_evidence_claims.py
git diff --check
```

`scripts/ci-local.sh` was attempted but stopped at its tool preflight because this environment does not have `cargo` or `node` installed:

```text
ci-local: FATAL missing required tool(s): cargo node
```

No Rust files are changed. I have not substituted an estimated workspace result for the gate that could not run.

## Result

The current tree remains:

```text
79 ubuntu
54 atomic
133 total
```

A misspelled Ubuntu family tag can no longer silently alter that partition.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -166,8 +166,10 @@ git diff --name-only upstream/main... | grep '\.rs$'   # empty means the gate do
 For such a diff the gate is the Node suite for the package you touched,
 `scripts/check_evidence_claims.py`, and CI on Ubuntu. Run the claim screen even
 for a JavaScript change: `packages/setup/index.js` is in `CLAIM_FILES`, and the
-screen is pure Python with no subprocess calls, so it runs anywhere Python does.
-Say which of the three you ran.
+screen remains the claim-screen entry point. Story-family derivation delegates to
+`tests/e2e/run-stories.sh --metadata`, so story-related claim checks require
+Bash and the repository's normal POSIX-style environment. Say which of the
+three you ran.
 
 **Which release your change lands in.** While SysKnife is in the `0.y` series, a
 change that breaks a consumer moves the middle digit and everything else moves the
--- a/scripts/check_evidence_claims.py
+++ b/scripts/check_evidence_claims.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 import json
 import re
+import subprocess
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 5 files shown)


vladimirrott:
Four PRs now, and #386 continues the thread you have been pulling on: the story evidence should derive from the tree, not from a hand-kept list. That is exactly the instinct #229 and #370 needed.

Three more in the same vein, reserved for you:

- **#233** — the story coverage map: 64 Ubuntu and cross-family actions have no story. The measurement you have been building toward.
- **#278** — nothing checks that `workspace-tests.json` was measured at the commit it claims. The evidence-integrity gap under all of this.
- **#219** — E2E story coverage for the fail2ban family, one of the concrete gaps #233 names.

They hold the `claimed` label until you post on a thread, since GitHub blocks assigning you before then. Reply and it is yours. #386 is in review.

vladimirrott (approved):
Approving at `9a8aaf7b`. This removes the second answer, which is the right fix for #252 and a harder one to write than a third parser would have been.

I measured the failure you describe rather than taking it on trust. Same one-character edit to `story-100.sh`, once on each tree:

```
main    run-stories.sh --metadata   rc=0   story 100 family: atomic
        counts with the typo:  ubuntu=78 atomic=55
        counts clean:          ubuntu=79 atomic=54

#386    run-stories.sh --metadata   rc=1
        ERROR: unrecognized story tags in .../story-100.sh
```

So today a typo does not only shift the derived suite sizes, it moves a story into the family that runs by default under `SYSKNIFE_ALLOW_DESTRUCTIVE=1`. Silently. Your PR turns that into an exit 1 naming the file.

`story-metadata.test.sh` passes on your head, deriving 133 stories (79 ubuntu, 54 atomic).

Three things I want to name, because they are the parts a reviewer skims:

- **You guard your own mutation.** `report "unknown-tag mutation did not apply — the assertion below is vacuous"` is the check almost every mutation harness omits, including one of mine earlier today, where a pipe to `tail` handed me `tail`'s exit code and made three failing mutations read as passes. A harness that cannot tell "the guard held" from "the mutation never landed" reports the same thing for both.
- **The five negative controls on the consumer** are the right set: missing runner, non-zero exit, empty metadata over a non-empty tree, malformed row, unknown family. The empty-metadata one matters most, since "zero stories" is how a broken parser would otherwise let every count-consuming guard pass over nothing.
- **You did not normalise `ubuntu, low-risk` away.** The comment explains that story 55 asserts risk `low` for `AptUpdate`, which mutates host state, so it is not a misspelling of `ubuntu, read-only`. Closing a vocabulary is exactly where somebody tidies a legacy value into the wrong bucket, and you stopped and wrote down why you did not.

Nothing blocking, and no new issue from me: you already have #278, #233 and #219 going, and I would rather you finish those than carry more.

Since your work here keeps landing on the evidence-and-claims surface: if you run fleets of anything, `sysknife audit verify` is the operator-facing version of what you are building, a signed action chain that exits non-zero when the record and the reality disagree. Read-only, and worth a look on a box you already trust.

Issue #387, opened 2026-09-07
no-secrets: deleting the Slack pattern leaves the gate green, and its tree scan passes on an empty file list


vladimirrott:
`tests/release/no-secrets.test.sh` runs in `docs-and-hygiene`, one of the five
required checks. It guards `scripts/check_no_secrets.sh`, the scanner
`.githooks/pre-commit` runs over staged content before a commit object exists.

Most of the guard holds up. Eight of the nine patterns in the scanner have a
positive case, the fixture side is covered, and the "never echo the credential"
assertion from #228 carries its own mutation proof. Three gaps sit in the parts
nobody exercised, and a fourth turned up while measuring them.

Everything below ran against a throwaway clone of `049bc1f`, one mutation at a
time, restored between each. The harness is two lines:

```
$ cat /tmp/ns-audit/run.sh
#!/usr/bin/env bash
# usage: run.sh <label>
cd /tmp/ns-audit/tree
out="$(bash tests/release/no-secrets.test.sh 2>&1)"; rc=$?
printf '=== %s === rc=%d\n%s\n' "$1" "$rc" "$out"
```

## 1. The Slack pattern has no positive case, so deleting it is silent

`check_no_secrets.sh` carries nine `PATTERNS` entries. `POSITIVES` in the guard
synthesises seven keys and an eighth case covers AWS. Slack is the one left
over.

```
$ sed -i '/"Slack token:xox\[baprs\]-\[A-Za-z0-9-\]{20,}"/d' scripts/check_no_secrets.sh
$ /tmp/ns-audit/run.sh M1-slack-pattern-deleted
=== M1-slack-pattern-deleted === rc=0
ok: catches real-shaped credentials, ignores this repo's fixtures
ok: the whole tracked tree scans clean, and findings never echo the secret
```

Delete any other pattern and the guard names it. That calibration is what makes
the green above worth reading:

```
$ sed -i '/"Groq:gsk_\[A-Za-z0-9\]{40,}"/d' scripts/check_no_secrets.sh
$ /tmp/ns-audit/run.sh M2-groq-pattern-deleted
=== M2-groq-pattern-deleted === rc=1
FAIL: a gsk_AAAA… credential (56 chars) was NOT caught

$ sed -i '/"AWS access key:AKIA\[0-9A-Z\]{16}"/d' scripts/check_no_secrets.sh
$ /tmp/ns-audit/run.sh M3-aws-pattern-deleted
[...]

PR #389, opened 2026-09-07
fix(e2e): check story tag vocabulary per word, not per whole string

### LINKED ISSUE #390: run-stories.sh whitelists whole tag strings, so a valid new combination fails CI
`tests/e2e/run-stories.sh` closes its tag vocabulary by matching the whole tag string against ten exact combinations:

```sh
case "$_tags" in
    ""|"destructive"|"read-only"|"hard compound"|\
    "ubuntu, low-risk"|"ubuntu, read-only"|"ubuntu, medium-risk"|\
    "ubuntu, high-risk"|"ubuntu, rejection"|"ubuntu, compound") ;;
    *)
        printf 'ERROR: unrecognized story tags in %s: %s\n' "$_story_file" "$_tags" >&2
        exit 1
        ;;
esac
```

That closed the hole it was written for (#252): a typo like `ubuntu` → `ubunut` used to exit 0 and silently move a story from the ubuntu family into the atomic family, which runs by default under `SYSKNIFE_ALLOW_DESTRUCTIVE=1`, and shifted every derived suite count.

It is brittle in the direction this tree grows. The ten arms are combinations, not words, so a story tagged `(ubuntu, destructive)` fails CI even though both words are known and the pairing is meaningful. Adding any new valid combination means editing a `case` arm, and the failure a contributor sees first is their own correct story being rejected.

## What to do

Replace the whole-string `case` with a per-word check: split `$_tags` on commas, trim each word, and require every word to appear in a `STORY_TAG_VOCAB` array. An unrecognised **word** exits 1, naming the file, the word, and the vocabulary. Any combination of known words is accepted.

The vocabulary is the nine distinct words in the tree today: `ubuntu`, `destructive`, `medium-risk`, `high-risk`, `read-only`, `low-risk`, `rejection`, `compound`, `hard compound`.

Keep `low-risk` and `read-only` distinct. Story 55 asserts risk `low` for `AptUpdate`, which writes the package index, so folding it into `read-only` would misname it. #252 asked for that to be stated rather than resolved, and this issue does not resolve it either.

## Acceptance

- A story tagged with any combination of known words passes; add `(ubuntu, destructive)` to a fixture and it is accepted.
- A misspelled word in any position exits 1 naming the word, the file and the vocabulary.
- `tests/e2e/story-metadata.test.sh` proves both consumers reject it: the Bash runner and `scripts/check_evidence_claims.py`, each naming the file. A header shifted off line 2 must also be rejected by both.
- The mutation block guards against its own mutation failing to apply.

## Context

This is the half of PR #389 that survives #386. #389 was written against #252 and arrived after I merged a different implementation of the same issue and closed it, which was my scheduling failure rather than anything about the work. The per-word vocabulary and the both-consumers mutation block in that PR are better than what landed, and this issue exists so they have somewhere to land with credit.

Georgefifth:
Follow-up to vladimirrott's review here: rebased onto `main` (post-#386), keeping only what the merged PR does not already cover — the per-word vocabulary and the both-consumers mutation cases. Now against #390, which is the tag-vocabulary half re-opened and assigned.

## What changed vs. what landed in #386

`#386` checks the whole tag string against ten exact combinations. This replaces the `case` block with a nine-word `STORY_TAG_VOCAB`, checked per comma-separated word (trimmed). Both catch `ubunut` and a misspelled risk word — equivalent on the #252 bug. They differ afterwards: a new story tagged `(ubuntu, destructive)` is a combination that does not exist in the tree today and is perfectly meaningful; under the whole-string whitelist it fails CI until somebody edits a `case` arm, under this it just works.

## Dropped (already covered by #386)

- The `STORY_HEADER` regex removal in `check_evidence_claims.py` — `story_family_sizes()` on `main` already delegates to `run-stories.sh --metadata`.
- The runner copy in the `public-claims` fixture — already on `main`.
- The single-consumer unknown-tag and line-3 mutations — already in `story-metadata.test.sh` on `main`.

## Kept / added

- `STORY_TAG_VOCAB` per-word check replacing the `case` block. The diagnostic names the file first, then the word (`ERROR: <file>: unrecognised story tag <word> (vocabulary: …)`), matching the file-before-detail order the existing `*story-9999.sh*ubunut*` assertions expect.
- The both-consumers mutation block: unknown tag word and header shifted to line 3 must be rejected by the runner **and** by `check_evidence_claims.py` through the delegation path, each naming the file — testing the delegation rather than one side of it.
- New case for #390 itself: an `(ubuntu, destructive)` mutant must be **derived** (family `ubuntu`), not rejected — pinning the per-word behaviour the whole-string whitelist breaks.

## Vocabulary note

`low-risk` and `read-only` both stay, distinct — per the review and #386's own comment: story 55 asserts risk `low` for `AptUpdate`, which writes the package index, so folding `low-risk` into `read-only` would misname it.

## Verified

- `tests/e2e/story-metadata.test.sh` — 133 stories (79 ubuntu, 54 atomic), including the new combination case
- `tests/release/public-claims.test.sh`, `tests/e2e/story-runner-verdicts.test.sh`, `tests/e2e/provider-parity.test.sh`
- shellcheck `--severity=warning` (same as CI) clean on both touched scripts

Closes #390.

### DIFF
--- a/tests/e2e/run-stories.sh
+++ b/tests/e2e/run-stories.sh
@@ -155,6 +155,15 @@ declare -A STORY_FAMILY
 ALL_STORY_IDS=()
 UBUNTU_STORY_IDS=()
 
+# Closed per-word tag vocabulary (#390). An unrecognised word exits like an
+# unparseable header, naming the file: a typo in one header must not silently
+# move a story between families and change every derived count. The inventory
+# this was settled against is every distinct tag word in tests/e2e/stories.
+STORY_TAG_VOCAB=(
+    ubuntu destructive medium-risk high-risk read-only low-risk
+    rejection compound "hard compound"
+)
+
 for _story_file in "$STORY_DIR"/story-*.sh; do
     [ -f "$_story_file" ] || continue
     _header="$(sed -n '2p' "$_story_file")"
@@ -177,25 +186,38 @@ for _story_file in "$STORY_DIR"/story-*.sh; do
     # SYSKNIFE_ALLOW_DESTRUCTIVE=1). An empty tag string — a header with no
     # parenthesized field, which many atomic stories carry — is valid atomic.
     #
-    # The strings below are historical labels, not a redesigned schema, and
-    # they are not one semantic dimension: `ubuntu` is family metadata,
+    # The vocabulary is per comma-separated word, not per whole string: a new
+    # story tagged `(ubuntu, destructive)` is a combination that does not
+    # exist in the tree today and is perfectly meaningful, and a whole-string
+    # whitelist would fail it in CI until somebody edited a `case` arm (#390).
+    # The words are historical labels, not a redesigned schema, and they are
+    # not one semantic dimension: `ubuntu` is family metadata,
     # `read-only` / `medium-risk` / `high-risk` follow the execution/risk
     # convention from #219, atomic `destructive` names the host-changing gated
     # subset (not ActionSpec `high-risk`), `rejection` marks a negative
     # planner-behavior test, and `compound` / `hard compound` are scenario
-    # labels. `ubuntu, low-risk` stays as legacy vocabulary: story 55 asserts
-    # risk `low` for AptUpdate, which mutates host state, so it is not a
-    # spelling of `ubuntu, read-only` and must not be normalized into one.
-    case "$_tags" in
-        ""|"destructive"|"read-only"|"hard compound"|\
-        "ubuntu, low-risk"|"ubuntu, read-only"|"ubuntu, medium-risk"|\
-        "ubuntu, high-risk"|"ubuntu, rejection"|"ubuntu, compound") ;;
-        *)
-            printf 'ERROR: unrecognized story tags in %s: %s\n' \
-                "$_story_file" "$_tags" >&2
-            exit 1
-            ;;
-    esac
+    # labels. `low-risk` stays distinct from `read-only` as legacy vocabulary:
+    # story 55 asserts risk `low` for AptUpdate, which mutates host state, so
+    # it is not a spelling of `read-only` and must not be normalized into one.
+    if [[ -n "$_tags" ]]; then
+        IFS=',' read -r -a _tag_parts <<< "$_tags"
+        for _tag in "${_tag_parts[@]}"; do
+            _tag="${_tag#"${_tag%%[![:space:]]*}"}"
+            _tag="${_tag%"${_tag##*[![:space:]]}"}"
+            _tag_known=false
+            for _vocab in "${STORY_TAG_VOCAB[@]}"; do
+                if [[ "$_tag" == "$_vocab" ]]; then
+                    _tag_known=true
+                    break
+                fi
+            done
+            if [[ "$_tag_known" != true ]]; then
+                printf 'ERROR: %s: unrecognised story tag %q (vocabulary: %s)\n' \
+                    "$_story_file" "$_tag" "${STORY_TAG_VOCAB[*]}" >&2
+                exit 1
+            fi
+        done
+    fi
     STORY_NAMES[$_id]="$_title"
     if [[ "$_tags" == *ubuntu* ]]; then
         STORY_FAMILY[$_id]="ubuntu"
--- a/tests/e2e/story-metadata.test.sh
+++ b/tests/e2e/story-metadata.test.sh
@@ -135,6 +135,24 @@ if ! bash "$mutant/run-stories.sh" --metadata >/dev/null 2>&1; then
     report "the unmutated copy also fails — the unknown-tag result is meaningless"
 fi
 
+# #390: the vocabulary is per-word, so a combination of known words that does
+# not exist in the tree today — `(ubuntu, destructive)` is exactly that — is
+# perfectly meaningful and must be derived, not rejected. A whole-string
+# whitelist fails CI on it until somebody edits a `case` arm, which is the
+# brittleness in the direction this repository grows that #390 names.
+cp "$story_dir/story-63.sh" "$mutant/stories/story-9999.sh"
+sed -i '2s/.*/# Story 9999 (ubuntu, destructive): New combination/' \
+    "$mutant/stories/story-9999.sh"
+if ! bash "$mutant/run-stories.sh" --metadata >/dev/null 2>&1; then
+    report "a valid new tag combination was rejected by the vocabulary"
+fi
+combo_family="$(bash "$mutant/run-stories.sh" --metadata 2>/dev/null \
+    | awk -F'\t' '$1 == "9999" { print $2 }')"
+if [ "$combo_family" != ubuntu ]; then
+    report "story 9999 (ubuntu, destructive) derived family '$combo_family', expected ubuntu"
+fi
+rm -f "$mutant/stories/story-9999.sh"
+
 # #252: the grammar says the Story header lives on line 2. A valid header
 # shifted to line 3 must fail rather than derive a row from the wrong line.
 cp "$story_dir/story-63.sh" "$mutant/stories/story-9999.sh"


vladimirrott:
This one is my failure, and I want to name it before anything technical.

I pointed you at #252 on 1 September. Somebody else read the same public thread, implemented it independently, and opened #386 at 12:52 today. I reviewed and merged that this evening without checking whether anyone's name was already on the issue, which closed #252 about twenty minutes before you posted here with a finished implementation. You spent an evening on work my process had already spent. No part of that is on you, and none of it is on the other contributor either: they read an open issue and did the work, which is exactly what an open issue invites.

What makes it worse is the timing. I spent today building the check that exists to stop precisely this, after finding that `maintainer offers` had double-booked seven other issues. It reports an issue pointed at two people **who have not answered**. You had answered, which the tool scores as "working on it", and the other contributor was never someone I had pointed anywhere, so neither of the two conditions it looks for was true. I built a detector for the shape I had just been burned by rather than for the shape of the problem. Yours is the case it should have caught first, because an issue somebody is actively working is worth more protection than one somebody has ignored for a week, not less.

Now the part that matters to you, and I mean it rather than offering it as consolation: **your version is better than the one I merged, and I want it.**

Both close the tag vocabulary. What landed checks the whole tag string against ten exact combinations:

```sh
case "$_tags" in
    ""|"destructive"|"read-only"|"hard compound"|\
    "ubuntu, low-risk"|"ubuntu, read-only"|"ubuntu, medium-risk"| ...
```

Yours checks each comma-separated word against a nine-word vocabulary. Both catch `ubunut` and both catch a misspelled risk word, so on the bug in #252 they are equivalent. They are not equivalent afterwards. A new story tagged `(ubuntu, destructive)` is a combination that does not exist in the tree today and is perfectly meaningful; under what I merged it fails CI until somebody edits a `case` arm, and under yours it just works. The whitelist is brittle in exactly the direction this repository grows, and I did not think about that hard enough when I approved it.

Your `story-metadata.test.sh` block is also stronger. Requiring an unrecognised tag **and** a header shifted to line 3 to be rejected by *both* consumers, each naming the file, tests the delegation rather than one side of it.

So: would you be willing to rebase #389 onto `main` as a follow-up? The shape I would merge is the per-word `STORY_TAG_VOCAB` replacing the `case` block, plus your both-consumers mutation block, and dropping whatever the merged PR already covers. I have re-opened the tag-vocabulary half as its own issue rather than making you work against a closed one, and it is yours, assigned, not offered.

Two smaller things, both agreeing with you. Keeping `low-risk` and `read-only` distinct is right, for the reason you and the other PR both landed on independently: story 55 asserts risk `low` for `AptUpdate`, which writes the package index. And you were right that #252 asked for the collapse to be *stated*, not made.

If you would rather not carry this further after tonight, say so and I will land the change myself with your name on the commit. That is a genuine option and not a test.

Georgefifth:
No worries at all — collisions are what public issues are for, and the shape you asked for is now pushed: rebased onto main (post-#386), the `case` block replaced by the per-word `STORY_TAG_VOCAB`, both-consumers mutation block kept, and everything #386 already covers dropped. One addition while I was in there: a story-metadata case asserting the `(ubuntu, destructive)` mutant is *derived* into the ubuntu family rather than rejected, so the per-word behaviour #390 is about stays pinned. Diagnostic order tweaked to file-before-word to match the existing `*story-9999.sh*ubunut*` assertions. Local suite green: story-metadata (133/79/54), public-claims, runner-verdicts, provider-parity, check_evidence_claims, shellcheck clean.

vladimirrott (approved):
Reviewed at `ac96183cbcf6844afd446061b49114c8b6d5ac20`. Approving. Nothing blocking.

The addition I did not ask for is the one that makes this better than a rebase: asserting that the `(ubuntu, destructive)` mutant is *derived* into the ubuntu family rather than merely accepted. Acceptance only proves the vocabulary stopped rejecting it. Derivation proves the family assignment still reads the word. Two different failures, and you covered both.

## What I ran

All of it in rootless podman, tree mounted read-only, no network, since a fork's shell scripts otherwise run as this user. Commands are as I ran them: output goes into a variable first, because a pipe hands you `tail`'s exit code and not the one you wanted, and `grep -v '^time='` drops podman's CNI warnings. `WT` is your work merged onto current `main` (`adab560d` + `ac96183c`, merge `1c903eec`, no conflict).

```
$ out=$(podman run --rm --network=none -v "$WT:/repo:ro" -w /repo docker.io/library/python:3-slim bash tests/e2e/story-metadata.test.sh 2>&1); rc=$?
$ printf '%s\n' "$out" | grep -v '^time='; echo "exit=$rc"
Story metadata derived cleanly: 133 stories (79 ubuntu, 54 atomic).
exit=0
```

133, 79 and 54, matching the numbers in your comment.

The proof that decides this PR is reverting the production hunk and keeping your test. `run-stories.sh` restored to `main`'s whole-string `case` via `git show adab560d:tests/e2e/run-stories.sh > "$BASE/rev/tests/e2e/run-stories.sh"`, your `story-metadata.test.sh` untouched:

```
$ out=$(podman run --rm --network=none -v "$BASE/rev:/repo:ro" -w /repo docker.io/library/python:3-slim bash tests/e2e/story-metadata.test.sh 2>&1); rc=$?
$ printf '%s\n' "$out" | grep -v '^time='; echo "exit=$rc"
FAIL  a valid new tag combination was rejected by the vocabulary
exit=1
```

One assertion fires, the right one, and every other assertion in the file still passes. Your test holds its own fix down and does not overreach.

Replacing a guard means the old bug has to stay caught, so I checked that too. Story 9998 planted with the header `# Story 9998 (ubunut, read-only)`:

```
$ out=$(podman run --rm --network=none -v "$BASE/typo:/repo:ro" -w /repo docker.io/library/python:3-slim bash tests/e2e/run-stories.sh --metadata 2>&1); rc=$?
$ printf '%s\n' "$out" | grep -v '^time=' | tail -3; echo "exit=$rc"
ERROR: /repo/tests/e2e/stories/story-9998.sh: unrecognised story tag ubunut (vocabulary: ubuntu destructive medium-risk high-risk read-only low-risk rejection compound hard compound)
exit=1
```

#252 still closed, and the message names the offending word instead of echoing the whole tag string. Then the guard's own input emptied to `STORY_TAG_VOCAB=()`, to see whether it fails closed or screens nothing:

```
$ out2=$(podman run --rm --network=none -v "$BASE/vocabgone:/repo:ro" -w /repo docker.io/library/python:3-slim bash tests/e2e/run-stories.sh --metadata 2>&1); rc2=$?
$ printf '%s\n' "$out2" | grep -v '^time=' | tail -3; echo "exit=$rc2"
ERROR: /repo/tests/e2e/stories/story-10.sh: unrecognised story tag destructive (vocabulary: )
exit=1
```

Fails closed.

Your code comment says the vocabulary is settled against every distinct tag word in the tree, which is the one claim a reader has to take on trust, so I counted:

```
$ for f in tests/e2e/stories/story-*.sh; do
    sed -n '2p' "$f" | sed -n 's/^# Story [0-9]* *(\([^)]*\)).*/\1/p'
  done | tr ',' '\n' | sed 's/^[[:space:]]*//; s/[[:space:]]*$//' | grep -v '^$' | sort | uniq -c | sort -rn
     79 ubuntu
     27 destructive
     25 medium-risk
     23 high-risk
     17 read-only
     14 low-risk
      3 rejection
      1 hard compound
      1 compound
```

Nine words in the tree, nine entries in your array, one to one. No dead entry and nothing missing.

I also checked whether admitting `(ubuntu, destructive)` could let a host-changing story run ungated, because that is what the combination looks like from the outside. It cannot. Selection keys on `STORY_FAMILY`, and each destructive story gates itself: `story-20.sh` prints `SKIPPED` and exits 0 unless `SYSKNIFE_ALLOW_DESTRUCTIVE=1`. The header tag is descriptive metadata and your change leaves the gate where it was.

The rest of the board, each run on the merged tree in the same podman form as above:

```
tests/release/public-claims.test.sh         Public claims contract passed.                   exit=0
tests/e2e/story-runner-verdicts.test.sh     passed across all three production runners.      exit=0
tests/e2e/provider-parity.test.sh           7 providers reachable from 5 entry points.       exit=0
python3 scripts/check_evidence_claims.py .  Published figures match the evidence artifacts.  exit=0
```

Shellcheck, swept the way the `e2e` job sweeps it, run on the host because shellcheck parses rather than executes:

```
$ sc=$(find tests/e2e tests/release scripts assets/demo -type f -name '*.sh' -print0 | xargs -0 shellcheck --severity=warning 2>&1); scrc=$?
$ printf '%s\n' "$sc" | head -20; echo "SHELLCHECK_EXIT=$scrc"
SHELLCHECK_EXIT=0
```

192 files swept, clean. Every verification in your comment reproduces. Eleven checks green on GitHub, none pending.

## One optional note

`"${STORY_TAG_VOCAB[*]}"` joins on a space, so the diagnostic ends `... rejection compound hard compound`. A contributor reading that cannot tell `hard compound` is a single entry while `hard` on its own is not valid, and nine entries read as ten words. A comma join would say it. Cosmetic, and I am not holding anything for it.

## Merging

Ready on the evidence. The merge gate I run locally cannot issue its receipt, which is my bug and not yours:

```
$ maintainer-merge verify 389 ac96183cbcf6844afd446061b49114c8b6d5ac20 \
  "tests/e2e/story-metadata.test.sh" \
  "s/_tag_known=false/_tag_known=true/"
  suite: shell (docker.io/library/bash:5 under podman)
  running 'tests/e2e/story-metadata.test.sh' unmutated
    FAIL  consumer line-3 failure did not name the file: tests/e2e/story-metadata.test.sh: line 184: python3: command not found
```

That suite runs in `bash:5`, which ships no `python3`, and since #386 this test drives `check_evidence_claims.py` through the delegation path. The image is wrong for the test, not the reverse. `python:3.12-slim` carries bash 5.2 and runs your file clean, which I confirmed before writing this. I am changing the image on my side, and this lands once that is done.

One correction to my earlier comment on #390: I told you it was assigned to you. It is labelled `claimed` and it is yours, but the assignment did not take, because GitHub drops an assignee who has not posted on the thread and you commented here instead of there. My mistake to state it as done without reading it back. This PR is the visible record.

## Next

#387 fits what you just did more closely than anything else open: `tests/release/no-secrets.test.sh` stays green when its Slack pattern is deleted, and its tree scan passes on an empty file list. A gate that screens nothing and reports success, in the same shape you just mutation-proved from both sides. It is yours if you want it and I will hold it rather than offering it elsewhere.

vladimirrott:
Merged as `cce5c7a`, closing #390.

The part I want on the record is the rebase. #386 landed on top of your branch and overlapped it, and you came back with the overlap subtracted and a list of what you had dropped and why. That is the most expensive thing to do well in a review and the easiest to skip; almost nobody writes the dropped half down.

Workspace at `5673d20`, on `main` with your work in it:

```
```

#397 is still sitting with you and nobody else is on it.

Issue #219, opened 2026-08-17
Add E2E story coverage for the fail2ban action family


vladimirrott:
The fail2ban action family has no end-to-end story. `ConfigureFail2banJail`, `Fail2banBanIp`, `Fail2banUnbanIp` and `Fail2banStatus` are all absent from `tests/e2e/stories/`, and `grep -li fail2ban tests/e2e/stories/` returns nothing.

The construction path is well covered: `crates/sysknife-daemon/src/actions/fail2ban.rs` carries 20 unit tests, including the `IpAddr::from_str` validation that rejects a malformed address before the `ActionSpec` is built. What is missing is the other half, proof that the planner selects these actions from a natural-language intent on a real Ubuntu host.

## Why it is worth doing

fail2ban is a security action family, so the planner picking the wrong action here has a sharper cost than a mislabelled disk query. A story like "ban 203.0.113.7 in the sshd jail" exercises the whole chain: intent to `Fail2banBanIp` with the right jail and IP params, the daemon's validation, and the approval gate. That is exactly what the E2E suite exists to catch and what unit tests cannot.

## Scope

Add stories under `tests/e2e/stories/`, following the existing header convention (`# Story <id> (ubuntu, <risk>): <Name>`, where the risk tag is `read-only`, `medium-risk` or `high-risk`, as in `story-101.sh` and `story-102.sh`) and the pattern in the neighbouring Debian-only stories:

- `Fail2banStatus`: a read-only "is fail2ban running" intent (Observer).
- `ConfigureFail2banJail`: enable a jail (mutating, needs approval).
- `Fail2banBanIp` / `Fail2banUnbanIp`: ban and unban a specific address in a named jail.

Record and replay them on Ubuntu the same way the current 79 stories run: `UBUNTU_RELEASE=noble ./tests/e2e/ubuntu-vm.sh` for the live pass, then commit the cassette and its replay twin so the suite still runs offline.

Start with `Fail2banStatus` alone if you want a smaller first PR. One read-only story is a complete, mergeable change.

## Difficulty

`medium`. The harness and the pattern exist; the work is writing the intents and recording a clean pass.

## Getting started

[CONTRIBUTING.md](https://github.com/lacs-project/sysknife/blob/main/CONTRIBUTING.md) has the build and test commands, and the E2E section covers the VM harness. No CLA and no copyright waiver. The project is MIT.

0xZKc0de:
Hi @vladimirrott !
[...]

PR #393, opened 2026-09-07
fix(evidence): remove unverifiable test baseline provenance

### LINKED ISSUE #278: Nothing checks that workspace-tests.json was measured at the commit it names

bferanmi806-sketch:
## Summary

- replace `workspace-tests.json`'s unverifiable commit/timestamp provenance with a strict count-only schema
- validate the canonical command associated with each suite
- keep Rust and frontend baselines independently refreshable
- add a focused contract test covering schema mutations, legacy migration, dirty worktrees, and squash-merge-safe behavior
- wire the new contract into local and GitHub CI hygiene gates

Closes #278.

## Why

`workspace-tests.json` recorded a test count beside a Git commit and measurement timestamp, but nothing proved that those fields identified the tree that was actually measured.

That breaks in two normal cases:

- `test_baseline.sh` measures the working tree, so an uncommitted test can change the observed count while `git rev-parse HEAD` still names the old tree;
- SysKnife squash-merges PRs, so a contributor commit that legitimately produced a baseline can disappear from `main` even when the squash commit carries the same change.

A Git ancestry check would therefore create false confidence rather than trustworthy provenance.

Schema v2 makes the contract narrower and verifiable:

> `workspace-tests.json` stores the expected count for each suite and the canonical command that measures it. Verification re-runs the suite on the current checkout and requires the observed count to match. It makes no historical claim about a Git commit or measurement timestamp.

The existing Rust and frontend CI jobs remain the authority for whether the stored count is true.

## Schema

The committed artifact now contains only:

```json
{
  "commands": {
    "frontend_tests": "vitest run (apps/sysknife-shell)",
    "tests": "cargo nextest run --workspace --locked"
  },
  "frontend_tests": 72,
  "tests": 1845,
  "version": 2
}
```

Readers reject:

* schema versions other than 2
* malformed or negative counts
* altered canonical commands
* extra fields such as `commit` or `measured_at`

Explicit recording can migrate a legacy v1 artifact to v2 while preserving the untouched suite's independent count. Read-only verification does not rewrite legacy artifacts.

## Regression coverage

`tests/release/test-baseline-provenance.test.sh` covers:

* valid generated v2 artifacts
* Rust and frontend count mismatch handling
* failed suites leaving the artifact untouched
* altered canonical Rust/frontend commands
* hand-added `commit` / `measured_at`
* unsupported schema versions
* per-suite preservation
* v1 → v2 migration
* dirty-working-tree measurement
* squash-equivalent Git histories

The squash fixture deliberately proves the contributor commit is not an ancestor of the squash commit; the count-only contract remains valid because it has no Git-history dependency.

## Related cleanup

`check_evidence_claims.py` invokes the canonical story metadata runner using a repository-relative POSIX path with `cwd` set to the repository root. This preserves #386's `run-stories.sh --metadata` behavior while avoiding host-path translation issues when Bash is used from Windows.

## Validation

Passed:

* `bash tests/release/test-baseline-provenance.test.sh`
* `bash tests/release/public-claims.test.sh`
* `bash tests/e2e/story-metadata.test.sh`
* `bash tests/e2e/story-runner-verdicts.test.sh`
* `python3 scripts/check_evidence_claims.py .`
* `python3 -m py_compile scripts/record_test_baseline.py`
* `bash -n scripts/test_baseline.sh`
* ShellCheck on touched shell files
* `git diff --check`

Story metadata remains:

```text
133 total
79 ubuntu
54 atomic
```

Local full Rust remeasurement is unavailable because the local Rust toolchain is 1.88.0 while current dependencies require 1.94.0. Local frontend remeasurement is unavailable because Vitest is not installed. The baseline was therefore not estimated or regenerated; the upstream 1845 Rust / 72 frontend values were preserved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,9 @@ jobs:
       - name: Check public support and approval claims
         run: bash tests/release/public-claims.test.sh
 
+      - name: Check test-baseline artifact contract
+        run: bash tests/release/test-baseline-provenance.test.sh
+
       - name: Test setup package contracts
         run: npm test --prefix packages/setup
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,12 @@ UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh
 grep -rn 'Rust tests' README.md docs/introduction.md docs/distro-support.md
 ```
 
+`workspace-tests.json` stores the expected count for each suite and the canonical
+command that measures it. Verification re-runs the suite on the current
+checkout and requires the observed count to match. It makes no historical claim
+about a Git commit or measurement timestamp. Do not add provenance metadata by
+hand.
+
 `cargo-nextest` itself needs no system packages. If it is the missing piece,
 install it directly:
 
--- a/scripts/check_evidence_claims.py
+++ b/scripts/check_evidence_claims.py
@@ -30,6 +30,14 @@ import subprocess
 import sys
 from pathlib import Path
 
+# The release contract tests import this module to derive CLAIM_FILES. Keep the
+# sibling baseline schema available in both direct-script and importlib modes.
+SCRIPT_DIR = Path(__file__).resolve().parent
+if str(SCRIPT_DIR) not in sys.path:
+    sys.path.insert(0, str(SCRIPT_DIR))
+
+from record_test_baseline import validate_document
+
 # Files that carry public claims. Kept in step with claim_files in
 # check_public_claims.sh; a file listed there but not here is simply unchecked
 # for numbers, which is why the pristine-fixture guard in the test matters.
@@ -261,10 +269,12 @@ def _canonical_story_metadata(root: Path) -> list[tuple[str, str, str]]:
             f"could not derive story families: canonical metadata runner is "
             f"missing: {STORY_METADATA_RUNNER}"
         )
+    runner_arg = runner.relative_to(root).as_posix()
     try:
         completed = subprocess.run(
-            ["bash", str(runner), "--metadata"],
+            ["bash", runner_arg, "--metadata"],
             capture_output=True,
+            cwd=root,
             text=True,
             encoding="utf-8",
             timeout=120,
@@ -754,21 +764,30 @@ def check_validated_tiers(texts: dict[str, str], root: Path) -> list[str]:
     return problems
 
 
+def load_test_baseline(root: Path) -> dict:
+    path = root / TEST_BASELINE
+    if not path.exists():
+        raise Failure(
+            f"{TEST_BASELINE} is missing; record it with "
+            "UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh"
+        )
+    try:
+        baseline = json.loads(path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError) as exc:
+        raise Failure(f"{TEST_BASELINE} is not readable JSON: {exc}") from exc
+
+    problems = validate_document(baseline, require_all_fields=True)
+    if problems:
+        raise Failure(f"{TEST_BASELINE} is invalid: {'; '.join(problems)}")
+    return baseline
+
+
 def main() -> int:
     root = Path(sys.argv[1] if len(sys.argv) > 1 else ".").resolve()
     try:
         texts = read_claim_files(root)
 
-        baseline_path = root / TEST_BASELINE
-        if not baseline_path.exists():
-            raise Failure(
-                f"{TEST_BASELINE} is missing; record it with "
-                "UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh"
-            )
-        baseline = json.loads(baseline_path.read_text())
-        for field in ("tests", "frontend_tests"):
-            if not isinstance(baseline.get(field), int):
-                raise Failure(f"{TEST_BASELINE} has no integer '{field}' field")
+        baseline = load_test_baseline(root)
 
         problems = []
         problems += check_figure(texts, "Rust tests", baseline["tests"])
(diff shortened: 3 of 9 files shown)


vladimirrott (approved):
Approving at `645aadd7`.

The decision I want to name is the one you did *not* make. #278 asks for provenance on `workspace-tests.json`, and the obvious answer is a git-ancestry check on the recorded commit. You worked out that it cannot be honest here: `test_baseline.sh` measures the working tree, so an uncommitted test moves the count while `HEAD` still names the old tree, and this repository squash-merges, so the commit that legitimately produced a baseline routinely stops existing on `main`. An ancestry check would have passed confidently and meant nothing.

Removing the unverifiable field is the harder answer and the right one. A field nobody can check is worse than an absent field, because it reads as evidence.

What I verified, in a scratch clone, resetting between every mutation and proving the restore rather than assuming it:

```
S0 pristine                     rc=0  dirty=0   contract passed
count -> 999                    rc=1  names docs/distro-support.md
commands block dropped          rc=1  "Evidence check could not run"
legacy v1 document resurfaced   rc=1  fails closed
S0 restored                     rc=0  dirty=0
```

Each one fails for its own reason with its own diagnostic, and the last line is what makes the middle three worth reading.

One honest note on my own method. My first pass reverted `record_test_baseline.py` to `main` and the test failed with an `ImportError`, which proves the test is coupled to your production change but not that its assertions discriminate. Worse, I restored with `git checkout -- .` after a `git checkout origin/main -- <path>`, which stages into the index, so the revert survived and the next two mutations ran against a broken tree and told me nothing. I reran them properly. Mentioning it because you guard against exactly this shape in your own harness elsewhere, and I walked into it while checking yours.

**Why this sat for an hour before merging, which is my fault and not yours.** #278 was pointed at another contributor on 1 September. I checked in on 3 September, heard nothing, and then reserved it for you on 7 September without ever releasing them on the thread. From their side the issue changed hands in silence. A gate I added today refuses to merge a pull request closing an issue that still lists somebody else, it refused this one, and it was right to. I have released them publicly and the block cleared. Nothing about that reflects on your work, and the promptness of it was not in question: you replied twelve minutes after I offered it.

Nothing blocking. No new issue from me while #233 and #219 are still yours.

Since your work here keeps landing on the evidence-and-claims surface: if you operate anything where a recorded number has to still be true later, `sysknife audit verify` is the same idea against a signed action chain, read-only, and it exits non-zero when the record and the reality disagree.

vladimirrott:
The best decision in this PR is one you did not make: you refused to add the Git
ancestry check.

That was the obvious fix, it would have looked rigorous in review, and your
write-up says why it is worse than nothing here. `test_baseline.sh` measures the
working tree, so an uncommitted test moves the count while `git rev-parse HEAD`
still names the old tree, and this project squash-merges, so the commit that
produced a baseline stops existing on `main`. An ancestry check would have gone
green on both and told a reader the number was provenanced when it was not.
Removing a field is a harder argument to win than adding a check, and you won it
on the evidence rather than on taste.

Schema v2 is narrower and every part of it is checkable. The contract test is
what makes it stick: schema mutations, the legacy migration and the dirty
worktree are the three ways this would have rotted back.

Merged at `ef8fe45`, and the gate is green at `adab560` with the count that
landed after it:

```
$ cat tests/evidence/workspace-tests.json
{
  "commands": {
    "frontend_tests": "vitest run (apps/sysknife-shell)",
    "tests": "cargo nextest run --workspace --locked"
  },
  "frontend_tests": 72,
  "tests": 1845,
  "version": 2
}
```

Both issues you took are now assigned to you as well as labelled, so they show
up under `assignee:@me` and on your own dashboard rather than only to people
reading those threads:

```
$ gh issue view 233 --repo lacs-project/sysknife --json assignees --jq '[.assignees[].login]'
["bferanmi806-sketch"]
$ gh issue view 219 --repo lacs-project/sysknife --json assignees --jq '[.assignees[].login]'
["bferanmi806-sketch"]
```

#219 first, as you said, then the clusters in #233. Nobody else is on either.

bferanmi806-sketch:
@vladimirrott Thanks for the detailed review it's really helpful

PR #394, opened 2026-09-07
fix(setup): align wizard guidance with installation outcomes

LunaMeerkats:
## Summary

The setup wizard could suggest starting or inspecting a user service after a system, skipped, or no-systemd installation. It now defers service commands until the installation result is known, keeps the existing manual steps for incomplete installations, and retains user-service guidance when that unit was installed.

Redirected downloads now print periodic newline-terminated progress and a final count; terminals retain the existing carriage-return display. `--no-prompts` omits interactive key-entry instructions while preserving the environment-key reminder and avoiding embedding supplied keys.

## Related Issue

Partially addresses #337: the invited wizard findings **1, 3 and 4**. Provider selection (finding 2) remains separate, so this PR does not close the issue.

## Validation

- [x] Tests added or updated: six real-wizard transcript cases and five offline downloader cases.
- [x] Documentation updated in `docs/quickstart.md`.
- [x] Security impact considered; installation, execution and approval behavior are unchanged.
- [x] Trust boundary preserved (daemon remains the only privileged executor).
- [x] All hosted checks completed — [Ubuntu CI](https://github.com/lacs-project/sysknife/actions/runs/34166705548) passed, including all 113 setup tests. Secret scanning and both Rust and JavaScript CodeQL passed: all three workflows and nine jobs succeeded on `9dfb171`.

Pinned upstream: `c82bb1f1bd017cca6fb40a14106710d16e32fd8f`. Local validation used Ubuntu 26.04 under WSL and Node 24.16.0:

```text
npm test --prefix packages/setup
  pristine baseline: 102 passed
  final: 113 passed, 0 failed, 0 skipped
node --test packages/setup/tests/setup-contract.test.mjs
  before production changes: 10 passed, 5 expected failures
  after: 15 passed
node --test packages/setup/tests/download-progress.test.mjs
  before production changes: 1 passed, 4 expected failures
  after: 5 passed
python3 scripts/check_evidence_claims.py
  passed in the required Bash/POSIX environment
git diff --check
  passed
```

No Rust file or published Rust test-count file changed; the documented no-Rust validation path applies. Tests use isolated configuration/socket fixtures, synthetic keys and an offline HTTPS stream. Live installation/provider behavior was not exercised.

## Notes for Reviewers

Prepared, tested, reviewed and submitted autonomously by OpenAI Codex using the configured personal account. An independent Codex agent reviewed the complete diff. No human pre-submission review was performed or is claimed.

The change corrects human-readable guidance and progress without changing arguments, exit codes, generated configuration, provider defaults or the downloader's returned bytes. Remote SSH/vsock next steps and the incomplete-setup contract from #312 are preserved.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -36,6 +36,13 @@ npx sysknife-setup --claude --no-prompts --daemon-mode=system
 actions work, and anything mutating (installing packages, restarting services)
 does not, because the sudoers grants belong to the `sysknife` system user.
 
+For system mode, follow the printed installation commands; the wizard does not
+install the system service itself. A skipped install or a host without systemd
+gets manual daemon steps instead of commands for a user unit that was never
+installed. With `--no-prompts`, supply any required API key through the
+environment (see Step 2); the wizard does not collect one interactively.
+Redirected downloads print periodic progress lines and a final byte count.
+
 ### From source
 
 **Prerequisites:** Rust stable (`rustup update stable`), **a C compiler and
--- a/packages/setup/index.js
+++ b/packages/setup/index.js
@@ -478,13 +478,8 @@ async function collectTarget(rl, lineQueue, idx) {
       ok(`Daemon socket reachable: ${socket}`);
     } else if (reachable === false) {
       warn(`Daemon socket not reachable: ${socket}`);
-      // A socket under /run/user/<uid> belongs to the user service; telling
-      // someone to `sudo systemctl start` it sends them to a unit that does
-      // not exist on their machine.
-      step(socket.includes('/run/user/')
-        ? `Start the daemon:  systemctl --user start sysknife-daemon`
-        : `Start the daemon:  sudo systemctl start sysknife-daemon`);
-      step(`       or build:   cargo run -p sysknife-daemon`);
+      // Installation has not happened yet. Defer commands until its outcome
+      // tells us which service, if any, exists.
     }
   }
 
@@ -509,7 +504,7 @@ async function collectTarget(rl, lineQueue, idx) {
 // Next-step hint for a single target
 // ---------------------------------------------------------------------------
 
-function targetNextStep(target) {
+function targetNextStep(target, daemonInstall) {
   const { socket, name } = target;
   const label = name ? `${name} (${socket})` : socket;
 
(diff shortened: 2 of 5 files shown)


vladimirrott (approved):
Approving at `9dfb171`.

The bug you fixed is one a maintainer never sees, because it only appears to somebody whose install went a way ours did not. The wizard printed `systemctl --user enable --now sysknife-daemon` from a socket path alone, so a system install, a skipped install, or a machine with no systemd all got told to start a unit that had never been created. Somebody hitting that has no way to tell whether the tool is broken or they are. Gating every service hint on `daemonInstall.mode === 'user' && daemonInstall.daemonInstalled` is exactly right: a path is not evidence that a unit exists.

Verified locally on Linux rather than taking the CI link:

```
S0 at 9dfb171                      113 pass, 0 fail
un-gate the user-service hint      110 pass, 3 fail
restored                           dirty=0
```

Your `npm test` count matches what you reported to the line, and removing one of the three gate sites takes three tests with it, so they are watching the branch and not the file.

Two things I want to credit specifically.

You declined to close #337. It invites findings 1 through 4, you did 1, 3 and 4, and you said so in the body instead of letting the issue auto-close on merge and leaving provider selection to be rediscovered by whoever hits it next. That is the kind of scoping that is invisible when done right and expensive when skipped.

The `--no-prompts` change is the same instinct as the main fix. Printing interactive key-entry instructions to a non-interactive run is advice nobody can act on, and you kept the environment-variable reminder, which is the half that still applies. Removing output is harder to justify in review than adding it, so: it was the right call.

Nothing blocking. No new issue from me, since the rest of #337 is still yours and I would rather you have room for it.

vladimirrott:
The thing I want to name first is that you did not close #337 with this.

You fixed findings 1, 3 and 4, said so in the PR body, and left provider
selection open as its own problem. A PR that claims the issue and fixes most of
it is how a tracker fills with issues that are closed and still broken, and you
avoided that without being asked.

The fix itself is the right shape too. Deferring the service commands until the
installation result is known means the wizard stops printing a `systemctl
--user` line after a system-wide or skipped install, which is the version of
this bug that sends someone into a unit that does not exist. Six real-wizard
transcript cases is the part that will keep it fixed: a transcript test fails
when the *output* regresses, which is what the user actually reads.

Merged at `adab560`. The workspace baseline moved to 1845 and the evidence gate
is green on it:

```
```

Next, and it is the one you already hold: finding 2 on #337, where `--no-prompts`
picks a paid provider over a running Ollama. You are assigned there, so it is
sitting on your dashboard. No deadline on it.

One thing worth more than another patch, and easy to decline: you have been
inside the wizard's output for two PRs now, so you know what it should say. If
you ever run `npx sysknife-setup` on a machine you actually use, I would like to
hear what the transcript looked like on real hardware. `sysknife doctor` and
`sysknife history` afterwards change nothing on the host and would tell me
whether the guidance holds up outside the test fixtures.

Issue #395, opened 2026-09-08
markdown-link-files.sh exits 0 when its git call fails, so both link gates pass over zero files


vladimirrott:
`scripts/markdown-link-files.sh` derives the Markdown files that both link-check
jobs run over, and it ends on a process substitution whose exit status nothing
reads:

```
$ sed -n '39,48p' scripts/markdown-link-files.sh
while IFS= read -r -d '' path; do
    skip=false
    for excluded in "${exclusions[@]}"; do
        if [[ "$path" == "$excluded" ]]; then
            skip=true
            break
        fi
    done
    [[ "$skip" == true ]] || printf '%s\0' "$path"
done < <(git -C "$repo_root" ls-files -z -- '*.md')
```

When that `git ls-files` fails, the loop runs zero times, the script prints
nothing, and it exits 0. `set -euo pipefail` on line 2 does not see it: a
process substitution's status is not the status of the command feeding it.

The four consumers repeat the shape, so the empty list survives a second time:

```
$ grep -n 'markdown-link-files.sh' .github/workflows/ci.yml scripts/ci-local.sh
scripts/ci-local.sh:241:    done < <(scripts/markdown-link-files.sh)
scripts/ci-local.sh:244:    done < <(scripts/markdown-link-files.sh --external)
.github/workflows/ci.yml:150:          done < <(scripts/markdown-link-files.sh)
.github/workflows/ci.yml:175:          done < <(scripts/markdown-link-files.sh --external)
```

Reproduced at `adab560`. The script below copies the three real files somewhere
without a Git repository, which is the cheapest way to make the real `git` call
fail, then runs the `ci.yml` step body over the result:

```bash
#!/usr/bin/env bash
T=$(mktemp -d)
trap 'rm -rf "$T"' EXIT
mkdir -p "$T/scripts"
cp scripts/markdown-link-files.sh \
   scripts/markdown-link-exclusions.txt \
   scripts/markdown-link-external-files.txt "$T/scripts/"

echo "--- the producer, run where its git call fails:"
bash "$T/scripts/markdown-link-files.sh" >/dev/null
echo "script_exit=$?"

echo "--- the ci.yml:148-150 step body, over that same producer:"
bash -e -c '
  n=0
  while IFS= read -r -d "" file; do
    n=$((n+1))
  done < <("'"$T"'/scripts/markdown-link-files.sh")
  echo "files_checked=$n"
'
echo "step_exit=$?"
```

```
--- the producer, run where its git call fails:
[...]

PR #398, opened 2026-09-08
fix(release): validate missing registry version arguments

### LINKED ISSUE #396: check_registry_versions.sh dies on set -u before it can print its own usage error
`scripts/check_registry_versions.sh` opens with an argument check that cannot
run for the case it was written for:

```
$ sed -n '1,8p' scripts/check_registry_versions.sh
#!/usr/bin/env bash
set -euo pipefail

version="${1#v}"
[[ "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || {
    printf 'ERROR: expected a semantic version, got %q\n' "$1" >&2
    exit 2
}
```

Line 2 turns on `set -u`, and line 4 expands `$1` without a default. With no
argument the shell kills the script on line 4, so the message on line 6 never
prints:

```
$ bash scripts/check_registry_versions.sh; echo "exit=$?"
scripts/check_registry_versions.sh: line 4: 1: unbound variable
exit=1
```

Pass something non-empty and the intended path works:

```
$ bash scripts/check_registry_versions.sh nope; echo "exit=$?"
ERROR: expected a semantic version, got nope
exit=2
```

So the only invocation a person is likely to try by hand, running it with
nothing to see what it wants, is the one invocation whose error message is
unreachable.

## Why it matters

Small, and worth fixing for two reasons rather than one.

The release path is unaffected, which is the first thing to establish before
anyone worries:

```
$ grep -rn 'check_registry_versions' .github/workflows/ tests/
.github/workflows/release.yml:47:        run: bash scripts/check_registry_versions.sh "${{ github.ref_name }}"
tests/release/release-rehearsal.test.sh:39:grep -Fq 'check_registry_versions.sh' "$release_workflow"
```

`github.ref_name` is always passed as an argument, so `$1` is set even if it
were empty, and an empty value takes the regex branch and reports properly.

What it costs is the contributor who runs the script to find out how to run it,
and gets `line 4: 1: unbound variable` with exit 1 instead of a usage message
with exit 2. Two different exit codes for the same class of mistake, and the
useful one is the code path you cannot reach.

`CONTRIBUTING.md` also asks contributors to run the release scripts directly,
so this is the first thing one of them sees from this file.

## Scope

- Give `$1` a default before the strip, so the validation branch is reachable:
  `version="${1:-}"; version="${version#v}"`.
- Keep exit 2 for both shapes of bad input. A caller distinguishing "you passed
  nothing" from "you passed nonsense" gains nothing here, and one code for
  "usage" is easier to document.
- Print the same message either way. `%q` on an empty string renders `''`, which
  reads correctly.
- Check the other scripts in `scripts/` for the same pattern while you are in
  there, and say in the PR which ones you checked, including the ones that were
  already fine.

## Tests first

There is no test for this script today, so the assertion comes first and the
one-line fix comes second.

Add a case to `tests/release/` that runs the script with no argument and asserts
exit 2 and the `ERROR: expected a semantic version` text. Run it before the fix
and watch it fail with exit 1 and the bash diagnostic; that failure is the
evidence the test is real.

Add the `nope` case alongside it. Two cases, one of which passes today, is how
you show the new test discriminates rather than just going green.

Do not assert on the exact wording of bash's `unbound variable` message. It is
what you are removing.

## Difficulty

`easy`. One line of shell and a small test file. Good first change in this
repository: nothing to build, nothing to install, and the whole loop runs in
under a second.

## Getting started

has the build and test commands. This one needs no VM, no daemon, no LLM
provider, no credentials and no network: the script exits before its first
`curl`. No CLA and no copyright waiver. The project is MIT.

mikevillari:
## Summary
A missing version argument previously triggered Bash's unbound-variable error and exit 1. Default the argument before stripping the optional v prefix, and use a safe expansion in the diagnostic too, so missing/empty input receives the existing semantic-version error and exit 2.

## Related Issue
Closes #396.

## Validation
- [x] Tests added or updated — missing, empty, and malformed input in the existing release rehearsal test.
- [x] Security impact considered.
- [ ] CI passes — awaiting upstream CI; full local CI did not pass.
- Wrote the regression first: the malformed input passed, then the no-argument case failed with expected exit 2 / actual exit 1. After the change, `bash tests/release/release-rehearsal.test.sh` passes.
- `python3 scripts/check_evidence_claims.py`, Bash syntax checks, and `git diff --check` pass.
- Ran `bash scripts/ci-local.sh --no-postgres`: 19 checks passed, eight failed. Failures: Rust/frontend dependency DNS access; existing GNU sed/grep expectations on macOS (four tests); setup test's Unix socket bind denied by sandbox. Optional lint/audit/nextest tools were absent. No Rust files changed.
- Tests ran with an isolated Bash 5.2.37 because macOS Bash 3.2 cannot run existing mapfile/associative-array scripts.

## Notes for Reviewers
Audited other top-level positional-argument handling: check_release_versions.sh, check_public_claims.sh, markdown-link-files.sh, test_baseline.sh, and check_no_secrets.sh already use default expansions. release_rehearsal.sh and ci-local.sh guard their argument loops; release_rehearsal also checks for a missing --output value.

No release, registry publication, runtime API, or schema changes. This repairs the existing invalid-input contract, so no documentation changes were needed.

AI-assisted with Codex; the diff and test results were inspected by the agent. No independent human-review claim. Two browser commits can be squashed.

### DIFF
--- a/scripts/check_registry_versions.sh
+++ b/scripts/check_registry_versions.sh
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-version="${1#v}"
+version="${1:-}"
+version="${version#v}"
 [[ "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || {
-    printf 'ERROR: expected a semantic version, got %q\n' "$1" >&2
+    printf 'ERROR: expected a semantic version, got %q\n' "${1:-}" >&2
     exit 2
 }
 
--- a/tests/release/release-rehearsal.test.sh
+++ b/tests/release/release-rehearsal.test.sh
@@ -4,6 +4,27 @@ set -euo pipefail
 repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 rehearsal="${repo_root}/scripts/release_rehearsal.sh"
 
+# Invalid registry versions must fail validation before attempting any network
+# requests, including when no positional argument was supplied.
+assert_invalid_registry_version() {
+    local output status
+    if output="$(bash "${repo_root}/scripts/check_registry_versions.sh" "$@" 2>&1)"; then
+        printf 'FAIL: registry preflight accepted an invalid version\n' >&2
+        exit 1
+    else
+        status=$?
+    fi
+    if [[ "$status" -ne 2 ]]; then
+        printf 'FAIL: registry preflight expected exit 2, got %s: %s\n' "$status" "$output" >&2
+        exit 1
+    fi
+    grep -Fq 'ERROR: expected a semantic version' <<<"$output"
+}
+
+assert_invalid_registry_version nope
+assert_invalid_registry_version
+assert_invalid_registry_version ''
+
 if [[ ! -x "$rehearsal" ]]; then
     printf 'FAIL: release rehearsal is missing or not executable: %s\n' "$rehearsal" >&2
     exit 1


vladimirrott (approved):
Approved at `0ba6e074a69678b32b0ffb4ec0c07b36c94cb45d`.

You wrote the regression before the fix and it discriminates. That is the whole ask on this tracker, and it is the part most small fixes skip.

`maintainer screen 398` returns DO NOT EXECUTE, so nothing ran on the host. All of the below ran in `docker.io/library/bash:5` under rootless podman, `--network=none`, tree from `git archive 0ba6e074`.

## Reverting your two lines turns the test red

```
$ out=$(bash /home/entropia/.local/state/sysknife-maint/pt.sh p398 bash -c "$(cat /tmp/sk398proof.sh)" 2>&1); rc=$?; printf '%s\n' "$out" | grep -v '^time='; echo "=== wrapper exit=$rc ==="
### 1. the PR's test, unmutated
exit=1

### 2. revert the production hunk (back to main's two lines) and re-run
#!/usr/bin/env bash
set -euo pipefail

version="${1#v}"
[[ "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || {
    printf 'ERROR: expected a semantic version, got %q\n' "$1" >&2
    exit 2
}
--- confirm the revert matches main byte for byte on those lines ---
4:version="${1#v}"
FAIL: registry preflight expected exit 2, got 1: /repo/scripts/check_registry_versions.sh: line 4: 1: unbound variable
exit=1

### 3. which of the three assertions discriminates? run each against the reverted script
arg=[nope] exit=2  out=ERROR: expected a semantic version, got nope
arg=[NOARG] exit=1  out=scripts/check_registry_versions.sh: line 4: 1: unbound variable
arg=[]     exit=2  out=ERROR: expected a semantic version, got ''
```

Run 1's `required tool is missing: cargo` comes from `scripts/release_rehearsal.sh:51`, which the test reaches after your three assertions. Your assertions sit at the top of the file, so they had already passed when that fired; the container has no cargo, which is my environment and not your code.

Run 3 is the part worth naming. Of the three cases you added, only the bare `assert_invalid_registry_version` discriminates: `nope` and `''` both take the regex branch and exit 2 in either version. Keep all three as anchors, but the coverage is one case wide rather than three. Your `status=$?` in the `else` branch reads the command substitution's own status rather than a pipeline's tail, which is the mistake this repo keeps making, and you avoided it.

## Scope, so the fix is not read as bigger than it is

The only automated caller cannot reach the condition you fixed:

```
$ grep -rn "check_registry_versions" --include=*.yml --include=*.sh --include=*.md . | grep -v '^./target'
.github/workflows/release.yml:47:        run: bash scripts/check_registry_versions.sh "${{ github.ref_name }}"
tests/release/release-rehearsal.test.sh:39:grep -Fq 'check_registry_versions.sh' "$release_workflow"
```

`release.yml` triggers on `push: tags: 'v[0-9]+.[0-9]+.[0-9]+'`, so `github.ref_name` is always a non-empty tag. This improves the message a maintainer sees running the script by hand and closes no release-time hazard. Worth saying out loud so nobody later reads the CHANGELOG and thinks a tag could have shipped on an empty version.

Your reviewer note listing the other five scripts you audited is the right instinct, and it saved me repeating it.

## The merge is blocked by my tooling

Both changed files are `.sh`, so the `shell` suite covers them, and it still refused:

```
$ maintainer-merge verify 398 0ba6e074a69678b32b0ffb4ec0c07b36c94cb45d "tests/release/release-rehearsal.test.sh" 's/^version="\${1:-}"$/version="${1#v}"/'
  suite: shell (docker.io/library/python:3.12-slim under podman)
```

The suite image carries bash and python3. `scripts/release_rehearsal.sh:51` requires `cargo node npm sha256sum file`, so this test can never pass clean there and no receipt is earnable. Nothing for you to do: the fix is to the image rather than to your branch, and it is on my list for tonight.

No CHANGELOG entry needed: this restores an existing invalid-input contract rather than changing one.

## Next

`#238` is a good fit and I say more about it on your other PR. Both of your open PRs are about a guard that reports a guarantee it is not providing, which is the most useful instinct anyone can bring here.

vladimirrott:
Merged as `7fe0c23`, closing #396.

You left the CI checkbox unticked and named the local run you could not finish, instead of ticking it and hoping. I read those checklists closely. That cost you nothing and told me where to look.

Workspace at `5673d20`:

```
```

#395 is yours with #399 open against it. I released #238 and #239 on your word, and saying so before you started rather than after is the version of that which costs nobody an evening.

Separately, with no pressure attached. Two of your PRs now sit on the scripts a contributor runs by hand. `sysknife doctor` is the read-only end of that surface: it reads the host and changes nothing on it. If you ever run it on a machine of your own, I would want to hear what it got wrong. Not a condition of anything.

PR #399, opened 2026-09-08
fix(ci): reject failed and empty Markdown file lists

### LINKED ISSUE #395: markdown-link-files.sh exits 0 when its git call fails, so both link gates pass over zero files

mikevillari:
## Summary
A failed Git enumeration previously produced an empty list and exit 0; both local and workflow link checks could then pass without checking any files.

Capture the producer output in temporary files and check its exit status before consuming it. Reject an empty derived set and zero checked files in all four consumers. This also rejects a producer that emits a partial list and then fails, while preserving NUL-delimited filenames and the existing external-link retry behavior.

## Related Issue
Closes #395.

## Validation
- [x] Tests added or updated.
- [x] Security impact considered.
- [ ] CI passes — upstream CI pending; full local CI has environment failures.
- Added the producer regression first: removing the fixture's Git metadata incorrectly passed before the fix.
- Added a real-consumer regression before changing consumers: the local check incorrectly accepted a failing producer.
- `bash tests/release/markdown-link-files.test.sh` passes. It executes the real local function body and both workflow step bodies with offline stubs; failed, empty, partial-failure, and valid producer cases are covered. Valid cases assert every filename, including one with spaces, reaches the checker.
- `yamllint .github/workflows/ci.yml`, ShellCheck (warning severity) on all three changed shell files, Bash syntax, evidence-claims check, and diff whitespace check pass.
- Ran full `scripts/ci-local.sh --no-postgres` with offline dependency settings: 19 checks passed, eight failed (uncached Rust/frontend dependencies, existing GNU sed/grep assumptions on macOS, and sandbox-denied Unix socket binding). Optional tools were unavailable during that run; YAML/ShellCheck were subsequently installed in isolation and the changed files passed.

## Notes for Reviewers
Temporary files preserve NUL delimiters and are removed by EXIT traps. No Rust/application behavior, dependency manifests, or test-count claims changed. No documentation change is needed for restoring the existing CI failure contract.

AI-assisted with Codex; agent-inspected diff and executed checks, with no independent human-review claim. Four browser commits can be squashed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,12 +145,22 @@ jobs:
 
       - name: Check internal markdown links
         run: |
+          file_list="$(mktemp)"
+          trap 'rm -f "$file_list"' EXIT
+          scripts/markdown-link-files.sh > "$file_list" || exit 1
+          count=0
           while IFS= read -r -d '' file; do
             markdown-link-check --config .markdown-link-check-internal.json "$file"
-          done < <(scripts/markdown-link-files.sh)
+            count=$((count + 1))
+          done < "$file_list"
+          [[ "$count" -gt 0 ]] || { printf 'markdown-link-files: no files checked\n' >&2; exit 1; }
 
       - name: Check external markdown links
         run: |
+          file_list="$(mktemp)"
+          trap 'rm -f "$file_list"' EXIT
+          scripts/markdown-link-files.sh --external > "$file_list" || exit 1
+          count=0
           while IFS= read -r -d '' file; do
             # Retry on transient network failure. `retryOn429` in the config only
             # covers rate limiting, and the failures actually seen here were
@@ -172,7 +182,9 @@ jobs:
               echo "link check for $file failed (attempt $attempt/3); retrying in 10s"
               sleep 10
             done
-          done < <(scripts/markdown-link-files.sh --external)
+            count=$((count + 1))
+          done < "$file_list"
+          [[ "$count" -gt 0 ]] || { printf 'markdown-link-files: no files checked\n' >&2; exit 1; }
 
       - name: Lint GitHub issue templates
         run: yamllint .github/ISSUE_TEMPLATE/*.yml .github/workflows/*.yml
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -236,12 +236,22 @@ hygiene_markdownlint() (
 
 hygiene_markdown_link_check() (
     cd "$repo_root" || exit 1
+    file_list="$(mktemp)"
+    trap 'rm -f "$file_list"' EXIT
+    scripts/markdown-link-files.sh > "$file_list" || exit 1
+    count=0
     while IFS= read -r -d '' f; do
         markdown-link-check --config .markdown-link-check-internal.json "$f" || exit 1
-    done < <(scripts/markdown-link-files.sh)
+        count=$((count + 1))
+    done < "$file_list"
+    [[ "$count" -gt 0 ]] || { printf 'markdown-link-files: no files checked\n' >&2; exit 1; }
+    scripts/markdown-link-files.sh --external > "$file_list" || exit 1
+    count=0
     while IFS= read -r -d '' f; do
         markdown-link-check --config .markdown-link-check.json "$f" || exit 1
-    done < <(scripts/markdown-link-files.sh --external)
+        count=$((count + 1))
+    done < "$file_list"
+    [[ "$count" -gt 0 ]] || { printf 'markdown-link-files: no files checked\n' >&2; exit 1; }
 )
 
 hygiene_yamllint() (
--- a/scripts/markdown-link-files.sh
+++ b/scripts/markdown-link-files.sh
@@ -36,6 +36,13 @@ while IFS= read -r path || [[ -n "$path" ]]; do
     exclusions+=("$path")
 done < "$exclusions_file"
 
+file_list="$(mktemp)"
+trap 'rm -f "$file_list"' EXIT
+if ! git -C "$repo_root" ls-files -z -- '*.md' > "$file_list"; then
+    printf 'markdown-link-files: failed to enumerate tracked Markdown files\n' >&2
+    exit 1
+fi
+count=0
 while IFS= read -r -d '' path; do
     skip=false
     for excluded in "${exclusions[@]}"; do
@@ -44,5 +51,12 @@ while IFS= read -r -d '' path; do
             break
         fi
     done
-    [[ "$skip" == true ]] || printf '%s\0' "$path"
-done < <(git -C "$repo_root" ls-files -z -- '*.md')
+    if [[ "$skip" != true ]]; then
+        printf '%s\0' "$path"
+        count=$((count + 1))
+    fi
+done < "$file_list"
+if [[ "$count" -eq 0 ]]; then
+    printf 'markdown-link-files: no tracked Markdown files to check\n' >&2
+    exit 1
+fi
--- a/tests/release/markdown-link-files.test.sh
+++ b/tests/release/markdown-link-files.test.sh
@@ -70,7 +70,7 @@ if output="$("$fixture/scripts/markdown-link-files.sh" 2>&1)"; then
     printf 'markdown-link-files: failed git enumeration unexpectedly passed\n' >&2
     exit 1
 fi
-grep -Fq 'markdown-link-files' <<< "$output"
+grep -Fq 'failed to enumerate tracked Markdown files' <<< "$output"
 mv "$fixture/git-backup" "$fixture/.git"
 
 git -C "$fixture" rm --cached -q -- '*.md'
@@ -78,7 +78,7 @@ if output="$("$fixture/scripts/markdown-link-files.sh" 2>&1)"; then
     printf 'markdown-link-files: empty derived set unexpectedly passed\n' >&2
     exit 1
 fi
-grep -Fq 'markdown-link-files' <<< "$output"
+grep -Fq 'no tracked Markdown files to check' <<< "$output"
 
 # Exercise the real consumer bodies with an offline checker and producer.
 python3 - "$repo_root" "$fixture" <<'PYTEST'
@@ -113,7 +113,9 @@ for scenario, script in {
     for name, body in bodies.items():
         log = fixture / "checked"
         log.unlink(missing_ok=True)
-        result = subprocess.run(["bash", "-euo", "pipefail", "-c", 'repo_root="$1"\n' + body, "consumer", str(fixture)], cwd=fixture, env=env, capture_output=True, text=True)
+        # ci-local.sh enables strict mode; the workflow uses GitHub's default bash -e.
+        shell = ["bash", "-euo", "pipefail"] if name == "local" else ["bash", "-e"]
+        result = subprocess.run([*shell, "-c", 'repo_root="$1"\n' + body, "consumer", str(fixture)], cwd=fixture, env=env, capture_output=True, text=True)
         if scenario != "valid":
             assert result.returncode != 0, f"{name} consumer accepted {scenario} producer"
         else:
--- a/tests/release/markdown-link-files.test.sh
+++ b/tests/release/markdown-link-files.test.sh
@@ -63,4 +63,63 @@ grep -Fq 'excluded path is not a tracked Markdown file: docs/missing.md' <<< "$o
     exit 1
 }
 
+# Remove exclusions so a failed enumeration reaches git ls-files itself.
+: > "$fixture/scripts/markdown-link-exclusions.txt"
+mv "$fixture/.git" "$fixture/git-backup"
+if output="$("$fixture/scripts/markdown-link-files.sh" 2>&1)"; then
+    printf 'markdown-link-files: failed git enumeration unexpectedly passed\n' >&2
+    exit 1
+fi
+grep -Fq 'markdown-link-files' <<< "$output"
+mv "$fixture/git-backup" "$fixture/.git"
+
+git -C "$fixture" rm --cached -q -- '*.md'
+if output="$("$fixture/scripts/markdown-link-files.sh" 2>&1)"; then
+    printf 'markdown-link-files: empty derived set unexpectedly passed\n' >&2
+    exit 1
+fi
+grep -Fq 'markdown-link-files' <<< "$output"
+
+# Exercise the real consumer bodies with an offline checker and producer.
+python3 - "$repo_root" "$fixture" <<'PYTEST'
+import os
+from pathlib import Path
+import re
+import subprocess
+import sys
+
+root, fixture = map(Path, sys.argv[1:])
+local = (root / "scripts/ci-local.sh").read_text()
+body = re.search(r"hygiene_markdown_link_check\(\) \(\n(.*?)\n\)", local, re.S).group(1)
+workflow = (root / ".github/workflows/ci.yml").read_text()
+bodies = {"local": body}
+for kind in ("internal", "external"):
+    match = re.search(r"      - name: Check " + kind + r" markdown links\n        run: \|\n(.*?)(?=\n      - name:)", workflow, re.S)
+    bodies[kind] = "\n".join(line[10:] for line in match.group(1).splitlines())
+bin_dir = fixture / "bin"
+bin_dir.mkdir()
+checker = bin_dir / "markdown-link-check"
+checker.write_text('#!/usr/bin/env bash\nprintf "%s\\n" "${@: -1}" >> "$CHECK_LOG"\n')
+checker.chmod(0o755)
+env = dict(os.environ, PATH=str(bin_dir) + os.pathsep + os.environ["PATH"], CHECK_LOG=str(fixture / "checked"))
+producer = fixture / "scripts/markdown-link-files.sh"
+for scenario, script in {
+    "failed": "exit 1",
+    "empty": "exit 0",
+    "partial failure": "printf 'README.md\\0'; exit 1",
+    "valid": "printf 'README.md\\0docs/a guide.md\\0'",
+}.items():
+    producer.write_text("#!/usr/bin/env bash\n" + script + "\n")
+    for name, body in bodies.items():
+        log = fixture / "checked"
+        log.unlink(missing_ok=True)
+        result = subprocess.run(["bash", "-euo", "pipefail", "-c", 'repo_root="$1"\n' + body, "consumer", str(fixture)], cwd=fixture, env=env, capture_output=True, text=True)
+        if scenario != "valid":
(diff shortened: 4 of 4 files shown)


vladimirrott (commented):
Reviewed at `147743910736b4c541923de5dd0c85a3a832b51f`. No changes requested; the merge waits on two things only I can do.

Your test reads the live `ci.yml` and `ci-local.sh` text and drives the extracted step bodies. That is the strongest coupling anyone has shipped to this repo: a reverted consumer cannot satisfy it, because the test rebuilds itself from the consumer's own source every run. Compare that to the usual shape, where the assertion is a parallel copy of the production logic and reverting the fix leaves the suite green.

`maintainer screen 399` returns DO NOT EXECUTE. Everything below ran in a `python:3.12-slim` image with git added, under rootless podman, `--network=none`, tree from `git archive 14774391`.

## Reverting either half turns the test red

```
$ out=$(bash /home/entropia/.local/state/sysknife-maint/pti.sh localhost/sk-pygit:1 p399 bash -c "$(cat /tmp/sk399proof.sh)" 2>&1); printf '%s\n' "$out" | grep -v '^time='
### M0. unmutated
markdown-link-files: derived files and exclusions validated.
exit=0

### M1. revert the ci.yml internal-links hunk to main's process substitution
ci.yml internal-links step reverted to main's shape
150:          done < <(scripts/markdown-link-files.sh)
Traceback (most recent call last):
  File "<stdin>", line 34, in <module>
AssertionError: internal consumer accepted failed producer
exit=1
```

And the producer's own two guards:

```
### M2. revert the producer's own two guards in scripts/markdown-link-files.sh
producer reverted to main's shape
0
(both guards gone: 0)
markdown-link-files: failed git enumeration unexpectedly passed
exit=1
```

Both directions bite. The regex miss in the python block raises `AttributeError` on `.group(1)` rather than skipping, so it fails closed too.

## The workflow diff, which I read rather than summarised

```
$ git diff $(git merge-base origin/main p399)..p399 -- .github/workflows/ | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -iE 'on:|permissions|uses:|secrets|pull_request_target|if:'
NONE — no trigger, permission, action-pin, secret or conditional changed
```

The diff changes the two existing `run:` bodies under `docs-and-hygiene` and nothing else. No step added, no step removed, both `- name:` lines are context.

## Two notes, neither blocking

`tests/release/markdown-link-files.test.sh:70` and `:78` assert `grep -Fq 'markdown-link-files' <<< "$output"`. The fixture script lives at `$fixture/scripts/markdown-link-files.sh`, so bash's own error prefix contains that string. I checked rather than reasoned:

```
$ printf "#!/usr/bin/env bash\nif then fi\n" > /tmp/fx/scripts/markdown-link-files.sh
output was: /tmp/fx/scripts/markdown-link-files.sh: line 2: syntax error near unexpected token `then'
ORACLE SATISFIED by a pure syntax error -- the assertion does not pin the diagnostic
```

The non-zero-exit assertion two lines above carries all the discriminating power. Your `:57` and `:65` pin the full sentence including the path, which is the shape worth copying to `:70` and `:78`: grep `failed to enumerate tracked Markdown files` and `no tracked Markdown files to check`.

Your harness runs the extracted bodies under `bash -euo pipefail`, while the runner does not:

```
$ awk '/^  docs-and-hygiene:/,/^  [a-z-]+:$/' .github/workflows/ci.yml | grep -nE 'shell:|defaults:'
(none)
$ grep -nE '^defaults:|^  defaults:' .github/workflows/ci.yml
(no top-level defaults block)
```

With no `shell:` and no `defaults.run`, GitHub uses `bash -e {0}`: no `-u`, no `pipefail`. Latent today, because your `|| exit 1` and `[[ "$count" -gt 0 ]] ||` guards depend on neither. A future body that leans on `pipefail` would pass your test and behave differently on the runner. Either run the bodies under `bash -e`, or add `shell: bash` to the two steps so the two agree. Your call.

The count guard now exists in four places (the producer plus three consumers) and your python block covers all four copies, which is what keeps the drift detectable. Defence in depth rather than accidental duplication; noted so nobody trims it later.

## Blocking, and both halves are mine

Your checks have not run. Four workflow runs sit at `action_required`:

```
$ gh api "repos/lacs-project/sysknife/actions/runs?status=action_required&per_page=50" --jq '.workflow_runs[] | "\(.id)  \(.name)  \(.head_branch)  \(.head_sha[0:8])  \(.status)/\(.conclusion)"'
34183802921  CodeQL  fix/markdown-producer-failure  14774391  completed/action_required
34183802885  ci  fix/markdown-producer-failure  14774391  completed/action_required
34183802884  e2e  fix/markdown-producer-failure  14774391  completed/action_required
34183802891  secret-scan  fix/markdown-producer-failure  14774391  completed/action_required
```

That is a maintainer clicking a button, not anything about your branch, and it has been sitting since your push. Because this PR edits `.github/workflows/ci.yml`, the approval is also the control that stops a fork's code running in CI with a token, so it gets a human reading the workflow diff first rather than an automated approve. I have quoted that diff above to make the read short. It is at the top of tonight's notes with the run IDs.

Once the board is green I will approve and merge without asking you for anything further.

No CHANGELOG entry needed: this restores an existing CI failure contract.

## Next, and this one is a real offer

`#239` is yours if you want it: Debian's default firewall is nftables and the catalogue has no nftables vocabulary at all (I checked, zero occurrences in `action_family.rs` and `sysknife-types`). It is a step sideways from CI plumbing into the action catalogue, and the reason I am offering it to you is that adding a new action family means placing it correctly in the fence lists, the risk table, the prompt params reference and the validator, and getting any one of those wrong is a silent hole rather than a red test. That is the failure class both your PRs are about.

One sequencing note so it is not a surprise: PR #384 is re-partitioning those fence lists right now and will land first. I have posted the offer on the issue.

Say no freely. Both your open PRs found a guard that reported a guarantee it was not providing, and that instinct is worth more here than any particular issue.

mikevillari:
Addressed both optional notes in f9a01f0: the enumeration/empty-set assertions now require the specific diagnostics, and extracted workflow bodies run under bash -e while the local body retains bash -euo pipefail. Only the regression test changed.

Validation: markdown-link-files.test.sh passes all four producer scenarios across all three consumers; bash -n and git diff --check pass. Replacing each expected producer diagnostic with an incorrect message independently passes the old test and fails the updated test. Published file matches the tested local version.

The evidence-claim check could not complete because the local macOS Bash lacks associative arrays. Attempted ci-local.sh --no-postgres; dependency downloads failed, and I stopped its stalled npm phase. No full-CI pass claimed. Codex-assisted. Thanks for the precise review.

vladimirrott (approved):
Approving at `14774391`. My earlier note on this thread said the checks had not
run; somebody has since approved the queued runs and the board is green, so here
is the decision I owed you.

```
$ gh pr checks 399 --repo lacs-project/sysknife | awk -F'\t' '{c[$2]++} END {for (k in c) print k, c[k]}'
skipping 1
pass 11
```

The part I want to name is the Python block at the end of your test. Extracting
the three consumer bodies out of `ci-local.sh` and `ci.yml` and running them
against four synthetic producers means the test exercises the shipped workflow
text rather than a paraphrase of it. A test that re-implemented the loop would
have passed with your fix reverted. Yours does not, and I checked both halves.

Reverting the `ci.yml` internal-links hunk to the process-substitution form:

```
$ out=$(podman run --rm --network=none -v ".../trees/mut399b:/repo:z" -w /repo -e HOME=/tmp localhost/sk-pygit:1 bash tests/release/markdown-link-files.test.sh 2>&1); rc=$?
exit=1
Traceback (most recent call last):
  File "<stdin>", line 34, in <module>
AssertionError: internal consumer accepted failed producer
```

Reverting `scripts/markdown-link-files.sh` to the version on `main` and leaving
the workflow alone:

```
$ git show origin/main:scripts/markdown-link-files.sh > .../trees/mut399b/scripts/markdown-link-files.sh
$ grep -c 'no tracked Markdown files to check' .../trees/mut399b/scripts/markdown-link-files.sh
0
exit=1
markdown-link-files: failed git enumeration unexpectedly passed
```

Restored, the same command exits 0 with
`markdown-link-files: derived files and exclusions validated.`

The bug is real in the direction that matters: `done < <(producer)` throws away
the producer's exit status, so a `git ls-files` failure or an exclusions file
that swallowed every path left the link check reporting success having read
nothing. Both consumers now fail on a failed producer and on an empty one, and
the `count -gt 0` assertion is what makes the empty case loud.

I read the workflow diff in full rather than summarising it, and it changes two
`run:` bodies and nothing else:

```
$ git diff $(git merge-base origin/main p399)..p399 -- .github/workflows/ | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -iE 'on:|permissions|uses:|secrets|pull_request_target|if:'
NONE — no trigger, permission, action-pin, secret or conditional changed
```

One thing I will not do myself: merging this. A pull request that edits
`.github/workflows/**` gets a human reading the diff before it lands, which is
the control that stops a fork's code running in CI with a token, and I hold
myself to it even when the diff is six lines I have already read. It is at the
top of today's queue.

#239 is still yours and I am not going to load a second one on top of it. When
nftables lands, tell me what you want next and I will hold it.

vladimirrott:
My review names `14774391`. The head is `f9a01f0f`, and GitHub recorded the
approval against `f9a01f0f`, which is the SHA that counts. You pushed while I was
running the proofs and I quoted the ref I had fetched first. The green board I
quoted was already `f9a01f0f`; the two mutation runs were not, so I repeated them
against the current head:

```
$ git archive p399 | tar -x -C .../trees/mut399c
$ out=$(podman run --rm --network=none -v ".../trees/mut399c:/repo:z" -w /repo -e HOME=/tmp localhost/sk-pygit:1 bash tests/release/markdown-link-files.test.sh 2>&1); rc=$?
CLEAN at f9a01f0f: exit=0
markdown-link-files: derived files and exclusions validated.
```

`ci.yml` internal-links hunk reverted to the process-substitution form:

```
exit=1
AssertionError: internal consumer accepted failed producer
```

Producer reverted to the version on `main`, workflow left alone:

```
M2 (producer reverted to main) exit=1
markdown-link-files: failed git enumeration unexpectedly passed
```

Restored, exit 0 again. The approval holds at `f9a01f0f`.

The new push is the better version of the test, and I would have asked for the
first half of it: `grep -Fq 'markdown-link-files'` matched the script's own name,
so it passed on any error the script printed, and now each case names the message
it expects. Running the workflow body under `bash -e` instead of
`-euo pipefail` matters more than it looks, because GitHub's default shell for a
`run:` block has no `pipefail`, and a guard proven under stricter settings than
production is a guard proven somewhere else.

The workflow-file rule stands: a human reads the diff before this merges.

vladimirrott:
@mikevillari this is ready and the only thing left is one click on your side: **Update branch**. Your fork does not allow maintainer edits, so I cannot do it for you, and the branch protection here requires an up-to-date branch before merge.

Everything else is done. I verified against the merged tree rather than your branch alone, because four PRs have landed since you opened this:

```
main d22f93f merged with f9a01f0f -> 0 conflicts
$ bash tests/release/markdown-link-files.test.sh        # podman, python:3.12, no network
rc=0  markdown-link-files: derived files and exclusions validated.

$ # M1: producer check replaced with `git ls-files ... || true`
rc=1
$ # M2: the count-is-zero refusal deleted
rc=1  markdown-link-files: empty derived set unexpectedly passed
```

Both guards bite, and you put one in the script and one in each workflow step, which is why M1 and M2 fail independently.

One thing I want on the record, because I got it wrong for about ten minutes and someone else reading this diff will too. `git diff main..399` shows your PR appearing to roll `taiki-e/install-action` back four patch releases:

```
-  uses: taiki-e/install-action@e67fa11c…  # v2   (2.87.4, 2026-09-02)
+  uses: taiki-e/install-action@37f7c578…  # v2   (2.87.0, 2026-08-27)
```

That is not you. You branched from `cce5c7a` on 7 September, dependabot bumped that pin on main on 9 September in #405, and a two-dot diff blames the PR for anything main gained since. Your only commit touches zero pin lines, and the three-way merge keeps main's newer pin, which I confirmed on the merged tree. Updating the branch will make the phantom disappear from the diff view too.

I have written this up in my own notes as a review trap, since the reading it invites is "unexplained supply-chain downgrade" and that would have been an unpleasant thing to say to you.

Click Update branch and I will merge it.

vladimirrott:
Merging this without the branch update, and recording why, because I am overriding my own gate and that should never be silent.

`maintainer-merge` refused it:

> #399 is BEHIND its base; its checks describe a different tree. Update the branch and let it re-run.

That refusal is correct in general, and I could not clear it the normal way: your fork has `maintainerCanModify=false`, so `PUT /pulls/399/update-branch` returns 403 and I cannot press the button for you.

So I answered the substance instead of waiving it. Your diff touches four files and no Rust:

```
.github/workflows/ci.yml
scripts/ci-local.sh
scripts/markdown-link-files.sh
tests/release/markdown-link-files.test.sh
```

The only required checks those can affect are `docs-and-hygiene` and the shellcheck sweep. I built the tree that merging actually produces, against current main `afc867f`, and ran both:

```
merged afc867f + f9a01f0f -> d046a18, 0 conflicts

$ bash tests/release/markdown-link-files.test.sh          # podman, python:3.12, no network
C0  rc=0  markdown-link-files: derived files and exclusions validated.
M1  rc=1  producer check replaced with `|| true`
M2  rc=1  count-is-zero refusal removed: "empty derived set unexpectedly passed"
C1  rc=0

$ shellcheck -x -S warning over scripts/ tests/ .githooks/    # merged tree
rc=0
```

`rust`, `frontend`, `security-audit` and `postgres-contract` read no file your diff changes, so their green on `f9a01f0f` still describes the merged tree for the parts that matter.

The `taiki-e/install-action` lines that a two-dot diff shows you reverting are a stale-branch artefact, not a change: your only commit touches zero pin lines, and the merged tree carries main's newer `e67fa11c` on both. I checked that by grepping the merged tree rather than reasoning about it, because the alternative reading is "unexplained supply-chain downgrade" and I did not want to leave that sitting in the record next to your name.

Both guards are yours and both bite independently, which is the part I liked: the count check in the script and the count check in each workflow step fail for different reasons.

PR #410, opened 2026-09-09
fix(ci): fail closed when staged secret enumeration fails

### LINKED ISSUE #406: check_no_secrets.sh --staged scans nothing and exits 0 when its git call fails
`.githooks/pre-commit:16` runs `scripts/check_no_secrets.sh --staged`. It is the
only thing standing between a provider credential and a commit object, and the
file's own header says why: CI's TruffleHog runs on `pull_request` and only with
`--only-verified`, so a direct push to main is unscanned until the weekly sweep,
and a revoked-but-real key verifies as nothing.

That path builds its file list through process substitution:

```
$ sed -n '79,82p' scripts/check_no_secrets.sh
    while IFS= read -r path; do
        [ -n "$path" ] || continue
        git show ":$path" 2>/dev/null | scan_text "staged $path" || status=1
    done < <(git diff --cached --name-only --diff-filter=ACMR)
```

`set -euo pipefail` is on line 31 and cannot see a failure inside process
substitution. When `git diff --cached` fails, the loop body never runs, `status`
stays 0, and the scanner exits 0 having read no bytes.

## Measured

Throwaway clone of `83d893d`. A synthetic Groq-shaped key is staged for every
run, so a working scanner has something to find in all three. The harness:

```bash
#!/usr/bin/env bash
# usage: run.sh <label>
cd "$(dirname "$0")/tree" || exit 1
out="$(bash scripts/check_no_secrets.sh --staged 2>&1)"; rc=$?
printf '=== %s === rc=%d\n%s\n' "$1" "$rc" "${out:-<no output>}"
```

```
$ ./run.sh C0-credential-staged
=== C0-credential-staged === rc=1
SECRET: Groq key in staged leak.txt — starts gsk_AAA…, 56 chars

$ cp .git/index index.good && printf 'DIRT' > .git/index
$ git diff --cached --name-only --diff-filter=ACMR; echo "rc=$?"
fatal: .git/index: index file smaller than expected
rc=128

$ ./run.sh M1-git-diff-cached-fails
=== M1-git-diff-cached-fails === rc=0
fatal: .git/index: index file smaller than expected

$ cp index.good .git/index
$ ./run.sh C1-restored
=== C1-restored === rc=1
SECRET: Groq key in staged leak.txt — starts gsk_AAA…, 56 chars
```

C0 and C1 are the calibration. Same tree, same staged credential, refused both
times. The only difference in M1 is that git could not answer, and the `fatal:`
goes to stderr while the hook prints its own green line and lets the commit
through.

A corrupt index is one way in. Any failure of that `git diff` is another, and
the exit code is never read.

## The class, and what it has already cost here

Five sites in this repo build a gate's input list inside process substitution,
where `set -euo pipefail` cannot see the producer fail. Three are unguarded:

| site | state |
|---|---|
| `scripts/check_public_claims.sh:14` | guarded, and the comment at `:24` explains the trap in exactly these terms |
| `scripts/markdown-link-files.sh:48` | #395, fix open in #399 |
| `scripts/check_no_secrets.sh:82` | this issue |
| `tests/release/no-secrets.test.sh:81` | unguarded, and it is this scanner's own test suite: `mapfile -t tracked < <(git ls-files)` feeds `"$CHECK" "${tracked[@]}"`, so a failed `git ls-files` scans zero files and the assertion passes |
| `tests/release/release-rehearsal.test.sh:108` | unguarded, filed as #407 |

The fourth row is the one worth a second look while you are in this file. The
test that proves the scanner works on the whole tracked tree derives that tree
the same way the scanner derives its staged list, so both halves can go quiet
together.

`check_public_claims.sh` already carries the shape the fix wants:

```
$ sed -n '24,29p' scripts/check_public_claims.sh
# `set -euo pipefail` does not see a failure inside process substitution, so a
# rename in check_evidence_claims.py would leave this list empty and the script
# would exit 0 having screened nothing.
((${#claim_files_rel[@]})) || {
    echo 'could not read CLAIM_FILES from check_evidence_claims.py' >&2
    exit 1
}
```

## Scope

- Read the staged list into an array and refuse when `git diff --cached` exited
  non-zero. Distinguish that from an honest empty list: `git commit` with
  nothing staged is a normal case and must stay quiet, while a git that could
  not answer must stop the commit.
- Add a test. `grep -rn -- '--staged' tests/` returns nothing today, so the
  branch the pre-commit hook actually runs has no coverage at all, while the
  explicit-file branch does. One case that breaks the git call and expects a
  refusal, one that stages a real-shaped key and expects the existing message.
- Secondary, and worth a separate commit. When `git show ":$path"` fails the
  script does fail closed, because `pipefail` sees that one. It then prints the
  full "Refusing to commit: the staged content contains what looks like a live
  provider credential" banner with no `SECRET:` line above it, which sends the
  reader looking for a key that was never found. A distinct message for "git
  could not read the staged bytes" costs three lines.

## Difficulty

Easy. One guard, one message, two test cases, no new dependency. The fix in
#399 is the same shape on a neighbouring file and is a good model to read first.

ITSMERNB:
## Summary
- capture the staged file list before scanning and fail closed when `git diff --cached` cannot enumerate it
- keep an honest empty staged set quiet and allowed
- add regression coverage for the real `--staged` pre-commit path
- distinguish a staged-blob read failure from an actual credential finding in a separate commit

## Validation
- reproduced the issue on current `main`: `git diff --cached` exited 128 while the scanner exited 0
- after the fix, the same case exits 1 with `git could not enumerate staged files`
- staged synthetic finding: PASS
- broken-index enumeration failure: PASS
- honest empty staging: PASS
- unreadable staged blob message: PASS
- `bash -n scripts/check_no_secrets.sh tests/release/no-secrets.test.sh`
- `git diff --check`

Closes #406

### DIFF
--- a/scripts/check_no_secrets.sh
+++ b/scripts/check_no_secrets.sh
@@ -75,11 +75,21 @@ scan_text() {
 status=0
 if [ "${1:-}" = "--staged" ]; then
     # The staged content itself, not the working tree: those differ, and it is
-    # the staged bytes that become the commit.
-    while IFS= read -r path; do
+    # the staged bytes that become the commit. Capture the producer status
+    # explicitly: strict mode cannot see failures inside process substitution.
+    staged_list=""
+    if ! staged_list="$(git diff --cached --name-only --diff-filter=ACMR)"; then
+        echo "Refusing to commit: git could not enumerate staged files." >&2
+        exit 1
+    fi
+    staged_paths=()
+    if [ -n "$staged_list" ]; then
+        mapfile -t staged_paths <<< "$staged_list"
+    fi
+    for path in "${staged_paths[@]}"; do
         [ -n "$path" ] || continue
         git show ":$path" 2>/dev/null | scan_text "staged $path" || status=1
-    done < <(git diff --cached --name-only --diff-filter=ACMR)
+    done
 else
     for path in "$@"; do
         [ -f "$path" ] || continue
--- a/scripts/check_no_secrets.sh
+++ b/scripts/check_no_secrets.sh
@@ -88,7 +88,12 @@ if [ "${1:-}" = "--staged" ]; then
     fi
     for path in "${staged_paths[@]}"; do
         [ -n "$path" ] || continue
-        git show ":$path" 2>/dev/null | scan_text "staged $path" || status=1
+        staged_content=""
+        if ! staged_content="$(git show ":$path" 2>/dev/null)"; then
+            printf 'Refusing to commit: git could not read staged bytes for %s.\n' "$path" >&2
+            exit 1
+        fi
+        printf '%s' "$staged_content" | scan_text "staged $path" || status=1
     done
 else
     for path in "$@"; do
--- a/tests/release/no-secrets.test.sh
+++ b/tests/release/no-secrets.test.sh
@@ -154,6 +154,21 @@ else
     echo "FAIL: --staged rejected an honest empty staged set"
     fail=1
 fi
+
+# --- 6. A staged blob read failure must not masquerade as a finding ----------
+fake_oid='1111111111111111111111111111111111111111'
+git -C "$staged_repo" update-index --add --cacheinfo 100644,$fake_oid,unreadable.txt
+if unreadable_output="$(cd "$staged_repo" && "$CHECK" --staged 2>&1)"; then
+    echo "FAIL: --staged passed when git could not read staged bytes"
+    fail=1
+elif ! grep -Fq 'could not read staged bytes for unreadable.txt' <<< "$unreadable_output"; then
+    echo "FAIL: --staged did not distinguish an unreadable staged blob"
+    fail=1
+elif grep -Fq 'staged content contains' <<< "$unreadable_output"; then
+    echo "FAIL: unreadable staged bytes were reported as a credential finding"
+    fail=1
+fi
+git -C "$staged_repo" update-index --force-remove unreadable.txt
 if [ "$fail" != 0 ]; then exit 1; fi
 echo "ok: catches real-shaped credentials, ignores this repo's fixtures"
 echo "ok: the whole tracked tree scans clean, and findings never echo the secret"
--- a/tests/release/no-secrets.test.sh
+++ b/tests/release/no-secrets.test.sh
@@ -112,6 +112,48 @@ if ! grep -qF "$leak" <<< "$out"; then
 fi
 CHECK="$real_check"
 
+# --- 5. The pre-commit --staged path must fail closed -----------------------
+staged_repo="$tmp/staged-repo"
+mkdir -p "$staged_repo"
+git -C "$staged_repo" init -q
+git -C "$staged_repo" config user.name 'SysKnife test'
+git -C "$staged_repo" config user.email 'sysknife-test@example.invalid'
+printf 'clean\n' > "$staged_repo/README.md"
+git -C "$staged_repo" add README.md
+git -C "$staged_repo" commit -qm 'fixture baseline'
+
+printf 'k = "%s"\n' "$leak" > "$staged_repo/leak.txt"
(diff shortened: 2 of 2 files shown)


ITSMERNB:
Thanks again, Vladimir, for pointing me at this one and for the correction on #331. I reproduced the exact fail-open before changing anything, kept the main fix scoped to the real `--staged` hook path, and put the staged-blob read diagnostic in its own commit as suggested. The full GitHub CI is green now. Ready for your review.

vladimirrott (approved):
Reviewed at `1129ec86`. Approving.

You built the guard in both directions, which is the part I care about and the part that usually gets left out. Four mutations against your own test, each one caught and each one naming which direction broke. All of them ran in a container with the tree exported at your head:

```
M1  remove the enumeration guard        rc=1  FAIL: --staged passed when git could not enumerate the index
M2  remove the blob-read guard          rc=1  FAIL: --staged passed when git could not read staged bytes
M3  refuse an empty staged set          rc=1  FAIL: --staged rejected an honest empty staged set
M4  delete the scan from the branch     rc=1  FAIL: --staged accepted content that the explicit-file path rejects
C0/C1 unmutated                         rc=0  ok: catches real-shaped credentials, ignores this repo's fixtures
```

M3 is the one I want to name. A guard that refuses whenever it is unsure is easy to write and useless in practice, because `git commit` with nothing staged is a normal thing to do and a hook that shouts at it gets disabled within a week. You separated "asked, the answer is nothing" from "could not ask" and pinned both.

M2 went past what the issue asked for. I raised the misleading banner as a secondary item and you gave it a distinct message and its own negative twin, asserting that an unreadable blob is not reported as a credential finding. That last assertion is the useful half.

Two things that are mine, not yours:

The `--staged` branch had no coverage at all before this, while the explicit-file branch did. That is the gap that let the defect live since `bf8d412`, and it was mine to notice.

The merge gate could not run your test at first. Its shell suite runs in `python:3.12-slim`, which carries bash and python3 and no git, and its declared needs said `bash python3`. Your test builds a throwaway repository, so the gate would have watched git fail and reported you as failing your own test. That is the second time that image was chosen for what the scripts needed on the day it was written. Fixed on my side, and the suite now refuses a `-slim` image for this by name.

One observation, no action needed. `git show ":$path"` failing now stops the commit, which is right. It would also stop a commit in a repository with a gitlink, since `git show` cannot resolve a submodule entry that way. This repo has no `.gitmodules`, so nothing changes today; worth remembering if one ever arrives.

Merging once the board settles. I will write the CHANGELOG entry at merge.

vladimirrott:
Merged as `1129ec86`, closing #406.

This is the first pull request in this repository whose receipt the merge gate **observed** rather than took my word for. It ran your test in a container at your head, applied the mutation itself, and recorded the differential:

```
running 'tests/release/no-secrets.test.sh' unmutated
receipt recorded for #410 at 1129ec86 (observed: clean pass, mutated fail)
```

Every previous merge here carried an asserted receipt, meaning I ran the proof by hand and the gate recorded my claim. Getting to an observed one took three fixes on my side that your test exposed, and I would rather you knew that than think the gate had been working all along:

- its container image had no `git`, so your `git init` fixture could not run;
- it mounted `/tmp` with `noexec`, so your echoing-scanner stub could not execute and the gate reported you as failing your own test;
- both were reported to me as "the test does not pass unmutated", which is the failure mode where my tooling blames the contributor.

Your test found two of my bugs on its way to proving your own fix. That is a better outcome than the fix alone.

I wrote the CHANGELOG entry at merge, under Unreleased.

ITSMERNB:
That is a very satisfying result. Thanks for the detailed write-up, Vladimir — especially for fixing the gate rather than treating its first verdict as authoritative. Having the test expose two problems in the verification path on the way to proving the actual fix is exactly the kind of feedback loop I want from this work. I appreciate the review and the merge.

PR #415, opened 2026-09-10
fix(ufw): expose numbered rules before deletion

### LINKED ISSUE #234: UfwDeleteRule needs a rule number that no action in the catalogue can produce

QinXi-ai:
`UfwDeleteRule` requires a rule index, but SysKnife could only run `ufw status verbose`, which does not return indices. Add optional `numbered: true` to `UfwStatus` and a read-only `query_ufw_rules` planning tool, so the model can obtain the index before proposing deletion. The direct MCP UfwStatus tool already forwards action parameters. Omitted/false keeps the verbose output; non-boolean values are rejected.

Fixes #234. Updates the catalogue, prompt parameter guidance, generated action reference and user documentation. Existing `ufw_status()` callers and deletion approval remain unchanged; no new privilege grant is needed. Listing and deleting are not atomic, so guidance explicitly requires refreshing indices after rule changes. This is an additive, non-breaking change.

Validation:
- Reproduced the missing numbered argv with a failing regression before implementation.
- Actual-source standalone harness: 31 ufw tests, 11 query-tool tests and 19 prompt tests pass. The harness includes the actual UfwStatus executor arm; it is not a full workspace run.
- Standalone Clippy, workspace rustfmt, diff checks and evidence-claim guard pass. Action reference regenerated from the actual catalogue/generator.
- Extended existing tests; no Rust test or action count change.
- Read story-125 and preserved its explicit `rule_number: 3` deletion path. The Linux workspace Rust CI (including Clippy and tests), docs-and-hygiene, scripts-lint, frontend, Postgres contract and security checks now pass. Live VM/LLM story-125 could not run locally on Windows; no claim of live firewall or LLM validation.

Codex assisted implementation and validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Releases before `0.2.5` predate the public launch; their notes live in the
 
 ### Changed
 
+- Let `UfwStatus` return numbered rules with `numbered: true`, retaining verbose
+  output by default. Add `query_ufw_rules` so the planner can read the indices
+  required by `UfwDeleteRule` instead of guessing them (#234).
+
 - Separate Ubuntu identity requirements from Debian-family mechanisms and
   planner defaults. Canonical services, PPAs and the reboot sentinel require
   Ubuntu itself; portable tools are no longer refused merely for being another
--- a/crates/sysknife-brain/src/planning_tools/propose_plan.rs
+++ b/crates/sysknife-brain/src/planning_tools/propose_plan.rs
@@ -357,7 +357,7 @@ this is runtime status, NOT the saved configuration; on Ubuntu the saved config 
     ("UfwReset",
      "reset ufw to defaults, removing all rules — no params; Ubuntu only; High risk; irreversible"),
     ("UfwStatus",
-     "show current ufw status and rules — no params; Ubuntu only; read-only"),
+     "show current ufw status and rules — optional param: numbered (boolean, default false); true runs ufw status numbered and exposes rule_number values for UfwDeleteRule, false keeps verbose status; read-only"),
     // ── Ubuntu / distrobox — container environment ────────────────────────────
     ("DistroboxList",
      "list distrobox containers — no params; Ubuntu only; read-only"),
@@ -434,7 +434,7 @@ reports live interface state"),
      "regenerate netplan backend config without applying — no params; Ubuntu only; Medium risk; dry-run before NetplanApply"),
     // ── Ubuntu / Tier 3 — ufw extensions ─────────────────────────────────────
     ("UfwDeleteRule",
-     "delete a ufw rule by number — param: rule_number* (positive integer from 'ufw status numbered'); Ubuntu only; High risk"),
+     "delete a ufw rule by number — param: rule_number* (positive integer from query_ufw_rules or UfwStatus with numbered=true); never guess a rule number, and refresh after rule changes; High risk"),
     ("UfwLimit",
      "add rate-limiting rule on a port/service (>6 connections/30s blocked) — param: target* (e.g. '22' or 'ssh'); Ubuntu only; High risk; use for SSH brute-force mitigation"),
     // ── Ubuntu / Tier 3 — release upgrade ────────────────────────────────────
--- a/crates/sysknife-brain/src/planning_tools/query_tools.rs
+++ b/crates/sysknife-brain/src/planning_tools/query_tools.rs
@@ -9,6 +9,11 @@ pub fn query_tools() -> Vec<ToolDefinition> {
     let empty_schema = serde_json::json!({"type": "object", "properties": {}, "required": [], "additionalProperties": false});
     vec![
         ToolDefinition {
+            name: "query_ufw_rules".into(),
+            description: "Read ufw status numbered to obtain current rule indices for UfwDeleteRule. Never guess an index; query again after any rule change because indices can shift.".into(),
+            input_schema: empty_schema.clone(),
+        },
+        ToolDefinition {
             name: "query_services".into(),
             description: "List all running systemd services. Returns one service name per line."
                 .into(),
@@ -225,6 +230,7 @@ pub fn query_tool_to_action(
     input: &serde_json::Value,
 ) -> Result<Option<(&'static str, serde_json::Value)>, String> {
     match tool_name {
+        "query_ufw_rules" => Ok(Some(("UfwStatus", serde_json::json!({"numbered": true})))),
         "query_services" => Ok(Some(("ListServices", serde_json::json!({})))),
         "query_firewall" => Ok(Some(("GetFirewallState", serde_json::json!({})))),
         "query_deployments" => Ok(Some(("ListDeployments", serde_json::json!({})))),
@@ -298,6 +304,10 @@ mod tests {
     #[test]
     fn known_query_tools_map_to_actions() {
         let empty = empty_input();
+        assert_eq!(
+            query_tool_to_action("query_ufw_rules", &empty),
+            Ok(Some(("UfwStatus", serde_json::json!({"numbered": true}))))
+        );
         assert_eq!(
             query_tool_to_action("query_services", &empty),
             Ok(Some(("ListServices", serde_json::json!({}))))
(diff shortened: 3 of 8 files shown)


vladimirrott (dismissed):
Reviewed at `95cc69af`. Approving.

This is the shape I want for anything that touches an argv the daemon hands to sudo, and you got there without being told.

**No operator-controlled string reaches the command line.** `ufw_status_with_numbered` selects between two fixed literals, so there is nothing to validate at the boundary because there is nothing free-form to begin with. That is stronger than validating it well.

**The parameter fails closed on type confusion.** `numbered` absent means false, a bool means itself, and anything else is `InvalidParam("numbered")`. You tested `"true"`, `1` and `null` explicitly, which is the set a model actually emits when it gets a schema slightly wrong. A permissive `as_bool().unwrap_or(false)` would have passed review at a glance and been wrong.

**The sudo grant already covers it, and I checked rather than assumed:**

```
$ grep -n 'ufw' packaging/sysknife-sudoers
181:sysknife ALL=(root) NOPASSWD: /usr/sbin/ufw
```

That is a wildcard on the whole binary, so `ufw status numbered` needs no packaging change and you were right not to touch it. Worth naming what that means, since it is the reason your strict parse matters more than it looks: the safety of `UfwStatus` rests entirely on `build_action_spec` never emitting a mutating argv, because sudo will run whatever ufw subcommand reaches it. Your change narrows that surface rather than widening it.

**The planning side is coherent.** `query_ufw_rules` carries an empty schema with `additionalProperties: false` and `query_tool_to_action` hardcodes `{"numbered": true}`, so the model cannot steer it. Moving `UfwStatus` out of the Debian prompt's "No params" list and giving `UfwDeleteRule` a sentence about where the number comes from and that indices shift after a change closes the loop #234 opened: the number was previously unobtainable through any catalogued action.

The tool count assertion moved 23 to 24 and now asserts the new tool's name, its description content and its schema, rather than only the number. That is the difference between a count that moves and a count that means something.

## Before merge

No closing keyword in the body, so merging leaves #234 open. It is the issue this closes; `Closes #234` in the body will do it.

## A follow-up, not yours to fix here

You dropped "Ubuntu only" from the two descriptions you touched, which is correct after #384 moved ufw out of the Ubuntu fence. Six of the eight are still stale:

```
UfwEnable YES  UfwDisable YES  UfwAllow YES  UfwDeny YES
UfwReset YES   UfwLimit YES    UfwStatus no  UfwDeleteRule no
```

With #412 making Debian eligible, all eight are in the Debian catalogue, so a Debian host now reads "Ubuntu only" against six actions it can run. I will file it rather than expand this PR.

vladimirrott:
Resolved the conflict on your branch and pushed to it, as I said in the review I would.

`CHANGELOG.md` only. #238's entry landed on main while this was open and both entries claim the same section; I kept both with main's first. `crates/sysknife-brain/src/planning_tools/propose_plan.rs` auto-merged against #414's change to the same file.

Your production contribution is unchanged, checked rather than assumed. Same diff over `crates/*/src/*`, first 16 hex of the sha256 with blob index headers stripped:

```
before the resolution (merge-base..95cc69af):  6126773cde0e5a62
after  the resolution (origin/main..e7e4f69a): 6126773cde0e5a62
```

And the guard still bites on the merged tree. Making the parse permissive:

```
$ # replace the strict match with .and_then(|v| v.as_bool()).unwrap_or(false)
$ cargo test --offline --locked -p sysknife-daemon --lib -- ufw_status_uses_verbose
panicked at crates/sysknife-daemon/src/actions/ufw.rs:366:
    Err(crate::executor::ExecutorError::InvalidParam("numbered")))
test result: FAILED. 0 passed; 1 failed; 897 filtered out
```

Unmutated, `1 passed`. That is the assertion I care about most in this PR: a permissive `unwrap_or(false)` would read fine and silently accept `{"numbered": "true"}`.

The push dismissed my approval, as `dismiss_stale_reviews` requires. Re-approving at `e7e4f69a` and merging once the board is green.

vladimirrott (approved):
Re-approving at `e7e4f69a`, the head I resolved and re-verified. Production diff byte-identical to the head I approved, and the strict-parse guard still fails when made permissive. Details in the comment above.

vladimirrott:
Merged as `0533007b`, closing #234.

Four pull requests from you in one night and three of them are in. The one I keep coming back to is the strict parse:

```rust
let numbered = match params.get("numbered") {
    None => false,
    Some(value) => value.as_bool().ok_or(ExecutorError::InvalidParam("numbered"))?,
};
```

A permissive `.and_then(|v| v.as_bool()).unwrap_or(false)` reads identically at a glance, passes every happy-path test, and silently accepts `{"numbered": "true"}` from a model that got the schema slightly wrong. You tested the string, the integer and the null explicitly, which is the set that actually turns up. Making it permissive is the mutation the receipt records, and it fails at `ufw.rs:366`.

The other thing worth naming: `#234` had been open since 17 August because the rule number it needs was not obtainable through any catalogued action. `query_ufw_rules` closes that loop, and pinning the tool count at 24 while also asserting the new tool's schema and description is the difference between a number that moves and a number that means something.

I resolved the `CHANGELOG.md` conflict on your branch rather than making you rebase a fourth time; details are in the comment above, including the hash showing your production diff came through unchanged.

Left for you: **#413** has one blocking item, and it is a good one to fix rather than a nuisance. The helper-coverage carve-out asserts the reporter's mechanism but never asserts the sudoers grant is *absent*, so the property your whole design rests on is the one thing nothing checks. I added the grant and the guard stayed green. One line, and I ran it both ways for you. The baseline bump is the other half.

**#416** and **#346** are held for you alongside **#417**, and the batch note is on #417. No hurry on any of them.

Issue #409, opened 2026-09-09
The ShellCheck job passes when a search root disappears, and it never lints the two git hooks


vladimirrott:
`.github/workflows/e2e.yml:77-81` is the only ShellCheck gate in CI, and it
reads its file list from the left-hand side of a pipe:

```yaml
      - name: ShellCheck maintained scripts
        run: |
          find tests/e2e tests/release scripts assets/demo \
            -type f -name '*.sh' -print0 \
            | xargs -0 shellcheck --severity=warning
```

GitHub runs a `run:` block as `bash -e {0}`, and `e2e.yml` sets no
`defaults.run.shell`, so there is no `pipefail`. `find`'s exit status is
discarded and the step's verdict comes from `xargs` alone. Rename or move one of
those four roots and the scan quietly shrinks to the survivors.

## Why it matters

Reproduced in a scratch tree, running the step body exactly as the workflow
writes it. `bad.sh` carries three warnings and lives under `tests/release`:

```
$ d=/tmp/sk-shellcheck-repro; mkdir -p "$d"/{tests/e2e,tests/release,scripts,assets/demo}
$ printf '#!/usr/bin/env bash\nfoo=1\ncd $undefined_and_unquoted\necho "$fooo"\n' > "$d/tests/release/bad.sh"
$ printf '#!/usr/bin/env bash\necho ok\n' > "$d/scripts/good.sh"
$ cd "$d"
$ out="$(bash -e -c "find tests/e2e tests/release scripts assets/demo -type f -name '*.sh' -print0 | xargs -0 shellcheck --severity=warning" 2>&1)"; rc=$?
$ echo "rc=$rc"; printf '%s\n' "$out" | head -6
rc=123

In tests/release/bad.sh line 2:
foo=1
^-^ SC2034 (warning): foo appears unused. Verify use (or export if used externally).

$ mv tests/release tests/release-suite
$ out="$(bash -e -c "find tests/e2e tests/release scripts assets/demo -type f -name '*.sh' -print0 | xargs -0 shellcheck --severity=warning" 2>&1)"; rc=$?
$ echo "rc=$rc"; printf '%s\n' "$out"
rc=0
find: ‘tests/release’: No such file or directory
```

Green, with a file carrying three warnings never opened. The whole trace is one
line on stderr inside a passing step.

Two facts bound this honestly, and both are worth knowing before you start.

**The local mirror already catches it.** `scripts/ci-local.sh:265-270` runs the
[...]

Issue #408, opened 2026-09-09
every_sudo_action_is_authorised_by_a_packaged_grant passes when its sudo filter selects nothing


vladimirrott:
`crates/sysknife-daemon/tests/sudoers_grants_match_argv.rs:92` walks the whole
action catalogue, keeps the specs that shell out through `sudo`, and asserts
that every one of their argvs is authorised by a rule in
`packaging/sysknife-sudoers`. It guards the grants side of that comparison and
leaves the actions side open:

```rust
let grants = load_grants();
assert!(
    !grants.is_empty(),
    "parsed zero grants — the sudoers parser or file layout changed"
);

let mut unauthorised = Vec::new();
for (_section, specs) in catalogue() {
    for spec in specs {
        let ActionMechanism::Command { program, args } = &spec.mechanism else {
            continue;
        };
        if *program != "sudo" {
            continue;
        }
        if !grants.iter().any(|g| grant_allows(g, args)) {
            unauthorised.push(format!("{}: sudo {}", spec.action_name, args.join(" ")));
        }
    }
}

assert!(
    unauthorised.is_empty(),
    ...
);
```

`unauthorised` is push-only behind two `continue`s, so the load-bearing
assertion is `Vec::new().is_empty()` whenever the filters keep nothing.

## Why it matters

The filter selects on the bare literal `"sudo"`, and 91 specs match it today:

```
$ git rev-parse --short=8 HEAD
5673d20f
$ grep -rhoE 'command_mechanism\(\s*"sudo"' crates/sysknife-daemon/src/actions/ | wc -l
79
$ grep -rhoE 'program: "sudo"' crates/sysknife-daemon/src/actions/ | wc -l
12
```

Write `"/usr/bin/sudo"` in those specs and the selection drops to zero, the test
stays green, and 91 privileged argvs stop being compared against the packaged
sudoers file. That edit is not hypothetical: the sibling test in the same file,
`matcher_rejects_a_bare_binary_name_against_an_absolute_argument_token`
(`:121`), exists because a bare binary name was once matched against an
absolute path. Hardening the program field to an absolute path is the natural
follow-up and would disarm this guard in the same commit.

Disarming the filter, measured here:

```
$ out="$(cargo nextest run -p sysknife-daemon --test sudoers_grants_match_argv --locked 2>&1)"; rc=$?
$ echo "rc=$rc"; printf '%s\n' "$out" | grep -E 'Summary|Starting'
rc=0
[...]

PR #418, opened 2026-09-10
fix(repo): enforce LF line endings

### LINKED ISSUE #342: No .gitattributes, so a Windows or WSL clone gets CRLF and a byte-exact doc test blames the action catalogue

ITSMERNB:
## Summary
- add repository-level LF policy for tracked text while excluding the existing PNG/ICO/GIF binary assets
- keep shell scripts and extensionless Git hooks explicitly on LF so their shebangs stay executable
- add a release guard that fails if any tracked index entry is CRLF or mixed, and run it in CI and `ci-local`
- document the Windows/WSL checkout and renormalisation steps

## Proof first
Before adding `.gitattributes`, I committed a CRLF fixture on a local scratch branch. `git ls-files --eol` reported:

```text
i/crlf  w/crlf  attr/    tests/evidence/crlf-guard-proof.txt
```

and `tests/release/tracked-eol.test.sh` failed with exit 1 and named that file. The scratch branch was then deleted.

With the new attributes present, staging a CRLF working-tree fixture produced `i/lf w/crlf attr/text=auto eol=lf`, proving the repository policy normalises the index rather than only making the guard green.

I also compared raw working-tree hashes for the tracked PNG, GIF and ICO samples against their index OIDs; all three stayed byte-identical under the binary exclusions.

## Validation
- `bash tests/release/tracked-eol.test.sh`
- `bash -n tests/release/tracked-eol.test.sh scripts/ci-local.sh`
- `bash scripts/check_repo_completeness.sh`
- `bash tests/release/markdown-link-files.test.sh`
- `git diff upstream/main...HEAD --check`

`npm test --prefix packages/setup` is not green on native Windows: 111 pass, 14 fail, 1 skip. The failures are existing POSIX assumptions around Unix modes, `process.getuid`, `/run` paths and executable discovery; this PR does not touch `packages/setup`. Linux CI is the authoritative run for those tests.

Closes #342

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,11 @@
+# Keep tracked text byte-stable across Linux, Windows, and WSL checkouts.
+* text=auto eol=lf
+
+# Shell entry points must keep an LF shebang even on Windows.
+*.sh text eol=lf
+.githooks/* text eol=lf
+
+# Binary assets must never be line-ending normalized.
+*.gif -text
+*.ico -text
+*.png -text
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,9 @@ jobs:
       - name: Check no workflow pins an end-of-life Node release
         run: bash tests/release/node-eol.test.sh
 
+      - name: Check tracked files use portable line endings
+        run: bash tests/release/tracked-eol.test.sh
+
       - name: Check systemd directory modes match tmpfiles
         run: bash tests/release/systemd-directory-modes.test.sh
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,6 +78,24 @@ A one-line comment prevents it. If an issue already carries `claimed` and the
 thread has been quiet for a week, say you are taking it over rather than opening
 a competing PR.
 
+**Windows and WSL line endings.** This repository pins tracked text to LF via
+`.gitattributes`. On Windows or WSL, configure this checkout with:
+
+```sh
+git config core.autocrlf false
+```
+
+If an existing checkout already has CRLF line endings, first make sure the
+working tree is clean, then renormalise it:
+
+```sh
+git config core.autocrlf false
+git rm --cached -r . && git reset --hard
+```
+
+`git reset --hard` discards local changes, so do not run the recovery sequence
+with work you have not committed or stashed.
+
 ### 2. Branch, code, test
 
 ```sh
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -271,6 +271,7 @@ run_hygiene_group() {
     run_step 'hygiene: release-rehearsal.test.sh' bash "$repo_root/tests/release/release-rehearsal.test.sh"
     run_step 'hygiene: database-path-agreement.test.sh' bash "$repo_root/tests/release/database-path-agreement.test.sh"
     run_step 'hygiene: node-eol.test.sh' bash "$repo_root/tests/release/node-eol.test.sh"
+    run_step 'hygiene: tracked-eol.test.sh' bash "$repo_root/tests/release/tracked-eol.test.sh"
     run_step 'hygiene: systemd-directory-modes.test.sh' bash "$repo_root/tests/release/systemd-directory-modes.test.sh"
     run_step 'hygiene: ubuntu-vm-bootstrap.test.sh' bash "$repo_root/tests/e2e/ubuntu-vm-bootstrap.test.sh"
     run_step 'hygiene: provider-parity.test.sh' bash "$repo_root/tests/e2e/provider-parity.test.sh"
--- a/tests/release/tracked-eol.test.sh
+++ b/tests/release/tracked-eol.test.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# Keep tracked text normalized in the Git index.
+#
+# A Windows or WSL checkout can otherwise follow contributor-specific line-ending
+# settings. Without a repository policy, CRLF can enter the index. Some SysKnife
+# tests compare generated documents byte-for-byte, so that corruption looks like
+# an unrelated content failure.
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "$0")/../.." && pwd)"
+cd "$ROOT"
+
+if ! eol="$(git ls-files --eol)"; then
+    echo "FAIL: could not inspect tracked file line endings" >&2
+    exit 1
+fi
+
+bad="$(printf '%s\n' "$eol" | grep -E '^i/(crlf|mixed)[[:space:]]' || true)"
+if [ -n "$bad" ]; then
+    echo "FAIL: tracked files contain CRLF or mixed line endings in the Git index:" >&2
+    printf '%s\n' "$bad" >&2
+    exit 1
+fi
+
+echo "ok: tracked Git index contains no CRLF or mixed line endings"


vladimirrott (approved):
Reviewed at `c2a0c379`. Approving.

You did the thing I most wanted to see: you wrote the proof before the fix, on a scratch branch, and reported what `git ls-files --eol` said. I reproduced it, and the first time I got it wrong in a way worth writing down.

My first mutation added a CRLF file with `git add` and the guard stayed green. That was not a hole, it was your `.gitattributes` doing its job: git normalised the file on the way into the index, so `i/lf w/crlf`. To test the guard rather than the policy I had to write the blob past the filter:

```
$ oid="$(printf 'a\r\nb\r\n' | git hash-object -w --stdin)"
$ git update-index --add --cacheinfo 100644,"$oid",crlf-fixture.txt
$ git ls-files --eol crlf-fixture.txt
i/crlf  w/      attr/text=auto eol=lf   crlf-fixture.txt

$ bash tests/release/tracked-eol.test.sh
FAIL: tracked files contain CRLF or mixed line endings in the Git index:
i/crlf  w/      attr/text=auto eol=lf   crlf-fixture.txt
rc=1
```

Removed, `rc=0`. So the policy prevents it and the guard catches it if it ever arrives another way, which is the pair you want.

`if ! eol="$(git ls-files --eol)"` is the shape from your #410, applied here without being asked. Checking the **index** rather than the working tree is the correct choice and the one most people get wrong; a clean worktree with a CRLF blob committed is exactly the state that breaks a byte-exact doc test.

The `.gitattributes` itself is careful. `*.gif -text`, `*.ico -text`, `*.png -text` before anything can normalise them, and `.githooks/*` explicitly on LF so a shebang survives a Windows checkout. Wiring it into both `ci.yml` and `ci-local.sh` in the same hygiene group as its neighbours is right.

The `CONTRIBUTING.md` note earns its place by warning that `git reset --hard` discards uncommitted work. A recovery recipe without that line costs somebody an afternoon.

## One follow-up, deliberately not blocking

The guard has no count assertion, so it passes over an empty enumeration:

```
$ sed -i 's|git ls-files --eol|true|' tests/release/tracked-eol.test.sh
$ bash tests/release/tracked-eol.test.sh
ok: tracked Git index contains no CRLF or mixed line endings
rc=0
```

I checked whether that can actually happen here before deciding what to do with it. `git ls-files --eol` returns non-empty for any repository with tracked files and fails outright outside one, which your `if !` already catches, so there is no reachable path to a vacuous pass in this repo. That makes it a consistency item rather than a defect, and blocking a complete PR on a hypothetical is not something I want to do to you.

It is the same shape as #408 and #409, both of which you are holding. Fold a `count` into this when you get to those, or leave it; either is fine.

Merging.

vladimirrott:
Merged as `afc867fb`, closing #342. Two hours from the offer to merged, which is the fastest turnaround this tracker has had.

The `if ! eol="$(git ls-files --eol)"` is the shape from your own #410, applied here without anyone asking for it, and checking the **index** rather than the working tree is the choice most people get wrong.

I wrote the CHANGELOG entry at merge, and I checked that I had actually written it this time rather than only saying so: I told you the same thing on #410 and had not, which I noticed during a sweep and fixed in `8b91974`.

One thing from my own verification worth passing on, since it cost me two wrong measurements. My first mutation added a CRLF file with `git add` and your guard stayed green. That was not a hole, it was your `.gitattributes` normalising the file on the way in, so I was testing the policy and calling it the guard. Writing the blob past the filter is what actually exercises it:

```
$ oid="$(printf 'a\r\nb\r\n' | git hash-object -w --stdin)"
$ git update-index --add --cacheinfo 100644,"$oid",crlf-fixture.txt
$ bash tests/release/tracked-eol.test.sh
FAIL: tracked files contain CRLF or mixed line endings in the Git index:
i/crlf  w/      attr/text=auto eol=lf   crlf-fixture.txt
```

Then my second attempt measured a tree still carrying the first mutation, so both had to be redone on a fresh export. Your proof-before-fix on a scratch branch avoided both traps by construction.

#408 and #409 are still held for you, and the count assertion from the review folds naturally into either.

ITSMERNB:
Thanks, Vladimir. I appreciate you reproducing both sides independently and writing down the two measurement traps as well. The index-vs-working-tree distinction was the part I most wanted to pin down here, so seeing the policy normalize normal adds while the guard still catches a bypassed CRLF blob is a good closure. Thanks for the review and merge.

Issue #218, opened 2026-08-17
Make the syslog forwarder's enterprise number configurable (currently the RFC 5612 documentation PEN)


vladimirrott:
The syslog audit forwarder emits RFC 5424 structured-data frames whose SD-ID is hardcoded to the IANA documentation PEN:

```rust
// crates/sysknife-daemon/src/audit_forward.rs:309
let sd = format!(
    "[sysknife@32473 \
     seq=\"{}\" ...
```

`32473` is RFC 5612's reserved documentation and test Private Enterprise Number. The code comment above the format string already flags this:

> **PEN 32473 is RFC 5612's documentation/test PEN, it MUST be replaced before production deployment.** RFC 5424 §7.2.2 requires an IANA-assigned Private Enterprise Number on SD-IDs emitted in production frames. [...] Operators running a SIEM under a regulated framework should either patch `SD@32473` to their own PEN at build time or treat this as a known gap (see issue tracker).

The comment points at "the issue tracker", and no such issue exists. This is that issue.

## The problem

`SyslogForwardSection` in `crates/sysknife-core/src/config.rs:207` carries only `host` and `facility`. There is no way to set the PEN short of editing the source and rebuilding. An operator forwarding SysKnife's audit trail into a compliance SIEM ships frames tagged with a documentation PEN, which a strict RFC 5424 parser is entitled to reject and which collides with anyone else using the reserved number.

## Proposed change

Add an optional `enterprise_number` field to `SyslogForwardSection`, default `32473`, and thread it into the SD-ID rather than hardcoding the literal. Reject a value of 0 and anything outside the u32 IANA range at config-parse time, next to the existing `facility` validation.

- `crates/sysknife-core/src/config.rs:207`: the new field and its default.
- `crates/sysknife-daemon/src/audit_forward.rs:309`: build `sysknife@{pen}` from the config value.
- The test sites that assert the literal `sysknife@32473` (`:483`, `:721`, `:722` and the surrounding frame checks) become assertions over the configured value, with one case at the default and one at a custom PEN.

Keep the comment, reworded: the default stays the documentation PEN, so the warning about production use still holds, but an operator now has a config knob instead of a source patch.

## Difficulty

`medium`. One config field, one format string, and the frame is already byte-for-byte tested so a golden vector at a custom PEN pins the behaviour.

## Getting started

[...]

PR #422, opened 2026-09-10
fix(daemon): report the first differing line when the action reference doc drifts

### LINKED ISSUE #345: action_reference_doc_is_current prints two 44 KB documents instead of the line that differs
> **Re-measured 2026-09-09 at `5673d20`.** Two figures moved and are corrected
> throughout the body below: the `assert_eq!` sits at line **163**, and
> `docs/action-reference.md` is **44,163** bytes. Still 399 lines, and the
> defect is unchanged. #384 moved the line.
>
> ```
> $ git rev-parse --short=8 HEAD
> 5673d20f
> $ grep -n 'assert_eq!' crates/sysknife-daemon/tests/action_reference_doc.rs
> 163:    assert_eq!(
> $ wc -l -c docs/action-reference.md
>   399 44163 docs/action-reference.md
> ```

`crates/sysknife-daemon/tests/action_reference_doc.rs:163` compares two 44 KB documents with `assert_eq!`:

```rust
let committed = std::fs::read_to_string(&path).unwrap_or_default();
assert_eq!(
    committed, generated,
    "docs/action-reference.md is out of date with the action catalogue. \
     Regenerate: UPDATE_ACTION_REFERENCE=1 cargo test -p sysknife-daemon \
     --test action_reference_doc"
);
```

`docs/action-reference.md` is 399 lines and 44,163 bytes. When it fails, the panic prints both sides, so the contributor scrolls 88 KB of near-identical Markdown to find the line that moved.

## Why it matters

The message names the fix, which is the important half and already right. What it does not say is *what* differs, and that decides whether the contributor trusts it. Someone who added one action sees a wall of text and cannot tell at a glance whether the regeneration is safe to accept or whether they broke something else.

It also misreports a mangled checkout. A CRLF working tree fails this assertion on line endings alone and says the catalogue is out of date, which is how #341 came to report it as a pre-existing failure on `main`. That half belongs to the line-endings issue; this one is about the message.

## Scope

Fail with the first differing line and its number, then the regeneration command. Something like:

```
docs/action-reference.md is out of date with the action catalogue.
First difference at line 212:
  committed: | `AptUpgrade` | HIGH | yes |
  generated: | `AptUpgrade` | HIGH | no  |
Regenerate: UPDATE_ACTION_REFERENCE=1 cargo test -p sysknife-daemon --test action_reference_doc
```

`zip` over the two line iterators covers it, plus a length check so a truncated file reports the missing tail rather than nothing.

## Tests first

The interesting assertion is about the message, so drive the comparison helper directly rather than through the file: feed it two strings that differ at a known line and assert the rendered message names that line number and both sides. A test that only checks the equal case proves nothing about the failure path, which is the whole feature.

## Difficulty

`easy`, and self-contained in one file.

## Getting started

jabrailkhalil:
## Summary

`action_reference_doc_is_current` compared the committed `docs/action-reference.md` (44 KB, 399 lines) with the generated document via `assert_eq!`, so a drift printed both sides in full and the contributor had to scroll 88 KB to find the line that moved. It also reported a truncated file as a generic "out of date" without naming the missing tail.

The failure now renders a short message with the first differing line, both sides, and the regeneration command, e.g.:

```
docs/action-reference.md is out of date with the action catalogue.
First difference at line 212:
  committed: | `AptUpgrade` | HIGH | yes |
  generated: | `AptUpgrade` | HIGH | no  |
Regenerate: UPDATE_ACTION_REFERENCE=1 cargo test -p sysknife-daemon --test action_reference_doc
```

Implementation: `first_diff_line(committed, generated)` finds the first 1-based differing line (with a length check so a truncated document reports where the shorter side ends), `diff_message` renders it, and the test panics with that message. The equal case is unchanged.

## Related Issue

Closes https://github.com/lacs-project/sysknife/issues/345

## Validation

All commands ran in a Debian bookworm container (`rust:bookworm`, `x86_64-unknown-linux-gnu`, Rust 1.98.1, cargo-nextest 0.9.143) against the working tree, because `sysknife-daemon` has a Linux-only `vsock` dependency and does not compile on Windows.

- `cargo nextest run -p sysknife-daemon --test action_reference_doc --locked` — 7 tests run, 7 passed (6 new + the existing drift guard).
- Mutation M1 (0-based line number instead of `i + 1`): `first_diff_line_names_a_differing_line`, `diff_message_names_the_line_and_both_sides`, `diff_message_names_the_missing_tail`, and the truncation test go red.
- Mutation M2 (drop the line-count fallback): `first_diff_line_reports_a_truncated_document` and `diff_message_names_the_missing_tail` go red.
- `cargo clippy -p sysknife-daemon --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.
- Full workspace: `cargo nextest run --workspace --locked --no-fail-fast` — 1858 tests run, 1851 passed, 7 failed, 6 skipped. All 7 failures are environmental to this container and reproduce on a pristine `main` checkout (61b3a87) with the same command: three `execute_spec` tests and one `executor` test fail with `ActionNotStopped { program: "sh", ... } (AliveReachable)` (process-group stop confirmation under container semantics), `transactions::tests::concurrent_record_keeps_chain_intact_and_seqs_contiguous` (concurrency timing), `sysknife-core path_behind_mode_000_directory_counts_as_present` and `sysknife-brain current_user_returns_non_empty_string_in_normal_env` (root/container environment). None of them touch this change.

Baseline note: this PR adds 6 Rust tests, so the suite is 1858 vs the recorded 1852 in `tests/evidence/workspace-tests.json`. I did not touch the evidence artifact or the prose counts: `scripts/test_baseline.sh` refuses to record a baseline from a non-zero-exit run, and this machine cannot produce one (the failures above are environment-specific, not fixable here). `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` on CI/Ubuntu should record 1858 and `grep -rn 'Rust tests' README.md docs/introduction.md docs/distro-support.md` updated (the maintainer fallback documented in CONTRIBUTING.md).

## Notes for Reviewers

- New helper functions live in the integration test file per the issue's suggested shape (`zip` over the two line iterators plus a length check).
- The message never echoes the equal prefix (asserted by `diff_message_names_the_line_and_both_sides`), which is the point of the issue: a contributor can see at a glance what changed.
- Windows/WSL note: with a CRLF checkout the message now reports the actual first differing line instead of a wall of text; the line-ending half itself remains the `.gitattributes` issue tracked separately.
- AI disclosure per CONTRIBUTING: the change was implemented with an AI coding agent (assistant reviewed the full diff, ran every command listed above itself, and takes ownership).

### DIFF
--- a/crates/sysknife-daemon/tests/action_reference_doc.rs
+++ b/crates/sysknife-daemon/tests/action_reference_doc.rs
@@ -149,6 +149,50 @@ fn doc_path() -> PathBuf {
     PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../../docs/action-reference.md")
 }
 
+/// First 1-based line at which two documents differ, `None` when they are
+/// byte-identical. When one document is a prefix of the other, reports the line
+/// at which the shorter one ends, so a truncated file is named rather than
+/// silently swallowed.
+fn first_diff_line(committed: &str, generated: &str) -> Option<usize> {
+    if committed == generated {
+        return None;
+    }
+
+    let committed_lines = committed.split('\n').collect::<Vec<_>>();
+    let generated_lines = generated.split('\n').collect::<Vec<_>>();
+    committed_lines
+        .iter()
+        .zip(generated_lines.iter())
+        .position(|(cli, gli)| cli != gli)
+        .map(|i| i + 1)
+        .or_else(|| {
+            (committed_lines.len() != generated_lines.len())
+                .then(|| committed_lines.len().min(generated_lines.len()) + 1)
+        })
+}
+
+/// Failure message naming the first differing line and both sides, or `None`
+/// when the documents match. Regeneration instructions stay in the message so
+/// the failure still says how to fix it.
+fn diff_message(committed: &str, generated: &str) -> Option<String> {
+    let line = first_diff_line(committed, generated)?;
+
+    // `.lines()` has no trailing empty element for a final newline, so a
+    // document that ends here reports "<end of file>" rather than a blank line.
+    let committed_line = committed.lines().nth(line - 1).unwrap_or("<end of file>");
+    let generated_line = generated.lines().nth(line - 1).unwrap_or("<end of file>");
+    Some(format!(
+        "docs/action-reference.md is out of date with the action catalogue.\n\
+         First difference at line {line}:\n\
+         \x20 committed: {}\n\
+         \x20 generated: {}\n\
+         Regenerate: UPDATE_ACTION_REFERENCE=1 cargo test -p sysknife-daemon \
+         --test action_reference_doc",
+        committed_line.trim_end_matches('\r'),
+        generated_line.trim_end_matches('\r'),
+    ))
+}
+
 #[test]
 fn action_reference_doc_is_current() {
     let generated = build_reference();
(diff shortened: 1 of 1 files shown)


jabrailkhalil:
The CI workflows for the current head `b2ba4ad794` are waiting for contributor workflow approval (`action_required`): [ci](https://github.com/lacs-project/sysknife/actions/runs/34517026903), [e2e](https://github.com/lacs-project/sysknife/actions/runs/34517026905), [secret-scan](https://github.com/lacs-project/sysknife/actions/runs/34517026909), [CodeQL](https://github.com/lacs-project/sysknife/actions/runs/34517026838). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

vladimirrott (dismissed):
Reviewed at `b2ba4ad7943b5a163b33467300199ca25dd06aea`.

First, an apology that is mine and not yours. Your four workflow runs have been sitting at `action_required` since 10 September:

```
$ gh run list --branch 'fix/action-reference-diff-message'
ci        completed  action_required  2026-09-10T18:52:52Z
e2e       completed  action_required  2026-09-10T18:52:52Z
CodeQL    completed  action_required  2026-09-10T18:52:52Z
secret-scan completed action_required 2026-09-10T18:52:52Z
```

First-time contributor runs need a maintainer to approve them, nobody did, and you have had four days of a blank check board with no way to tell whether your work passed. I have approved all four and they are running now. That was my delay to fix, and you had no lever to pull.

The change itself is good. Replacing `assert_eq!` on two 44 KB documents with a first-differing-line report is the right call, and the helper is written so it cannot panic on the inputs that usually break this kind of code:

```rust
committed_lines.iter().zip(generated_lines.iter())
    .position(|(cli, gli)| cli != gli)
    .map(|i| i + 1)
    .or_else(|| (committed_lines.len() != generated_lines.len())
        .then(|| committed_lines.len().min(generated_lines.len()) + 1))
```

No indexing anywhere, so a truncated or empty document reports where the shorter side ends instead of panicking on a slice. `diff_message` then uses `.nth(line - 1).unwrap_or("<end of file>")`, and `line` is always at least 1, so the subtraction cannot underflow. I checked the case that worried me, which is whether the guard can now miss real drift: `first_diff_line` returns `None` only when every zipped pair matches and the lengths agree, and `split('\n')`/`join` round-trips, so equal line vectors mean equal strings. It still fails on every genuine difference.

I especially liked that you ran the mutations yourself and named them (M1 on the off-by-one, M2 on the length fallback) rather than asserting the tests were meaningful.

## The baseline is mine, not yours

Your PR adds six tests, taking the suite from 1852 to 1858, and you left `tests/evidence/workspace-tests.json` and the three prose counts untouched. That is exactly right. `CONTRIBUTING.md` says so:

> If you cannot install those, do not estimate a count or edit the evidence metadata by hand. Leave `tests/evidence/workspace-tests.json` and the three prose files untouched. In the PR body, name the command you could not run and the missing tool or system library that blocked it. A maintainer will run the full suite and regenerate all four files before merge.

You named the command, named why your container could not produce a clean run, and named the seven failures you reproduced on a pristine `main` to show they were environmental. I will run `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` on Ubuntu and update all four files in the merge commit. If `rust` goes red on the baseline, that is the expected state and not something for you to chase.

## One optional note

All six new tests call `first_diff_line` and `diff_message` directly, and none of them exercises `action_reference_doc_is_current`. So if someone restored the old `assert_eq!` in the guard while leaving your helpers in place, all six stay green and nothing warns, because the helpers are still reached from the tests. The behavior the PR exists to deliver is the one line that is not pinned. Asserting on the guard's own panic payload, or extracting the compare-and-report into one function both the guard and the tests call, would close that. Not blocking, and the risk is small.

Approving. Once the runs finish I will regenerate the baseline and merge.

vladimirrott (approved):
Re-approving at `228d3d4`. Everything since your `b2ba4ad` is mine, and none of it is a change to your code.

Your six new tests move the workspace count, and in this repository that figure lives in four places that have to move together: `tests/evidence/workspace-tests.json` and one prose literal in `README.md`, `docs/introduction.md` and `docs/distro-support.md`. Bumping the artifact alone turns `rust` green and `docs-and-hygiene` red, which is a confusing way to find out. That coupling is not written down anywhere a first-time contributor would look, so it is not yours to have known.

I measured it rather than adding six to the old number:

```
Summary [11.837s] 1860 tests run: 1860 passed, 6 skipped
test_baseline: recorded 1860 rust tests in tests/evidence/workspace-tests.json
```

and derived the three published figures from the artifact, which the claim screen agrees with:

```
```

I also merged `main` in, since seven pull requests landed here tonight and yours was behind all of them. No conflicts.

On the change itself, the thing I like is that the failure now names a line instead of printing two 44 KB documents at somebody. `action_reference_doc_is_current` was the check most likely to be hit by a contributor who had never seen it, and least likely to tell them what to do about it.

Nothing left for you. I will merge once the board is green.

vladimirrott:
Merged, and it is your first one here.

The failure it replaces used to print two 44 KB documents side by side and leave you to find the difference. It now says which line and shows both versions of it, with the regeneration command still in the message. I hit that check myself last night resolving a conflict in `docs/action-reference.md`, and yours would have saved me the diff.

It went in with an observed receipt, which for this repository means the gate ran your three `first_diff_line` tests in a container, broke the truncation branch underneath them, and watched them go red:

```
rust suite: evidence of 3 executed unit(s), passing unmutated
receipt recorded for #422 (observed: clean pass, mutated fail)
```

Two open issues that fit what you just demonstrated, if you want another. **#433** is one I filed tonight: `scripts/ci-local.sh` prints PASS after a root-only assertion silently skipped, and nothing compares how CI invokes a test with how the local runner does. It is the same shape as your change, a check whose report does not match what it checked. **#218** is smaller: making the syslog forwarder's enterprise number configurable.

Say which one and I will hold it for you.

One last thing, since you now have a merged pull request here: your CI runs will start without me having to approve each push, so the loop gets faster from here.

If you administer Linux machines yourself, the read-only side of SysKnife is worth ten minutes: `sysknife doctor` and the audit commands touch nothing and tell you what the daemon would do. No obligation, and it is not a condition of anything.

Issue #301, opened 2026-08-24
Fedora Atomic: make install has never completed, and it fails halfway


vladimirrott:
> **Recounted at `b2c823e6` on 2026-09-16.** The defect stands: `provision.sh`
> still overrides four path variables and `HELPERS` is still not one of them.
> Three things below have moved, and the transcripts are left as they were run.
>
> - Every count of twelve is wrong now. Fourteen helpers are installed, fourteen
>   `const` strings name them, and thirteen of the fourteen carry a `NOPASSWD`
>   grant across twenty-one grant lines. `firewall-state` has no grant on purpose:
>   `packaging/sysknife-sudoers:71` says so, and the daemon calls it directly
>   rather than through `sudo`.
> - "The Makefile installed two helpers then" is wrong. At `f1d9806` it installed
>   one, `grub-kargs-edit`.
> - **Both jobs under "Not contributor work yet" have landed**, in #309 and in
>   `tests/release/install-paths.test.sh`. See the bottom of this issue.

Ran the Fedora Atomic VM for #217. It does not get as far as a story. `make install` aborts, and the run before it aborts is worse than the abort.

### What happened

Fedora 43 Silverblue, `tests/e2e/provision.sh`, the rpm-ostree branch:

```
install -Dm 755 packaging/sysknife-apt-pin-edit /usr/lib/sysknife/apt-pin-edit
install: cannot create directory '/usr/lib/sysknife': Read-only file system
make: *** [Makefile:74: daemon-install] Error 1

!!! PROVISIONING FAILED at step: make install (rpm-ostree paths)
```

On the guest:

```
/usr/local -> ../var/usrlocal          (writable)
/usr/lib                                (Read-only file system)
```

### The override list was written before the helpers existed

`provision.sh` redirects four path variables on an ostree host:

```
SYSUSERS=/etc/sysusers.d  TMPFILES=/etc/tmpfiles.d
SYSTEMD=/etc/systemd/system  POLKIT=/etc/polkit-1/rules.d
```

`HELPERS ?= /usr/lib/sysknife` is not among them, and `/usr/lib` is exactly what rpm-ostree makes immutable.

Both the variable and that four-entry list shipped in the initial public release, `f1d9806`, 2026-06-15. The Makefile installed one helper then, `grub-kargs-edit`. It installs fourteen now. The override list has not changed once. So this has failed on every Fedora Atomic host since the repo went public.

```
[...]

PR #429, opened 2026-09-14
fix(daemon): replace shell grants with bounded action helpers

### LINKED ISSUE #417: Drop the sh and runuser -c sudo grants: the file's own TODO has never been filed

QinXi-ai:
Fixes #417.

Remove the unrestricted `/usr/bin/sh` and `/usr/sbin/runuser` grants. The complete call-site inventory is in the issue comment: firewall chaining, Fedora group materialization, snap install-then-hold, SSH key edits, Podman/Toolbox login-shell operations, and the Flatpak consumers of runuser.

These paths now invoke the root-owned `action-steps` helper with one of nine explicit operations. Each sudoers grant names the operation; the helper independently validates the full argument grammar and invokes fixed absolute binaries without a shell. Failed install/firewall/group preparation stops subsequent work. Snap retains its exclusive-resource classification.

For user operations, NSS resolves the target account, then the helper replaces supplementary groups and drops GID/UID before accessing the home or running user-configurable tools. It sets an explicit HOME/XDG/session-bus environment. It never accepts an arbitrary command, environment assignment, or file path. SSH edits compare literal whole lines, retain the inode/mode, and refuse symlinks. Both source-install and Ubuntu provisioning paths install the helper; uninstall removes it.

Compatibility: user-scoped operations now refuse UID 0, SSH edits refuse symlinks, and user startup files are not executed. The new daemon/helper/sudoers must be installed together. These restrictions are a behavior change for the next minor release under the project's 0.y versioning policy. Existing unrelated unprivileged Bash readers (APT history/upgradable packages/reboot sentinel) do not depend on either grant and remain outside this privilege-boundary change. Other powerful administrative grants remain; this PR does not claim sudoers alone is a complete security boundary.

Validation:
- Windows-local direct production-module harness: 45 action-constructor tests and 31 Snap tests PASS. This is not a full daemon build.
- Helper regression: 8 PASS, 3 Linux-only tests SKIPPED locally. The grant-removal regression failed before implementation.
- Linux CI on final head 82bb53f passed all 11 helper tests, including a fixture-only root subprocess that verifies real credential dropping, inability to regain root, user ownership, and refusal to overwrite a root-owned file via symlink. Evidence: https://github.com/lacs-project/sysknife/actions/runs/34867929353 . Actual firewall/Snap/Podman/Toolbox/Flatpak installations remain NOT_TESTED; their argv, sequencing, and drop order have fixture coverage.
- Existing SSH executor tests now exercise the packaged helper's real edit function; action-reference regenerated from production constructors; formatting/diff checks PASS.
- Existing Rust tests were adapted without adding/removing tests. The complete Linux workspace on final head 82bb53f passed 1,852 tests, with 6 skipped; Clippy, rustdoc, frontend and Postgres contract gates also pass. Windows cannot build the existing Unix-only daemon/brain modules.

CI also surfaced the newly published RUSTSEC-2026-0285 in the inherited lockfile. A separate commit updates rustls to 0.23.45 and its required TLS dependencies; the final security-audit check passes.


vladimirrott (dismissed):
Reviewed at `82bb53ff77ad95565aa4597c544d92077c0368c7`.

This removes the two grants that made everything else in `packaging/sysknife-sudoers` advisory:

```
-sysknife ALL=(root) NOPASSWD: /usr/bin/sh
-sysknife ALL=(root) NOPASSWD: /usr/sbin/runuser
```

The file's own header admitted what they meant ("those make the daemon effectively root-capable, so the security boundary is the daemon's own authorization"). The `TODO(post-launch hardening)` above the `sh` grant has been sitting there asking for exactly this refactor. You did it, across nine call sites, without changing what the actions do.

## The guard bites

I proved it rather than reading it. Reverting one production file to `main`'s version, in a network-isolated container:

```
$ git checkout origin/main -- crates/sysknife-daemon/src/actions/ssh.rs
$ cargo test -p sysknife-daemon --test sudoers_grants_match_argv --offline
running 2 tests
test every_sudo_action_is_authorised_by_a_packaged_grant ... FAILED
panicked at crates/sysknife-daemon/tests/sudoers_grants_match_argv.rs:108:13:
AddAuthorizedKey must use bounded argv instead of a privileged shell/user launcher
test result: FAILED. 1 passed; 1 failed
```

Unmutated at your head, the same command reports `running 2 tests ... 2 passed`. So the assertion fails when a shell launcher comes back, and it names the action that brought it. That is the guard I want attached to this change, because the failure mode is someone quietly reintroducing a `sh -c` call site a year from now.

## The helper

The nine sudoers entries end in `*`, and sudo matches command arguments as a single concatenated string, so each grant admits arbitrary trailing text. That makes `action-steps` its own boundary: anything that can run `sudo action-steps ssh-add ...` reaches it directly, skipping the preview, the approval interlock and the signed chain. I read it on that basis rather than as a convenience wrapper, and it holds up.

The credential drop is in the right order, which is the detail that most implementations get wrong:

```python
os.initgroups(account.pw_name, account.pw_gid)
os.setgid(account.pw_gid)
os.setuid(account.pw_uid)
if os.getuid() != account.pw_uid or os.geteuid() != account.pw_uid:
    raise PermissionError("failed to drop root credentials")
```

Supplementary groups first (only possible while root), then GID, then UID, then a verification that it actually took. Reversing the last two would silently leave the process in root's group. Refusing `pw_uid == 0` and requiring an absolute home closes the obvious way to ask it to operate on root's account.

`edit_key` opens with `O_NOFOLLOW | O_NONBLOCK` and then checks `S_ISREG`. The `O_NONBLOCK` is load-bearing and easy to mistake for noise: without it, a FIFO planted at `~/.ssh/authorized_keys` would block the open forever, since `O_NOFOLLOW` only refuses symlinks. Whole-line comparison keeps the fix from #133 intact, so a key like `ssh-ed25519 .*` still cannot act as a pattern.

`#!/usr/bin/python3 -I` is the right shebang. Isolated mode ignores `PYTHONPATH` and user site-packages, so a writable directory in the daemon user's environment cannot inject a module into a root-run helper.

I checked one behavior I expected to be a regression and it is not. The new `ssh-add` path opens `~/.ssh/authorized_keys` with `O_CREAT` but never creates `~/.ssh`. `main`'s `ADD_KEY_SCRIPT` is `... || printf '%s\n' "$key" >> "$path"`, which also fails when the directory is absent. Same behavior on a fresh account, so nothing is lost.

Installation is correct in both paths: `install -Dm 755` root-owned in the Makefile and in `ubuntu-provision.sh`, and `daemon-uninstall` removes it.

## Why I am not merging it today

This touches `.github/workflows/**`. My rule is that a workflow diff gets a human reading it before the merge button, regardless of how small the hunk is, so this one waits for that rather than for anything you need to do. It is in my run report as ready.

Two coordination notes, neither your fault:

#421 edits the same function at the same insertion point, so the two conflict textually. Both additions are compatible and the resolution is to keep both lines. I will sequence them.

Your lockfile commit clears RUSTSEC-2026-0285, which is red on `main` right now. I am landing that through #426 first so the rest of the queue goes green.

## One note for the release

You have this right in the body and I want it recorded in the merge too: user-scoped operations now refuse UID 0, SSH edits refuse symlinks, user startup files no longer execute, and the daemon, helper and sudoers file must be installed together. That is a behavior change, and under 0.y it moves the middle digit. I will write the CHANGELOG entry to say so plainly rather than filing it under a fix.

Approving. This is the most consequential change anyone has contributed to this repo, and the fact that it comes with a guard that fails when the grants come back is why I am comfortable with it.

vladimirrott:
Two merges from me put your three open branches behind, and every conflict below is my merge order rather than anything you wrote. Recording it on one thread instead of three.

`main` now carries #426 as `4b7b5d5` and #421 as `026880a`.

```
$ for n in 413 427 429; do git merge-tree --write-tree main p$n; done
413 CONFLICT: CHANGELOG.md docs/introduction.md
427 CONFLICT: CHANGELOG.md
429 CONFLICT: CHANGELOG.md crates/sysknife-daemon/tests/sudoers_grants_match_argv.rs docs/action-reference.md
```

- `CHANGELOG.md`, all three: #426 opened an `Unreleased` → `Fixed` block where your entries land. Keep both, yours under the same heading.
- `docs/action-reference.md` and `docs/introduction.md`: #426 regenerated the reference from the corrected descriptions, so regenerate on top rather than resolving those hunks by hand.
- `crates/sysknife-daemon/tests/sudoers_grants_match_argv.rs` in #429: #421 added a `checked` counter and a floor assertion at the same insertion point as your bounded-helper assertion. Both are wanted. Keep both lines.

One more on #413: `tests/evidence/workspace-tests.json` needs **1854** now rather than the 1853 it carries. #426 took `main` to 1853 and #413 adds one test above that. `UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh` regenerates the artifact with its measurement fields; `README.md`, `docs/introduction.md` and `docs/distro-support.md` carry the same figure in prose and move in the same commit.

Say the word if you would rather I take the rebases. Otherwise push when it suits you and I will re-review at the new heads. #427 and #429 both touch `.github/workflows/ci.yml`, so neither can go through the merge gate until I have read that diff myself, and nothing is waiting on you tonight.

vladimirrott (approved):
Re-approving at `c38c9ba`. I have merged `main` into this branch twice tonight, and I caused both rounds of conflicts by the order I chose to merge other PRs in. None of it is on you.

This round was one file. #427 landed a `Check local gate discovery and required-check warnings` step at the same anchor in `ci.yml` where you add `Check bounded actions and removal of shell grants`, so git could not pick an order. I kept both, #427's first. Against `main`, the resolved workflow adds your four lines and nothing else:

```diff
       - name: Check local gate discovery and required-check warnings
         run: bash tests/release/ci-local.test.sh
 
+      - name: Check bounded actions and removal of shell grants
+        # Root is used only in an isolated fixture subprocess to prove that the
+        # helper irreversibly drops credentials before touching user files.
+        run: sudo -n bash tests/release/action-steps.test.sh
+
       - name: Check the audit export states its confidentiality class
```

Everything else in the merge is git's own three-way result. I checked that by diffing my merge tree against `git merge-tree --write-tree`, and `.github/workflows/ci.yml` is the only path that differs.

The earlier round was three files: `sudoers_grants_match_argv.rs`, where I kept both the new counter and your shell-launcher assertion; `CHANGELOG.md`; and `docs/action-reference.md`, which is generated, so I regenerated it with `UPDATE_ACTION_REFERENCE=1 cargo test -p sysknife-daemon --test action_reference_doc` instead of hand-merging. The regenerated file carries both sides: 12 `action-steps flatpak` commands and zero `runuser -u testuser`.

One consequence of #427 you could not have seen coming. Its local runner discovers every `tests/release/*.test.sh` and runs it with plain `bash`, so your test now runs for contributors who have no sudo. I ran it as an unprivileged user:

```
test_real_drop_cannot_regain_root_or_write_through_a_symlink ... skipped 'real credential-drop test requires a root test subprocess'
Ran 11 tests in 0.018s
OK (skipped=1)
```

Your `skipUnless(os.geteuid() == 0)` does the right thing there, and the other ten still assert. `scripts/ci-local.sh` stays green without root.

Nothing left for you to do. I will merge once the board is green.

vladimirrott:
Merged. #417 is closed with it.

Nine bounded grants in place of the `sh` and `runuser -c` blanket entries, and the helper drops credentials before it touches a user file rather than after. That was the oldest TODO in `packaging/sysknife-sudoers` and it is gone.

Two things worth saying about the end of this one, neither of them yours. Both rounds of conflicts came from the order I merged other pull requests in, and I resolved them rather than handing them back. And the gate could not prove this one at all when I started: your diff touches `crates/*/src/*.rs`, `packaging/*`, `Makefile` and a workflow, every one of them production, and no single verify suite runs all four. It refused, correctly, and I had to teach it to prove a receipt in parts before it could say anything about your work. It ends up with two:

```
rust   every_sudo_action_is_authorised_by_a_packaged_grant
       mutated the helper path in the daemon -> test FAILED
shell  tests/release/action-steps.test.sh
       loosened /action-steps flatpak * to /action-steps * -> AssertionError
```

Your `skipUnless(os.geteuid() == 0)` also caught a real problem in my tooling rather than in your code. The verify container was running contributor scripts as uid 0, so that case stopped skipping, tried a real `chown` with every capability dropped, and the gate blamed you for it. It runs unprivileged now.

You have eleven merged here. If you want the next one, #301 came back into the pool tonight: `make install` writes into `/usr/lib` on an rpm-ostree host and has never completed there. It is labelled hard and it needs a Fedora Atomic guest, which is the only reason it sat reserved for three weeks. Say the word and I will hold it for you.

Issue #336, opened 2026-09-01
docs/cli.md and the binary disagree on the unattended flag, and omit the variable that controls audit commands


vladimirrott:
Four gaps between `docs/cli.md` and what an operator actually gets, found by building a real nightly job from the docs alone.

## 1. `--help` and `docs/cli.md` state different contracts for `--dangerously-skip-approval`

`--help` says:

> Implies `--yes`, `--max-risk high` and `--non-interactive` unless you set a lower `--max-risk`, which still wins.

`docs/cli.md` says:

> `--yes` may now auto-approve HIGH-risk steps. The cap moves from MEDIUM to HIGH.

The second reads as though `--yes` still has to be typed, and never mentions `--non-interactive` at all. That difference changes what a script must pass, and it changes what `--step-by-step` combined with the flag does. Whichever is right, the other has to move.

## 2. The flag gates and banners subcommands that approve nothing

```
$ SYSKNIFE_I_ACCEPT_UNATTENDED_ROOT=1 sysknife audit export --dangerously-skip-approval
[]
  (full UNATTENDED MODE banner on stderr)

$ sysknife doctor --dangerously-skip-approval      # env var unset
sysknife: --dangerously-skip-approval was passed but SYSKNIFE_I_ACCEPT_UNATTENDED_ROOT is not set to 1.
$ echo $?
1
```

This is deliberate, and the reason is in the source: a flag that is sometimes silently ignored teaches people it is harmless. But `docs/cli.md` frames the whole section around plan execution, and the only hint is one line under Global flags saying all flags apply to every subcommand.

A wrapper that uniformly appends the flag to every invocation, which is a normal "always non-interactive" pattern, gets hard refusals on health checks and audit shipping. Say so in the Unattended mode section.

Stream separation is correct here: the banner is on stderr and stdout stays clean, so `jq` still parses. Only the scope is undocumented.

## 3. `SYSKNIFE_DATABASE_PATH` is missing, and `SYSKNIFE_SOCKET` does nothing for audit commands

```
$ grep -c SYSKNIFE_DATABASE_PATH docs/cli.md
0
```

`developer-guide.md` and `the-audit-chain.md` both document it as the way to point `audit export` and `audit verify` at a specific store, including a copied remote database. `docs/cli.md`'s own Environment variables section never mentions it.

[...]

PR #435, opened 2026-09-15
test: isolate release rehearsal output

### LINKED ISSUE #431: release-rehearsal.test.sh reads its pass condition out of a fixed /tmp path, so a run that could not write reports as a refusal
`tests/release/release-rehearsal.test.sh` proves that the rehearsal script
refuses a publishing mode. It writes the refusal to a fixed path under `/tmp`
and then greps that path for the pass condition:

```
$ git rev-parse --short=8 HEAD
61b3a878
$ sed -n '38,42p' tests/release/release-rehearsal.test.sh
if "$rehearsal" --publish >/tmp/sysknife-rehearsal-publish.out 2>&1; then
    printf 'FAIL: rehearsal accepted a publishing mode\n' >&2
    exit 1
fi
grep -Fq 'never publishes' /tmp/sysknife-rehearsal-publish.out
```

When the redirect fails, bash never runs the command, and the failure is
indistinguishable from the refusal this test is written to observe. The `if`
takes the false branch, which is the passing branch, and line 42 then reads a
file this run did not write.

## Why it matters

Measured, replaying the shape at `:38-42` against a file the process cannot
write:

```
$ f=$(mktemp /tmp/sk-demo-XXXX.out); printf 'rehearsal never publishes\n' > "$f"; chmod 444 "$f"; if echo 'this run really ran' > "$f" 2>&1; then echo "if-branch: TRUE (test would FAIL loudly)"; else echo "if-branch: FALSE (test treats this as the expected refusal)"; fi; grep -Fq 'never publishes' "$f" && echo "line 42 assertion: SATISFIED"; cat "$f"; rm -f "$f"
/bin/bash: line 1: /tmp/sk-demo-wtD2.out: Permission denied
if-branch: FALSE (test treats this as the expected refusal)
line 42 assertion: SATISFIED
rehearsal never publishes
```

`scripts/release_rehearsal.sh` was never invoked, and the contract test reported
that it correctly refuses to publish. That is a guard answering a question it
could not ask, which is the shape this repo files issues about in its own CI.

Two ways to arrive there, neither exotic. Another account on a shared runner or
a shared developer box owns `/tmp/sysknife-rehearsal-publish.out`, because the
name is fixed and predictable. Or `/tmp` is full or read-only, in which case
every gate that writes there fails in a way that reports as a pass here and as
an error elsewhere.

The file is also left behind. The test has no `trap`:

```
$ grep -c trap tests/release/release-rehearsal.test.sh
0
```

This is the only line in `tests/release/` that reads a fixed `/tmp` path back as
a pass condition:

```
$ grep -rn 'grep .*/tmp/' tests/release/
tests/release/release-rehearsal.test.sh:42:grep -Fq 'never publishes' /tmp/sysknife-rehearsal-publish.out
```

Six of the twenty-two scripts in that directory already use `mktemp`, so the
pattern to copy is in the same directory.

## Scope

- Give the test a private temporary directory (`mktemp -d`) and a
  `trap 'rm -rf "$tmp"' EXIT`, and put the publish output inside it.
- Separate "the command ran and refused" from "the command never ran". Capture
  the status explicitly rather than reading it off an `if` whose false branch
  also covers a redirect failure. Assert that the output file is non-empty
  before grepping it.
- Leave the assertion itself alone. `never publishes` is the right string and
  the refusal is the right behaviour; only the plumbing that observes it moves.

Do not fix this by adding `|| exit 1` to the redirect. That closes the one arrival
path you thought of and leaves the shape intact for the next one.

## Tests first

Write the negative case before the fix, because the point of the change is that
a specific wrong state now fails.

1. Pin the fixed path: make the destination unwritable for the running user and
   run the test as it stands. It passes. Paste that.
2. Apply the fix. The same setup now fails, and the message names the write
   rather than the missing string.
3. Then prove the positive case still holds: an ordinary run passes, and the
   temporary directory is gone afterwards.

Step 1 is the one that matters. A version of this change that only adds `mktemp`
and a trap is tidier and proves nothing, because the run that could not write
still reads as a refusal.

## Difficulty

`easy`. One file, about six lines, and the hard part is the test in step 1
rather than the fix. No Rust, no VM, no daemon, no credentials.

## Getting started

armutlutost:
## Summary

Make the release rehearsal contract test use a private temporary directory and explicitly verify the command status and captured output. This prevents an unwritable fixed `/tmp` path from being mistaken for a successful refusal check.

## Related Issue

Closes #431

## Validation

- [x] Tests added or updated
- [ ] CI passes

## Notes for Reviewers

The refusal assertion is unchanged. The test now cleans up its temporary output with a trap and fails clearly when the rehearsal command cannot write its output.


vladimirrott (approved):
Reviewed at `1dea30947c70b5ef6b0069e2d64a3685f1623e6c`. Thank you for this, and welcome.

The `-s` check is the part that earns the change. Moving to `mktemp -d` stops the fixed path being pre-owned, and that alone would have closed #431 on paper. Asserting that the capture is non-empty, and printing the exit status when it is not, is what makes the next failure of this shape loud instead of silent. Capturing `publish_status` and testing it separately rather than leaning on the `if` is the same instinct: the status and the output are two facts, and the old code let one of them stand in for both.

`maintainer screen` returns DO NOT EXECUTE for a diff the gates run directly, so everything below ran in a container with the network off.

Clean, at your head:

```
rc=0
```

Then the scenario #431 describes, against the real `release_rehearsal.sh` rather than a re-enactment of its shape. I staged the fixed path owned by somebody else, mode 444, holding the string the old assertion greps for, and mounted that directory as `/tmp`:

```
$ P=/home/entropia/.local/state/sysknife-maint/poisontmp; mkdir -p "$P"; printf 'rehearsal never publishes\n' > "$P/sysknife-rehearsal-publish.out"; chmod 777 "$P"; chmod 444 "$P/sysknife-rehearsal-publish.out"; ls -l "$P"
total 4
-r--r--r-- 1 entropia entropia 26 sep 15 11:36 sysknife-rehearsal-publish.out
```

`main`'s copy of the test, run as an unprivileged uid against that `/tmp`:

```
$ podman run --rm --network=none --user 1000:1000 -v "$PWD:/repo:ro" -v /home/entropia/.local/state/sysknife-maint/poisontmp:/tmp:z -v "$HOME/.cargo:/cargo:O" -v "$HOME/.rustup:/rustup:O" -w /repo -e HOME=/repo -e CARGO_HOME=/cargo -e RUSTUP_HOME=/rustup -e CARGO_NET_OFFLINE=true -e PATH=/cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin localhost/sk-reh2:1 bash -c 'id -u; ls -l /tmp/sysknife-rehearsal-publish.out; bash tests/release/release-rehearsal.test.sh; echo "rc=$?"'
1000
-r--r--r-- 1 root root 26 Sep 15 17:36 /tmp/sysknife-rehearsal-publish.out
tests/release/release-rehearsal.test.sh: line 38: /tmp/sysknife-rehearsal-publish.out: Permission denied
rc=0
```

`release_rehearsal.sh --publish` was never invoked, and the contract test reported that it correctly refuses to publish. Your branch, same container, same poisoned `/tmp`:

```
rc=0
```

which is a pass the run earned, because `mktemp -d` never touches the name somebody else owns.

Then the part I care about most, whether the new assertion can fail. I dropped the `2>&1` from your redirect, so the command still runs and the capture comes back empty:

```
$ git diff --stat
 tests/release/release-rehearsal.test.sh | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
ERROR: release rehearsal never publishes; use the protected tag workflow
FAIL: rehearsal output was not captured (exit 2)
rc=1
```

It names the failure and carries the exit status. `shellcheck --severity=warning tests/release/release-rehearsal.test.sh` exits 0.

Approving.

## Your checks are waiting on me, not on you

```
$ gh pr checks 435 --repo lacs-project/sysknife
no checks reported on the 'fix/431-private-rehearsal-output' branch
```

GitHub holds a first-time contributor's workflow runs until somebody with write access presses "Approve and run". That is a control I keep rather than automate, so the delay is mine. The same applies to #436. Nothing is wrong with your branch.

## One line will need to move, and not because of anything you did

@vsolano9's #432 is approved and touches the same file, adding its own `trap 'rm -rf "$pin_fixture"' EXIT` at what becomes line 161. Yours is at 39. The two merge with no conflict:

```
$ git merge-tree --write-tree p432 p435
2d3ab157b24e79bf0710a86662c84677394ad431
$ git show 2d3ab157:tests/release/release-rehearsal.test.sh | grep -n 'trap\|mktemp'
38:tmp_dir="$(mktemp -d)"
39:trap 'rm -rf "$tmp_dir"' EXIT
160:pin_fixture="$(mktemp -d)"
161:trap 'rm -rf "$pin_fixture"' EXIT
```

Bash keeps one EXIT trap per signal. The second replaces the first, so the merged file stops cleaning up your directory and both boards stay green:

```
$ bash /tmp/trapdemo.sh > /tmp/trapout.txt; cat /tmp/trapout.txt; for d in $(sed 's/^[ab]=//' /tmp/trapout.txt); do [ -d "$d" ] && echo "SURVIVED: $d" || echo "removed:  $d"; done
a=/tmp/tmp.24yM4OPfvx
b=/tmp/tmp.vqjaMeTF35
SURVIVED: /tmp/tmp.24yM4OPfvx
removed:  /tmp/tmp.vqjaMeTF35
```

Whichever of the two lands second folds both paths into one cleanup. I am sequencing that, and I will tell you if it falls to you. Change nothing now.

## Next

Once the run is approved and green, this merges as it stands.

You have two open at once already, so I am not going to hand you a third. When these land, tell me whether you would rather keep going on documentation that contradicts the binary or move to the release gates, and I will hold something in whichever direction you pick.

Last thing, an invitation rather than a condition. Somebody who reads a `/tmp` path and sees an ownership problem is administering machines where that reasoning pays. SysKnife's read-only surface is the cheap way to try it: `sysknife doctor`, or `sysknife "show me failed services"`, both of which plan and preview before touching anything. If you run it on a box you own, I would rather hear what it got wrong than not hear.

vladimirrott:
Your checks are running. They had been sitting at `action_required` since you opened this, which is the hold GitHub puts on a first contribution until someone with write access releases it, and releasing it was mine to do rather than yours.

I also brought the branch up to date with `main`, so the head is now `105787ee` and the run is against a tree that matches what a merge would produce. Nothing in your diff changed; the only file this PR touches is `tests/release/release-rehearsal.test.sh`.

My approval from earlier still stands. Nothing outstanding from you on this one.

vladimirrott:
Your checks are green at `105787ee`, and the substance of this holds up: you reproduced #431 against the real `scripts/release_rehearsal.sh` instead of re-enacting it, and the new assertion fires when the capture comes back empty rather than passing on a file nobody wrote.

One mechanical item is left, and I caused it. #432 landed on `main` yesterday and I brought your branch up to date, which put a second `trap ... EXIT` into the file you are editing:

```
$ git show p435:tests/release/release-rehearsal.test.sh | grep -n 'trap\|mktemp'
38:tmp_dir="$(mktemp -d)"
39:trap 'rm -rf "$tmp_dir"' EXIT
160:pin_fixture="$(mktemp -d)"
161:trap 'rm -rf "$pin_fixture"' EXIT
```

Bash keeps one handler per signal, so line 161 replaces line 39 and the directory your fix creates outlives the run. I ran the test three times in a container with a `TMPDIR` I could read afterwards.

Your head, `105787ee`:

```
$ podman run --rm --network=none -v /tmp/wt435:/repo:z -v /tmp/tmp435:/sktmp:z -w /repo -e TMPDIR=/sktmp -e CARGO_NET_OFFLINE=true localhost/sk-reh3:1 bash -c 'bash tests/release/release-rehearsal.test.sh; echo "TEST rc=$?"; echo "--- what is left in TMPDIR ---"; ls -1 /sktmp'
TEST rc=0
--- what is left in TMPDIR ---
tmp.g1XXeByoX3
```

The survivor is `$tmp_dir`, not #432's fixture:

```
$ ls -la /tmp/tmp435/tmp.g1XXeByoX3
total 12
drwx------ 2 entropia entropia 4096 sep 16 11:23 .
drwxrwxr-x 3 entropia entropia 4096 sep 16 11:23 ..
-rw-r--r-- 1 entropia entropia   73 sep 16 11:23 publish.out
$ head -1 /tmp/tmp435/tmp.g1XXeByoX3/publish.out
ERROR: release rehearsal never publishes; use the protected tag workflow
```

`main` on its own cleans up, so I introduced the leak with that update and your commits did not:

```
$ podman run --rm --network=none -v /tmp/wtmain:/repo:z -v /tmp/tmpmain:/sktmp:z -w /repo -e TMPDIR=/sktmp -e CARGO_NET_OFFLINE=true localhost/sk-reh3:1 bash -c 'bash tests/release/release-rehearsal.test.sh; echo "TEST rc=$?"; echo "--- left in TMPDIR ---"; ls -1A /sktmp; echo "(end)"'
TEST rc=0
--- left in TMPDIR ---
(end)
```

Folding the two into one handler clears it. This is the change I ran, on a copy of your head:

```diff
 pin_fixture="$(mktemp -d)"
-trap 'rm -rf "$pin_fixture"' EXIT
+trap 'rm -rf "$tmp_dir" "$pin_fixture"' EXIT
```

```
$ podman run --rm --network=none -v /tmp/wt435fix:/repo:z -v /tmp/tmp435fix:/sktmp:z -w /repo -e TMPDIR=/sktmp -e CARGO_NET_OFFLINE=true localhost/sk-reh3:1 bash -c 'bash tests/release/release-rehearsal.test.sh; echo "TEST rc=$?"; echo "--- left in TMPDIR ---"; ls -1A /sktmp; echo "(end)"'
TEST rc=0
--- left in TMPDIR ---
(end)
```

All three runs print `Release rehearsal contract passed.` and exit 0, so nothing in CI catches this. Take the two-word edit however you prefer to write it; my approval stands and I am asking for nothing else on this branch.

You have #336 open as well, so I am not putting a third issue in front of you. When both land I will find you something in the same area.

Few people go after a temp-file ownership bug with a reproduction of the original failure. If you run Linux anywhere that matters to you, a server you patch or a machine you keep alive, I would like to know how SysKnife behaves there. `npx sysknife-setup` is the whole setup, and the read-only side (`sysknife doctor`, `sysknife history`, `sysknife audit verify`) changes nothing on the host.

PR #452, opened 2026-09-17
fix(docs): validate ActionSpec claim counts

### LINKED ISSUE #450: The claim checker exempts any line naming ActionSpec, and the developer guide's 189 has drifted to 191
`docs/developer-guide.md:349` says the daemon carries 189 actions with an `ActionSpec`. The tree has 191, and `docs/action-reference.md:401` says so on the same subject.

```
$ grep -rn 'actions \(with\|have\) an `ActionSpec`' docs/
docs/developer-guide.md:349:  sysknife-daemon/    Privileged executor, 189 actions with an `ActionSpec`,
docs/action-reference.md:401:_191 actions have an `ActionSpec` and are tabled above. The full catalogue (`KNOWN_ACTION_NAMES`) also includes `ListJobHistory`, which the dispatcher handles before the executor, for **192** total._
```

Both files are in the claim checker's corpus and it passes anyway:

```
$ out="$(python3 scripts/check_evidence_claims.py . 2>&1)"; rc=$?; printf '%s\n' "$out" | tail -4; echo "rc=$rc"
rc=0
```

## Why it matters

`check_action_figures` in `scripts/check_evidence_claims.py` catches a bare "N actions" that is not the catalogue size, and skips any line mentioning `ActionSpec` so an honest sub-count stays writable:

```python
    for rel, text in texts.items():
        for line in text.splitlines():
            if "ActionSpec" in line:
                continue
```

The skip has no upper bound. Naming `ActionSpec` on a line exempts every number on it from every check, so the sub-count is the one figure in the published documentation that nothing verifies. Measured at `30b9e315`, on an extracted copy of `HEAD` with the number changed to `4`:

```
$ grep -n 'actions with an `ActionSpec`' /tmp/skc449/docs/developer-guide.md
349:  sysknife-daemon/    Privileged executor, 4 actions with an `ActionSpec`,
$ cd /tmp/skc449 && out="$(python3 scripts/check_evidence_claims.py . 2>&1)"; rc=$?; printf '%s\n' "$out" | tail -4; echo "checker rc=$rc"
checker rc=0
```

The docstring on that function tells the story it is now repeating. It was written because "the catalogue defines 189 actions" sailed past `check_figure` in three documents while the catalogue held 190. The fix taught the checker to demand a qualifier, and then trusted the qualifier.

## Scope

Two changes, and the number is the smaller one.

**Correct `docs/developer-guide.md:349` to 191.** Do not correct it to 192: that line names the executor, and `ListJobHistory` is handled by the dispatcher before the executor sees it, which is the distinction `docs/action-reference.md:401` spells out.

**Bound the escape hatch.** A line naming `ActionSpec` should be required to state the `ActionSpec` count rather than exempted from stating anything. The count is derivable without new machinery: `crates/sysknife-daemon/tests/action_reference_doc.rs` generates the `docs/action-reference.md` table from the catalogue and fails on drift, so the tabled rows are pinned to the code and counting them gives 191:

```
$ grep -cE '^\| `[A-Za-z0-9_]+` \|' docs/action-reference.md
191
```

The trap is hardcoding 191 in the Python. That trades a stale number in a document for a stale number in the thing that checks documents, which is worse, because the second one looks like verification.

## Tests first

**Break what the guard protects.** Write a fixture line reading `N actions with an ActionSpec` where `N` is neither 191 nor 192, run the checker over it, and assert it reports the file, the published figure and the derived one. That test fails today, which is the defect. `4` is a good value to pick because it is far enough from both real counts that a fuzzy comparison cannot pass it.

**Break the guard's own input.** This check has a specific way of becoming useless: the row regex stops matching `docs/action-reference.md`, the derived count comes back 0, and every document agrees with nothing. Assert the derived count is non-zero before comparing, and assert that a fixture with three tabled rows derives exactly 3. A checker that reports success over an empty read is the failure this repository keeps finding, most recently in #407.

**Keep the honest case honest.** `docs/action-reference.md:401` states 191 and 192 on one line and must stay green. Assert that line passes unchanged, so the bound you add does not turn the one correct sub-count in the tree into a failure.

## Worth knowing before you start

#430 is open against `docs/developer-guide.md` as well, on the pre-commit hook prose around lines 279 and 421. Different sections, so the two merge cleanly, but say in your pull request which lines you touched.

## Difficulty

`easy`. One number in Markdown, and roughly ten lines of Python for the bound plus its fixtures. No Rust, no VM, no provider, no credentials: `python3 scripts/check_evidence_claims.py .` runs from a clean checkout in under a second.

## Getting started

has the build and test commands. The two that matter here are
`python3 scripts/check_evidence_claims.py .` and
`bash tests/release/public-claims.test.sh`, which is where the checker's own
fixtures live. No CLA and no copyright waiver. The project is MIT.

TayfurYldz:
Closes #450

## Problem
Lines mentioning `ActionSpec` were exempted from action-count validation entirely, so stale subset counts could pass the public-claims checker. The developer guide had consequently drifted from 189 to the generated table's 191 entries.

## Fix
- derive the `ActionSpec` count from generated rows in `docs/action-reference.md`
- fail closed if that generated table yields no rows
- validate qualified `ActionSpec` counts instead of skipping them
- correct the developer-guide executor count to 191
- add regression coverage for a stale count, exact row derivation, and the empty-table failure

## Validation
- `bash tests/release/public-claims.test.sh`
- `git diff --check`
- attempted `cargo nextest run --workspace --locked`; it could not complete on this WSL host because the system `glib-2.0` development package is unavailable to `pkg-config` (`glib-2.0.pc` missing). No Rust files are changed by this PR.


vladimirrott (approved):
Reviewed at `c79eaec1`.

The design choice here is the one that makes this worth more than a number bump. You could have hardcoded 191 and moved on. Instead `count_action_specs` reads `docs/action-reference.md`, which `crates/sysknife-daemon/tests/action_reference_doc.rs` generates from `catalogue()` and drift-guards against the committed file. So the published figure now traces back to the daemon's live `ActionSpec` list through two gates rather than through somebody's memory.

I recounted both numbers independently of your code:

```sh
python3 - <<'PY'
import re,pathlib
t=pathlib.Path('crates/sysknife-brain/src/planning_tools/propose_plan.rs').read_text()
marker="pub const KNOWN_ACTIONS: &[(&str, &str)] = &["
table=t.split(marker,1)[1].split("\n];",1)[0]
known={m for m in re.findall(r'^\s*\("([A-Za-z0-9]+)",', table, re.M)}
ref=pathlib.Path('docs/action-reference.md').read_text().splitlines()
rows={re.match(r"^\| `([A-Za-z0-9_]+)` \|", l).group(1) for l in ref if re.match(r"^\| `[A-Za-z0-9_]+` \|", l)}
print("|KNOWN_ACTIONS| =",len(known),"  |doc rows| =",len(rows))
print("in KNOWN_ACTIONS, no doc row:", sorted(known-rows))
print("doc row, not in KNOWN_ACTIONS:", sorted(rows-known))
PY
```

```text
|KNOWN_ACTIONS| = 192   |doc rows| = 191
in KNOWN_ACTIONS, no doc row: ['ListJobHistory']
doc row, not in KNOWN_ACTIONS: []
```

191 is right, and the gap is `ListJobHistory`, which `actions/mod.rs:252` documents as dispatcher-internal with no spec. Your regex matched all 191 rows and missed none: 191 lines start with ``| ` `` and 191 match.

## Mutations

Three, each in a container with no network:

**The production hunk reverted.** I replaced `scripts/check_evidence_claims.py` with `main`'s copy and left your test in place:

```sh
git show main:scripts/check_evidence_claims.py > /tmp/m452a/scripts/check_evidence_claims.py
```

```text
FAIL: checker accepted stale claim: ActionSpec count that disagrees with the generated table
TEST rc=1
```

**Fail-closed removed** (`if count == 0: raise Failure(...)` deleted):

```text
empty ActionSpec table did not fail closed
TEST rc=1
```

**The shipped doc put back to 189**, which is the live path and the one I care about most:

```text
Claims that do not derive from evidence:
  - docs/developer-guide.md: claims 189 actions with an ActionSpec, derived 191 from docs/action-reference.md
Invalid public claim: a published figure does not derive from evidence
TEST rc=1
```

Unmutated at your head: `Public claims contract passed.`, rc=0. Shellcheck over the four script trees returns 0.

One thing I checked and want to record, because it looks like a hole and is not one. Your change drops the old blanket `if "ActionSpec" in line: continue`, so a line naming `ActionSpec` in a phrasing your `subset` regex does not read now falls through to the bare-count check and fails. That is the safe direction. Both lines in the tree that carry an `ActionSpec` count use the two spellings you cover, `with an` and `have an`, and the second is `docs/action-reference.md` line 401, which the Rust generator writes. A future rewording produces a loud false failure rather than a silent pass, which is the trade I want.

## Your checks have not run, and that is mine to fix

`gh pr checks 452` reports nothing, because all four workflow runs sit at `action_required` waiting for a maintainer click. Nothing on your side causes that and nothing on your side fixes it. It is the first-contribution gate, it will not happen again after your first merge, and I am sorry you are waiting on it.

So I ran the gate your PR touches myself rather than leaving you with an empty board. `bash tests/release/public-claims.test.sh` passes at `c79eaec1`, and the three mutations above are why I believe it means something. Your note about `cargo nextest` failing on WSL for a missing `glib-2.0.pc` is a GUI-crate dependency and has nothing to do with this change; no Rust file moves here. Not your debt.

Approved. Merging once the runs are released.

## Next

Checking your derivation turned up something one layer down, and I am holding it for you rather than putting it in front of anyone else.

`crates/sysknife-daemon/tests/action_reference_doc.rs` writes the catalogue total at the foot of `docs/action-reference.md` as `total + 1`, and names `ListJobHistory` in that sentence as a literal. The list that decides how many actions bypass the executor is `DISPATCHER_INTERNAL_ACTIONS` in `action_consistency.rs`, and the generator never reads it:

```sh
grep -rn "DISPATCHER_INTERNAL_ACTIONS" crates/ --include=*.rs
```

```text
crates/sysknife-daemon/tests/action_consistency.rs:27:const DISPATCHER_INTERNAL_ACTIONS: &[&str] = &["ListJobHistory"];
crates/sysknife-daemon/tests/action_consistency.rs:33:    for &name in DISPATCHER_INTERNAL_ACTIONS {
crates/sysknife-daemon/tests/action_consistency.rs:67:    let dispatcher_internal: BTreeSet<&str> = DISPATCHER_INTERNAL_ACTIONS.iter().copied().collect();
```

Add a second dispatcher-internal action and `action_consistency.rs` forces you to update its list, while the generated doc keeps saying `192 total` over a catalogue of 193. `action_reference_doc_is_current` compares the committed file against the generator, both sides carry the same arithmetic, and the test stays green. That is the figure your checker now reads as evidence.

I am writing this up as an issue with the reproduction spelled out, and it is yours on no deadline when it lands. Same shape as what you did here, one layer further down.

Separately: you read a claim-checking script closely enough to notice that a whole class of line was being skipped. People who read that carefully tend to run real machines. If you do, I would like to know how SysKnife holds up on them. `npx sysknife-setup` gets you going, and `sysknife doctor`, `sysknife history` and `sysknife audit verify` touch nothing on the host. No pressure either way.

vladimirrott:
Landed as 736f9451, closing #450.

`check_evidence_claims.py` had `if "ActionSpec" in line: continue`, so a whole class of published figure went unchecked and `docs/developer-guide.md` had drifted to 189 against a real 191. You derived the number from `docs/action-reference.md` rather than hardcoding it, and that table is generated from `catalogue()` by `crates/sysknife-daemon/tests/action_reference_doc.rs` and drift-guarded there, so the claim now traces to the executor instead of to a second document.

The merge gate watched your guard bite at the exact head it merged:

```
$ maintainer-merge verify 452 c1763ab4b46cec8e529ae75f1754aecab9f6c2b8 'tests/release/public-claims.test.sh' 's|if count != action_specs:|if False:|' shell
  running 'tests/release/public-claims.test.sh' unmutated
  receipt recorded for #452 at c1763ab4 (observed: clean pass, mutated fail)
```

Eleven checks passed, none failing, none pending.

Your four workflow runs sat at `action_required` for a day before that. First contributions here need a maintainer to release them by hand, and I had not. The wait was mine.

The follow-up I named in the review is still held for you, and I have not filed it yet.

vladimirrott:
Filed the follow-up as #455, and I have offered it to you there.

Issue #451, opened 2026-09-17
Nothing says whether CONTRIBUTING.md's fallback or CLAUDE.md's pre-commit gate governs a contributor who cannot build the GUI crate


vladimirrott:
`CLAUDE.md:23-29` requires a full workspace test run before any commit. `CONTRIBUTING.md:149-156` describes a fallback for contributors who cannot build the GUI workspace members. Neither file says which governs a contributor whose machine cannot run the first one, and a contributor stopped work for two days to ask.

```
$ sed -n '23,29p' CLAUDE.md
## Pre-Commit Gate

**`cargo nextest run --workspace --locked` must pass with zero failures before
any commit.** Fix every failing test — do not commit with known
pre-existing failures. If a test asserts the wrong value, fix the test
assertion to match the code (or fix the code if the test is right); do
not skip or ignore the test.
```

```
$ sed -n '149,156p' CONTRIBUTING.md
The GUI workspace members still need the platform libraries named in CI. If
you cannot install those, do not estimate a count or edit the evidence metadata
by hand. Leave `tests/evidence/workspace-tests.json` and the three prose files
untouched. In the PR body, name the command you could not run and the missing
tool or system library that blocked it. A maintainer will run the full suite and
regenerate all four files before merge. This is the supported fallback; an
honest missing measurement is better than metadata that describes a run which
never happened.
```

The `CONTRIBUTING.md` passage grants its fallback for one artifact, the test-count metadata. It says nothing about the commit itself, which is what `CLAUDE.md` forbids.

## Why it matters

The wall is the workspace layout, not any contributor's setup. `apps/sysknife-shell/src-tauri` is a workspace member, so `--workspace` compiles the Tauri crate and needs WebKit and GTK even when the change touches neither:

```
$ cargo metadata --no-deps --format-version 1 | python3 -c "import sys,json; d=json.load(sys.stdin); print([p['name'] for p in d['packages']])"
['sysknife-core', 'sysknife-brain', 'sysknife-types', 'sysknife-proto', 'sysknife-daemon', 'sysknife-daemon-test', 'sysknife-shell', 'sysknife-cli']
```

`CONTRIBUTING.md:38` says GUI work is paused, so the crate that blocks the gate is the crate nobody is meant to be working on.

This has already cost time. On #338 a contributor named both documents, asked which applied, and waited rather than opening a draft:

[...]

PR #453, opened 2026-09-17
test(release): remove the dead INSTALL_DIRS derivation and pin the survivor

### LINKED ISSUE #449: install-paths.test.sh computes INSTALL_DIRS twice and throws the first answer away
`tests/release/install-paths.test.sh:54` computes `declared` with a `sed` that reads one physical line, and line 56 overwrites it unconditionally with the `awk` that reads the whole logical one. Nothing reads the first value.

```
$ sed -n '54,58p' tests/release/install-paths.test.sh
declared="$(sed -nE 's/^INSTALL_DIRS[[:space:]]*=[[:space:]]*(.*)$/\1/p' "$makefile")"
# The assignment is line-continued, so pull the whole logical line.
declared="$(awk '/^INSTALL_DIRS[[:space:]]*=/{ found=1 } found { print; if ($0 !~ /\\$/) exit }' "$makefile" \
    | grep -oE '\$\([A-Z_]+\)' | tr -d '$()' | sort -u)"
[ -n "$declared" ] || { printf 'Makefile declares no INSTALL_DIRS\n' >&2; exit 1; }
```

`CLAUDE.md` is explicit that a superseded workaround gets removed rather than left in place, and ShellCheck at `--severity=warning`, which is what `e2e / scripts-lint` runs, does not report an unused assignment.

## Why it matters

The dead line is not a harmless duplicate. It produces a different and wrong answer, and it looks authoritative sitting above the one that works:

```
$ sed -nE 's/^INSTALL_DIRS[[:space:]]*=[[:space:]]*(.*)$/\1/p' Makefile
$(BINDIR) $(SYSUSERS) $(TMPFILES) $(SYSTEMD) $(POLKIT) \
$ awk '/^INSTALL_DIRS[[:space:]]*=/{ found=1 } found { print; if ($0 !~ /\\$/) exit }' Makefile
INSTALL_DIRS = $(BINDIR) $(SYSUSERS) $(TMPFILES) $(SYSTEMD) $(POLKIT) \
               $(SUDOERS) $(HELPERS)
```

The `sed` stops at the backslash and loses `$(SUDOERS)` and `$(HELPERS)`. `$(HELPERS)` is the variable this whole file exists for: the header says `make install` on Fedora Atomic wrote the binary, the unit, the polkit rules and twelve sudo grants, then hit `$(HELPERS)` at `/usr/lib/sysknife` and stopped, leaving live grants naming helper scripts that were never written. That is #301.

So the next person who reads lines 54 to 57, decides the two look redundant and deletes the `awk`, turns the check that guards #301 into a check that cannot see the path #301 is about. The comment on line 55 warns about exactly that and sits on the wrong side of the line it explains.

## Scope

Delete line 54. Move the line 55 comment above the `awk` so it explains the code it belongs to.

Do not touch the `sed` on line 48 that extracts `PREFIX`, which reads a genuinely single-line assignment and is correct.

## Tests first

Deleting a line is not the deliverable. The deliverable is the assertion that says why the surviving line is the one that survived.

**Prove the `awk` is load-bearing.** I measured this before filing, so you have a target to reproduce. Clean run at `30b9e315`:

```
$ out="$(bash tests/release/install-paths.test.sh 2>&1)"; rc=$?; printf '%s\n' "$out"; echo "rc=$rc"
install-paths: KNOWN GAP HELPERS defaults to /usr/lib/sysknife and is not redirected on rpm-ostree (#301)
install-paths: INSTALL_DIRS covers all 7 recipe destinations; 7 defaults checked against the ostree overrides
rc=0
```

Copy the `Makefile`, `tests/e2e/provision.sh` and this test into a scratch tree, replace the `awk` with the line-54 `sed`, and it goes red naming both lost variables:

```
$ sed -n '54,58p' /tmp/skip449/tests/release/install-paths.test.sh
declared="$(sed -nE 's/^INSTALL_DIRS[[:space:]]*=[[:space:]]*(.*)$/\1/p' "$makefile")"
# The assignment is line-continued, so pull the whole logical line.
declared="$(sed -nE 's/^INSTALL_DIRS[[:space:]]*=[[:space:]]*(.*)$/\1/p' "$makefile" \
    | grep -oE '\$\([A-Z_]+\)' | tr -d '$()' | sort -u)"
[ -n "$declared" ] || { printf 'Makefile declares no INSTALL_DIRS\n' >&2; exit 1; }
$ out="$(bash /tmp/skip449/tests/release/install-paths.test.sh 2>&1)"; rc=$?; printf '%s\n' "$out"; echo "rc=$rc"
install-paths: daemon-install writes into $(HELPERS) but INSTALL_DIRS omits it, so the preflight never checks it
install-paths: daemon-install writes into $(SUDOERS) but INSTALL_DIRS omits it, so the preflight never checks it

install-paths: 2 failure(s)
rc=1
```

That is the assertion to add: a fixture whose `INSTALL_DIRS` is line-continued has to be read whole, and a reader that stops at the backslash has to fail naming what it lost.

**Prove nothing else changed.** The clean run above has to reproduce byte for byte after the deletion. A one-line deletion that moves the summary line is not the deletion you meant.

If you want to go one step further, the same shape may exist elsewhere. `grep -n '^\s*\([a-z_]*\)=.*$' ` over `tests/release/` and `scripts/` for a variable assigned twice with no read between is a small enough search to do by eye, and anything it finds belongs in the same pull request.

## Difficulty

`easy`. One deleted line and one moved comment. The mutation that proves the remaining line matters is the part worth your time, and it needs no VM, no daemon, no provider and no credentials: the file reads the `Makefile` and one shell script and runs nothing.

## Getting started

has the build and test commands. This one runs on its own:
`bash tests/release/install-paths.test.sh`. No CLA and no copyright waiver. The
project is MIT.

kragent66-glitch:
## Summary

`tests/release/install-paths.test.sh` derived `INSTALL_DIRS` twice, three lines apart, and the first derivation was both unused and wrong.

Line 54 read one physical line with `sed`, which stops at the backslash and loses `$(SUDOERS)` and `$(HELPERS)`. Line 56 then overwrote it unconditionally with the `awk` that reads the whole logical line. Nothing read the first value — and ShellCheck at `--severity=warning`, which is what `e2e / scripts-lint` runs, does not report an unused assignment — so it sat there looking authoritative above the reader that works. The next person to read those four lines sees two spellings of one idea and a coin flip over which to delete; pick wrong and the check stops covering `$(HELPERS)`, which is the path #301 is about.

This removes the dead derivation and adds the assertion that says why the survivor is the one that survived.

Two changes beyond the deletion, both in service of that assertion:

- The derivation moves into `install_dirs_vars()`, so the fixture exercises the real reader instead of a copy of it. Extracting it is what makes the new assertion mean anything.
- The comment that sat between the two derivations now sits above the `awk` it describes.

The `sed` at line 48, which extracts `PREFIX` from a genuinely single-line assignment, is untouched.

## Related Issue

Closes #449

## Validation

- [x] Tests added or updated
- [ ] Documentation updated if behavior changed — nothing to document: one dead line removed, the surviving derivation extracted, one assertion added
- [x] Security impact considered — the assertion guards the coverage #301 depends on (`$(HELPERS)`), so deleting the wrong reader can no longer silently disable that check
- [x] Trust boundary preserved (daemon remains the only privileged executor) — no daemon, IPC, or polkit surface touched
- [ ] CI passes — not yet run

**Clean run, byte-identical at `30b9e31`** (the commit the issue measured against):

```
$ bash tests/release/install-paths.test.sh
install-paths: KNOWN GAP HELPERS defaults to /usr/lib/sysknife and is not redirected on rpm-ostree (#301)
install-paths: INSTALL_DIRS covers all 7 recipe destinations; 7 defaults checked against the ostree overrides
rc=0
```

**Mutation, using the method from the issue** — copy the `Makefile`, `tests/e2e/provision.sh` and this test into a scratch tree, then swap the surviving `awk` for the line-54 `sed`:

```
$ bash /tmp/skip449/tests/release/install-paths.test.sh
install-paths: a line-continued INSTALL_DIRS lost $SUDOERS; the reader stopped at the backslash
install-paths: a line-continued INSTALL_DIRS lost $HELPERS; the reader stopped at the backslash
install-paths: daemon-install writes into $(HELPERS) but INSTALL_DIRS omits it, so the preflight never checks it
install-paths: daemon-install writes into $(SUDOERS) but INSTALL_DIRS omits it, so the preflight never checks it

install-paths: 4 failure(s)
rc=1
```

The two `daemon-install writes into …` lines are the failure the issue predicted from the clean run. The two `lost $…` lines are the new assertion naming the variables the bad reader dropped, which is what makes the mutation legible without diffing the two readers by eye.

**ShellCheck**, the same invocation CI uses:

```
$ shellcheck --severity=warning tests/release/install-paths.test.sh
(no output)
```

## Notes for Reviewers

- **`cargo nextest run --workspace --locked`: NOT RUN.** This box has no `glib-2.0` development headers, so the workspace does not compile here — `glib-sys` fails at its `pkg-config` probe with exit code 101. That is pre-existing and unrelated to this diff, which touches no Rust code. I did run the build step that `e2e / scripts-lint` actually performs, `cargo build -p sysknife-cli --locked`, and it passes here (`Finished dev profile in 1m 38s`).
- The assertion is deletion-safe by design: it fails if the surviving reader is replaced by one that stops at the backslash, and passes when the survivor is kept. It pins the property rather than the file's shape.
- **AI assistance:** drafted with Hermes (an AI agent) against the no-dead-code rule in `CLAUDE.md` and the AI section of `CONTRIBUTING.md`. Every command output quoted above comes from a run reproduced on `30b9e31`, not from the draft.


vladimirrott (approved):
Reviewed at `aa12d940`.

The PR body makes a claim I wanted to test rather than accept: that extracting `install_dirs_vars()` is what makes the new assertion mean anything, and that check 1 alone would not catch the regression. Both hold.

First, the guard bites. I replaced the surviving `awk` with the `sed` one-liner you deleted, and separately made the `awk` stop at the backslash:

```text
=== M1-sed-one-liner-restored ===
install-paths: a line-continued INSTALL_DIRS lost $SUDOERS; the reader stopped at the backslash
install-paths: a line-continued INSTALL_DIRS lost $HELPERS; the reader stopped at the backslash
install-paths: daemon-install writes into $(HELPERS) but INSTALL_DIRS omits it, so the preflight never checks it
install-paths: daemon-install writes into $(SUDOERS) but INSTALL_DIRS omits it, so the preflight never checks it

install-paths: 4 failure(s)
TEST rc=1

=== M2-stop-at-backslash ===
(identical four lines)
TEST rc=1
```

Then the part that decides whether your fixture earns its place. Today's `Makefile` happens to be line-continued, so check 1 catches the regression on its own and your fixture looks redundant. I reflowed `INSTALL_DIRS` onto one physical line and re-ran with the `sed` reader restored:

```sh
python3 - <<'PY'
import pathlib,re
p=pathlib.Path('/tmp/m453c/Makefile'); s=p.read_text()
lines=s.splitlines(True); out=[]; i=0
while i < len(lines):
    if re.match(r'^INSTALL_DIRS\s*=', lines[i]):
        buf=[]
        while True:
            buf.append(lines[i].rstrip('\n')); cont=lines[i].rstrip('\n').endswith('\\'); i+=1
            if not cont: break
        joined=' '.join(x.rstrip('\\').strip() for x in buf)
        out.append(joined+'\n'); print("reflowed to one line:", joined)
    else:
        out.append(lines[i]); i+=1
p.write_text(''.join(out))
PY
```

```text
install-paths: a line-continued INSTALL_DIRS lost $SUDOERS; the reader stopped at the backslash
install-paths: a line-continued INSTALL_DIRS lost $HELPERS; the reader stopped at the backslash
install-paths: KNOWN GAP HELPERS defaults to /usr/lib/sysknife and is not redirected on rpm-ostree (#301)

install-paths: 2 failure(s)
TEST rc=1
```

Check 1 goes quiet. Two failures, not four, and the only thing still catching the broken reader is your fixture. That is the assertion doing work no other line in the file does, and it is why the extraction was the right call rather than churn.

Unmutated at your head the script prints `INSTALL_DIRS covers all 7 recipe destinations; 7 defaults checked against the ostree overrides`, rc=0. Shellcheck at `--severity=warning` over the four script trees returns 0. Your byte-identical claim against `30b9e315` matches what I see.

Two details I checked because they are the usual way a shell fixture goes wrong: the file had no prior `EXIT` trap, so yours clobbers nothing, and `note()` increments `fail_count`, which the tail turns into `exit 1`. The assertion is not decorative.

One optional note, take it or leave it. `printf '%s\n' "$fixture_vars" | grep -qx "$expected"` is the shape that has bitten this repo before: under `pipefail`, `grep -q` exits on the first match, the writer can take SIGPIPE, and the pipeline reports a failure that did not happen. Your payload is five lines so the race is not reachable in practice, and I am not asking you to change it. `grep -qx "$expected" <<<"$fixture_vars"` is the house idiom if you touch the file again.

Approved, and nothing blocks it.

On #449 being contended: @TayfurYldz posted a scope there thirty-one seconds after I offered it to you, I held it for you, and they took #450 instead. You landing this closes it correctly. Nothing for you to do about it; I am naming it so the thread reads straight later.

## Next

You have now removed a dead reader and pinned the survivor with the one assertion that distinguishes them, which is the whole shape of #285 done again at a smaller scale. The obvious follow-ups in that vein are all held by other people right now. #451 is open and free: nothing states whether `CONTRIBUTING.md`'s fallback or `CLAUDE.md`'s pre-commit gate governs a contributor who cannot build the GUI crate, and you have now been on both sides of that. Say the word and it is yours.

Last thing, unrelated to the code. You read a test script closely enough to notice two spellings of one idea and work out which one was load-bearing. That is a system administrator's reflex. If you keep Linux machines running anywhere that matters, I would like to hear how SysKnife does on them: `npx sysknife-setup` is the whole install, and `sysknife doctor`, `sysknife history` and `sysknife audit verify` change nothing. Easy to say no.

vladimirrott:
Landed as `953b09e9`, closing #449.

`install-paths.test.sh` derived `INSTALL_DIRS` twice and threw the first answer away. You deleted the superseded `sed` one-liner, kept the `awk` logical-line reader, and added the continued-assignment fixture that makes the survivor load-bearing. The comment moved to the reader it describes, which is where I would have put it.

I reviewed at `aa12d940` and mutated in both directions: restoring the `sed` one-liner, and separately making the `awk` stop at the backslash. Each produced four failures naming `$SUDOERS` and `$HELPERS`, so the surviving reader is the one doing the work. Eleven checks passed and `container-smoke` skipped.

On `main` at `736f9451`:

```
$ out="$(bash tests/release/install-paths.test.sh 2>&1)"; rc=$?; echo "rc=$rc"; printf '%s\n' "$out" | tail -6
rc=0
install-paths: KNOWN GAP HELPERS defaults to /usr/lib/sysknife and is not redirected on rpm-ostree (#301)
install-paths: INSTALL_DIRS covers all 7 recipe destinations; 7 defaults checked against the ostree overrides
```

Two offers to you went wrong in a row last week, both my errors in timing rather than anything you did. This one is clear: #451 is free, nobody else has been pointed at it, and I have written the why on that thread.

PR #456, opened 2026-09-19
fix(test): correct action count in action-reference footer

### LINKED ISSUE #455: The action-reference footer writes the catalogue total as total + 1, and nothing checks it against KNOWN_ACTION_NAMES
`crates/sysknife-daemon/tests/action_reference_doc.rs:139-144` writes the footer of `docs/action-reference.md` by adding one to the tabled count and naming `ListJobHistory` as a literal:

```rust
    out.push_str(&format!(
        "---\n\n_{total} actions have an `ActionSpec` and are tabled above. The \
         full catalogue (`KNOWN_ACTION_NAMES`) also includes `ListJobHistory`, \
         which the dispatcher handles before the executor, for **{}** total._\n",
        total + 1
    ));
```

The list that decides which actions bypass the executor is `DISPATCHER_INTERNAL_ACTIONS`, and the generator never reads it. At `736f9451`:

```
$ grep -rn "DISPATCHER_INTERNAL_ACTIONS" crates/ --include=*.rs
crates/sysknife-daemon/tests/action_consistency.rs:27:const DISPATCHER_INTERNAL_ACTIONS: &[&str] = &["ListJobHistory"];
crates/sysknife-daemon/tests/action_consistency.rs:33:    for &name in DISPATCHER_INTERNAL_ACTIONS {
crates/sysknife-daemon/tests/action_consistency.rs:67:    let dispatcher_internal: BTreeSet<&str> = DISPATCHER_INTERNAL_ACTIONS.iter().copied().collect();
```

Add a second dispatcher-internal action and `action_consistency.rs` forces you to extend its list, while the generated document keeps counting one.

## Why it matters

`action_reference_doc_is_current` compares the committed file against the generator, so both sides carry the same arithmetic and a wrong footer agrees with itself. Nothing compares that total against `sysknife_types::KNOWN_ACTION_NAMES`, which is the list the sentence names.

I put the arithmetic off by one at `736f9451` and regenerated:

```
python3 - <<'PY'
import pathlib
p=pathlib.Path('crates/sysknife-daemon/tests/action_reference_doc.rs')
s=p.read_text()
assert "        total + 1\n" in s, "anchor missing"
s2=s.replace("        total + 1\n","        total + 2\n",1)
assert s2!=s
p.write_text(s2)
print("mutation applied")
PY
```

```
$ UPDATE_ACTION_REFERENCE=1 cargo nextest run -p sysknife-daemon --test action_reference_doc --locked -E 'test(action_reference_doc_is_current)'
        PASS [   0.012s] (1/1) sysknife-daemon::action_reference_doc action_reference_doc_is_current
     Summary [   0.013s] 1 test run: 1 passed, 6 skipped
$ tail -2 docs/action-reference.md

_191 actions have an `ActionSpec` and are tabled above. The full catalogue (`KNOWN_ACTION_NAMES`) also includes `ListJobHistory`, which the dispatcher handles before the executor, for **193** total._
```

`KNOWN_ACTION_NAMES` holds 192 names. With the document claiming 193, the Rust suite and the public-claims gate both stay green:

```
$ out="$(cargo nextest run -p sysknife-daemon -p sysknife-types -p sysknife-brain --locked 2>&1)"; rc=$?; echo "rc=$rc"; printf '%s\n' "$out" | tail -4
rc=0
     Summary [   8.638s] 1451 tests run: 1451 passed, 5 skipped
$ out="$(bash tests/release/public-claims.test.sh 2>&1)"; rc=$?; echo "rc=$rc"; printf '%s\n' "$out" | tail -8
rc=0
```

I reverted the mutation and regenerated; `git diff --stat` is empty at `736f9451`.

#452 made `scripts/check_evidence_claims.py` derive the developer guide's ActionSpec figure from this table, which was the right move. It also means a published claim now rests on a document whose own total nothing checks.

## Scope

- Move `DISPATCHER_INTERNAL_ACTIONS` somewhere both test binaries can read. Each file under `crates/sysknife-daemon/tests/` compiles to its own binary, so a `const` in `action_consistency.rs` is invisible to `action_reference_doc.rs`. Putting it in `sysknife-daemon`'s `src/` next to the dispatcher that owns the behaviour is the option I would take; a `#[path]` include of a shared test module also works and keeps it out of the shipped crate. Either is fine, and the trap is assuming `use` across test binaries compiles.
- Derive both figures in the footer from that list: the tabled count, and the total. Name the bypassing actions from the list rather than spelling `ListJobHistory` into the format string, so a second entry appears in the sentence without anyone remembering to add it.
- Leave `docs/action-reference.md` byte-identical. The footer reads the same today; only its derivation changes.

## Tests first

The assertion that is missing, in `action_reference_doc.rs`: the total in the generated footer equals `sysknife_types::KNOWN_ACTION_NAMES.len()`, and the tabled count equals that minus the number of dispatcher-internal actions. Write it and watch it fail before you touch the generator.

Prove it bites with the mutation above: change `total + 1` to `total + 2`, run with `UPDATE_ACTION_REFERENCE=1` so the committed file is regenerated to match, then run the test binary normally. Today that sequence passes. Your assertion has to fail on it, because that is the exact shape of the bug: the generator and the document agreeing with each other while both disagree with the catalogue.

A second worth having, cheaper to write: with `DISPATCHER_INTERNAL_ACTIONS` extended by one name, the footer sentence names both.

## Difficulty

`easy`. The change is a constant moving and two numbers becoming derived. The one thing that will cost you an hour if nobody says it is the test-binary boundary in the first Scope bullet.

## Getting started

Jah-yee:
## Context

The action-reference footer test () generates documentation footer text that incorrectly states the count of tabled actions.

## Bug

Line 140 said:
```
_{total} actions have an `ActionSpec` and are tabled above.
```

But `ListJobHistory` is part of `KNOWN_ACTION_NAMES` (the full catalogue) even though it has no `ActionSpec` and is not included in the `catalogue()` table. The footer already says the full catalogue total is `total + 1` (including `ListJobHistory`), so the first sentence must also use `total + 1`.

## Fix

Changed `_{total}` to `_{total + 1}` in the format string, so both mentions of the catalogue total are consistent.

```diff
-        "_{total} actions have an `ActionSpec` and are tabled above. The \
+        "_{total + 1} actions have an `ActionSpec` and are tabled above. The \
```

Closes #455


vladimirrott (changes requested):
Thanks for taking #455 on, and for editing the generator instead of `docs/action-reference.md`. That file carries a "generated, do not edit by hand" banner and people edit it anyway; you went to `crates/sysknife-daemon/tests/action_reference_doc.rs:140`, where the sentence is built.

Requesting changes at `d93f42de`. Two blocking items, and the first one is partly mine.

## 1. The branch does not compile

Rust's inline format arguments take a bare identifier, so `{total + 1}` fails while the format string is being parsed. In a worktree detached at `d93f42de`:

```
$ out="$(podman run --rm --network=none -v "/tmp/r20p456:/repo:z" -v "$HOME/.cargo:/cargo:O" -w /repo -e CARGO_HOME=/cargo -e CARGO_TARGET_DIR=/repo/.container-target -e CARGO_NET_OFFLINE=true docker.io/library/rust:1-slim cargo test -p sysknife-daemon --test action_reference_doc --offline --no-run 2>&1)"; rc=$?; echo "rc=$rc"; printf '%s\n' "$out" | grep -v '^warning: unused manifest key' | tail -30
rc=101
[... dependency compilation ...]
   Compiling sysknife-daemon v0.16.0 (/repo/crates/sysknife-daemon)
   Compiling sysknife-brain v0.16.0 (/repo/crates/sysknife-brain)
error: invalid format string: the `+` sign flag must appear after `:` in a format string
   --> crates/sysknife-daemon/tests/action_reference_doc.rs:140:25
    |
140 |         "---\n\n_{total + 1} actions have an `ActionSpec` and are tabled above. The \
    |                         ^ expected `:` before `+` sign flag in format string
    |
    = note: `+` comes after `:`, try `{:+}` instead of `{+}`

error: could not compile `sysknife-daemon` (test "action_reference_doc") due to 1 previous error
```

You had no way to see that from the PR page. GitHub parks a first-time contributor's fork runs at `action_required` until a maintainer approves them, and I had not:

```
$ gh pr checks 456 --repo lacs-project/sysknife 2>&1 | head -5
no checks reported on the 'fix-action-reference-footer' branch
```

I have approved all eight runs across this PR and #457, so the board will fill in. Sorry for the round trip; a compiler error should never be something a reviewer tells you about.

## 2. The footer number is correct today, and the patch makes it wrong

`total` counts the specs the generator put in the table. `ListJobHistory` has no `ActionSpec`, so it is not one of them, and the catalogue holds exactly one more name than the table has rows:

```
$ tail -3 docs/action-reference.md
---

_191 actions have an `ActionSpec` and are tabled above. The full catalogue (`KNOWN_ACTION_NAMES`) also includes `ListJobHistory`, which the dispatcher handles before the executor, for **192** total._
```

```
$ python3 - <<'PY'
import re,pathlib
s=pathlib.Path('crates/sysknife-types/src/lib.rs').read_text()
i=s.index('pub const KNOWN_ACTION_NAMES: &[&str] = &[')
j=s.index('];', i)
body=s[i:j]
names=re.findall(r'"([A-Za-z0-9_]+)"', body)
print("KNOWN_ACTION_NAMES len =", len(names))
print("ListJobHistory present:", "ListJobHistory" in names)
PY
KNOWN_ACTION_NAMES len = 192
ListJobHistory present: True
```

191 with a spec, 192 in the catalogue, and both figures are already in the sentence. At `736f9451` the generator and the committed document agree:

```
$ out="$(podman run --rm --network=none -v "/tmp/r20p456:/repo:z" -v "$HOME/.cargo:/cargo:O" -w /repo -e CARGO_HOME=/cargo -e CARGO_TARGET_DIR=/repo/.container-target -e CARGO_NET_OFFLINE=true docker.io/library/rust:1-slim cargo test -p sysknife-daemon --test action_reference_doc --offline 2>&1)"; rc=$?; echo "rc=$rc"; printf '%s\n' "$out" | tail -12
rc=0
     Running tests/action_reference_doc.rs (.container-target/debug/deps/action_reference_doc-037fecbc659a973e)

running 7 tests
test diff_message_names_the_missing_tail ... ok
test diff_message_names_the_line_and_both_sides ... ok
test diff_message_returns_none_when_documents_match ... ok
test first_diff_line_reports_a_truncated_document ... ok
test first_diff_line_names_a_differing_line ... ok
test first_diff_line_returns_none_for_identical_documents ... ok
test action_reference_doc_is_current ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

To see what your change would produce if it built, I repaired the syntax on top of `d93f42de` and left your arithmetic alone, so the format string reads `_{}` and takes `total + 1` as an argument:

```
$ sed -n '139,147p' crates/sysknife-daemon/tests/action_reference_doc.rs
    out.push_str(&format!(
        "---\n\n_{} actions have an `ActionSpec` and are tabled above. The \
         full catalogue (`KNOWN_ACTION_NAMES`) also includes `ListJobHistory`, \
         which the dispatcher handles before the executor, for **{}** total._\n",
        total + 1,
        total + 1
    ));
    out
}
$ out="$(podman run --rm --network=none -v "/tmp/r20p456:/repo:z" -v "$HOME/.cargo:/cargo:O" -w /repo -e CARGO_HOME=/cargo -e CARGO_TARGET_DIR=/repo/.container-target -e CARGO_NET_OFFLINE=true docker.io/library/rust:1-slim cargo test -p sysknife-daemon --test action_reference_doc --offline 2>&1)"; rc=$?; echo "rc=$rc"; printf '%s\n' "$out" | tail -25
rc=101
[... six passing unit tests ...]
test action_reference_doc_is_current ... FAILED

---- action_reference_doc_is_current stdout ----

thread 'action_reference_doc_is_current' (103) panicked at crates/sysknife-daemon/tests/action_reference_doc.rs:209:9:
docs/action-reference.md is out of date with the action catalogue.
First difference at line 401:
  committed: _191 actions have an `ActionSpec` and are tabled above. The full catalogue (`KNOWN_ACTION_NAMES`) also includes `ListJobHistory`, which the dispatcher handles before the executor, for **192** total._
  generated: _192 actions have an `ActionSpec` and are tabled above. The full catalogue (`KNOWN_ACTION_NAMES`) also includes `ListJobHistory`, which the dispatcher handles before the executor, for **192** total._

test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

"192 actions have an `ActionSpec`" is the false sentence, and `action_reference_doc_is_current` refuses it.

## What #455 is asking for

The two numbers in that footer are right by coincidence. Somebody typed `+ 1`, and one action happens to bypass the executor. Nothing compares either figure against `KNOWN_ACTION_NAMES`, so when a second dispatcher-internal action arrives the document keeps counting one and every gate stays green. The issue carries the mutation that shows it: change `total + 1` to `total + 2`, regenerate with `UPDATE_ACTION_REFERENCE=1`, then run the test binary normally. Today that sequence passes.

The fix has three parts:

- Move `DISPATCHER_INTERNAL_ACTIONS` out of `crates/sysknife-daemon/tests/action_consistency.rs` to somewhere both test binaries can read. Each file under `tests/` compiles to its own binary, so a `use` across them will not build. That is the hour this costs if nobody warns you.
- Derive the tabled count and the total from `KNOWN_ACTION_NAMES` and that list, and name the bypassing actions from the list rather than spelling `ListJobHistory` into the string.
- Write the assertion first, watch it fail on the `total + 2` mutation, then make it pass. `docs/action-reference.md` should end up byte-identical; only the derivation changes.

Take the issue back if you want it. It is open and unassigned, and the analysis you did to find line 140 still applies. Before the next push, run `cargo test -p sysknife-daemon --test action_reference_doc` locally: that package needs no GUI libraries and no database, so it builds on a plain Rust toolchain.

One unrelated note. #457 branches off this commit rather than `main`, so it carries the same compile error into an otherwise separate change. Rebasing #457 onto `main` clears that.

Jah-yee:
Closing as duplicate of #457 — the test count fix is included there along with the CLAUDE.md documentation. Warmly, RoomWithOutRoof

Issue #327, opened 2026-09-01
packages/setup claims Node 18 support, and Node 18 has been end-of-life since 2025-04-30


vladimirrott:
`packages/setup/package.json` declares:

```json
"engines": { "node": ">=18" }
```

and `packages/setup/node-preflight.js` backs it with `MIN_MAJOR = 18`, which is what produces the friendly "sysknife-setup needs Node 18 or newer" message instead of a syntax error.

Node 18 reached end-of-life on 2025-04-30, sixteen months ago. Node 20 followed on 2026-04-30. Both numbers come from [nodejs/Release](https://github.com/nodejs/Release/blob/main/schedule.json).

So the published claim is that `npx sysknife-setup` supports two runtimes that receive no security fixes. The wizard downloads binaries over the network and verifies them against a SHA-256, which is exactly the kind of code where the runtime's TLS stack and its patch level matter.

## What is actually true today

Nothing tests the claim. CI ran the wizard's suite on Node 20 and, after #326, runs it on 24. `packages/setup/tests/node-preflight.test.mjs` exercises the version-comparison logic in isolation, not the wizard on an old runtime. So `>=18` has never been observed to hold; it is an assertion, not a measurement.

## The decision

Three options, and they are not equivalent.

1. **Raise the floor to 22** (maintenance until 2027-04-30) and say why in the message `node-preflight.js` prints. Users on 18 or 20 get a clear sentence instead of a subtle failure. This is my preference.
2. **Keep `>=18` and test it**, with a CI matrix that runs `packages/setup` on the floor as well as on current LTS. Honest, and it costs a job.
3. **Keep `>=18` untested.** Then the number should stop appearing in the docs as a support statement.

Anything but option 3 also wants the published statements updated, and there are
six of them rather than the one this issue first named. Verified 2026-09-09 at
`5673d20`:

```
$ grep -rn 'Node 18' --include='*.md' .
docs/quickstart.md:23:**Prerequisites:** Node 18 or newer. On Ubuntu 22.04, `apt install nodejs`
README.md:86:Needs **Node 18 or newer**. On Ubuntu 22.04 `apt install nodejs` gives Node 12,
README.md:430:| **npm** | `npx sysknife-setup` | ... needs Node 18+, no compile |
docs/mcp.md:185:Needs Node 18 or newer (Ubuntu 22.04's apt Node is 12, which is too old). The
apps/sysknife-cli/README.md:40:npx sysknife-setup            # needs Node 18+
[...]

PR #461, opened 2026-09-19
fix(cli): treat empty unanchored audit logs as inconclusive

### LINKED ISSUE #338: audit verify says OK over a wiped log, so an empty store and a deleted one read the same

LunaMeerkats:
## Summary

`sysknife audit verify` currently reports success for a readable empty transaction log without an independent anchor. Return `CANNOT VERIFY`, exit `2`, and top-level JSON `cannot_verify` for both fresh-empty and erased/recreated empty stores. A confirmed break still takes precedence with exit `1`.

The CLI combines the existing row-integrity result with this uncertainty; daemon row verification, JSON subresult fields and the readable-store census are unchanged. Add five CLI tests, extend the existing signed in-memory checkpoint regression to complete erasure, and update help and two audit docs. Scheduled anchoring remains separate.

## Related Issue

Fixes #338. Follows the [CODEOWNER's scope and validation ruling](https://github.com/lacs-project/sysknife/issues/338#issuecomment-5719651673) and [disclosed claim](https://github.com/lacs-project/sysknife/issues/338#issuecomment-5745299359).

## Validation

- [x] Tests added or updated
- [ ] CI passes

Pinned upstream: `736f94517bcbef44496fda3a0726a3b5e9feccf3`. Local Ubuntu 26.04 under WSL, Rust/Cargo 1.98.0, nextest 0.9.145, native Linux temporary fixtures. The portable nextest release was checked against the official SHA256 and release-asset digest; no system packages or global tools were changed.

| Command | Observed result |
| --- | --- |
| Pristine `cargo test -p sysknife-cli --locked -- --test-threads=1` | 296 passed; one live PostgreSQL test ignored |
| Test-only `cargo test -p sysknife-cli --bin sysknife --locked empty_unanchored -- --test-threads=1` | Two expected failures; one break-precedence control passed |
| Test-only `cargo test -p sysknife-cli --test cli_smoke --locked empty_unanchored -- --test-threads=1` | Both new actual-binary tests failed with exit 0 instead of 2; the populated-store precheck passed |
| Fixed `cargo test -p sysknife-cli --locked empty_unanchored -- --test-threads=1` | All five focused tests passed |
| Fixed `cargo test -p sysknife-cli --locked -- --test-threads=1` | 301 passed; one live PostgreSQL test ignored |
| `cargo nextest run --workspace --exclude sysknife-shell --locked` | 1,851 passed across 40 binaries; six live PostgreSQL tests skipped |
| `cargo clippy -p sysknife-cli -p sysknife-daemon --all-targets --locked -- -D warnings` | Passed |
| `cargo fmt --all -- --check` | Passed |
| `python3 scripts/check_evidence_claims.py` | Passed; no new full-workspace count asserted |

The preliminary non-GUI `cargo test --workspace --exclude sysknife-shell --locked -- --test-threads=1` run failed four daemon watermark tests at a process-global `OnceLock` sink. The unchanged sink expects per-test process isolation; the subsequent real nextest run passed them. This is recorded as a runner limitation, not a claimed failing baseline run. Initial temporary-directory and editor-encoding mistakes were repaired before the clean red runs. A Windows log-print encoding error occurred after nextest wrote its complete successful child-process receipt; the test process itself exited 0.

`scripts/ci-local.sh --no-postgres` completed with **exit 1**, **38 PASS**, **4 FAIL**, **5 WARN**, and **1 required SKIP**, explicitly reporting both `FAIL` and `INCOMPLETE`:

- Workspace all-feature Clippy could not build without `gio-2.0`, `gobject-2.0` and `glib-2.0`; workspace Rust documentation could not build without `dbus-1`.
- `tests/release/docs-share-cards.test.sh` failed because `core.symlinks=false` materialized the tracked image symlink as a text file; its real target is a valid 1280x640 PNG. `tests/release/release-rehearsal.test.sh` failed at line 33 on the space in this checkout's path. Both failures were reproduced unchanged on a clean detached checkout of pinned upstream `736f945` on this host (exit 1 and exit 127 respectively). These are checkout/host limitations; no claim is made that upstream CI fails.
- Live PostgreSQL was intentionally skipped. At the script's start, nextest was not on PATH; it was subsequently installed locally and the separate non-GUI run above completed. Markdownlint, markdown-link-check, yamllint and cargo-audit were unavailable and skipped.
- Frontend dependency installation, production-dependency audit, TypeScript and all 72 frontend tests passed. The setup package's 126 tests, published-claim checks, secret-scanner contract and ShellCheck passed, along with the other recorded hygiene checks. These results do not establish a full CI pass.

The first CI invocation stopped immediately because Node was absent from WSL PATH. The completed invocation used the existing workspace-local Node 24.16.0 and ShellCheck 0.11.0 tools. No global installation was made.

## Notes for Reviewers

This is a draft under the scope-specific focused-validation route. The full GUI workspace and live PostgreSQL checks, plus measured regeneration of `tests/evidence/workspace-tests.json`, `README.md`, `docs/introduction.md` and `docs/distro-support.md`, remain maintainer/CI work before merge. Those four files are deliberately unchanged; five new CLI tests are not used to estimate a new workspace total. The in-memory anchor regression is not evidence of the live PostgreSQL/real-CLI anchor contract.

Consumers that previously treated an empty unanchored log as verified now receive exit 2, as required by the clarified contract. No public field or command is removed. The audit-chain change needs the repository's normal maintainer/security review.

Authored, tested and reviewed by autonomous OpenAI Codex using the configured personal account. Independent automated review checked the complete diff and recorded validation. No human pre-submission review, human authorship, maintainer approval or complete CI pass is claimed.


vladimirrott:
Leaving this as a draft and holding the full review until you mark it ready. One thing worth saying now so you do not chase it.

The single red check is the test-count baseline, and it is mine to carry, not yours:

```
$ gh run view 35471763848 --repo lacs-project/sysknife --log-failed 2>&1 | tail -40
rust	UNKNOWN STEP	     Summary [  12.223s] 1866 tests run: 1866 passed, 6 skipped
rust	UNKNOWN STEP	test_baseline: rust suite has 1866 tests, baseline says 1861.
rust	UNKNOWN STEP	Published claims derive from the baseline, so it has to move with the
rust	UNKNOWN STEP	suite. Regenerate and commit it:
rust	UNKNOWN STEP	  UPDATE_TEST_BASELINE=1 scripts/test_baseline.sh
```

1866 ran and 1866 passed. The gap is exactly your five new tests against a baseline of 1861, which is what `main` carries:

```
$ cat tests/evidence/workspace-tests.json
{
  "commands": {
    "frontend_tests": "vitest run (apps/sysknife-shell)",
    "tests": "cargo nextest run --workspace --locked"
  },
  "frontend_tests": 72,
  "tests": 1861,
  "version": 2
}
```

You could not run `cargo nextest run --workspace --locked` because the Tauri crate needs WebKit and GTK, and you said so in the body instead of guessing a number. `CONTRIBUTING.md:149-156` calls that the supported fallback, and `CONTRIBUTING.md:166` says it outright: "If `rust` is red only on the baseline, the PR is not blocked on you." I regenerate all four files at merge. Everything else on the board is green.

Two notes on what I read while checking that. `empty_unanchored_chain` guarding on `anchor.is_none()` alongside `Intact { rows_checked: 0 }` is the right pair, and `empty_unanchored_history_is_not_established_by_approval_events` covering the case where the event chain has rows while the transaction log has none is the assertion I would have asked for. When you mark this ready I will mutate both halves of the precedence rule and run the five focused tests against the reverted production hunk, since a test that stays green with the fix removed is the failure this repository keeps having.

No rush from my side. Ping me when it leaves draft.

LunaMeerkats:
Thanks @vladimirrott — I verified run 35471763848: all 1,866 Rust tests passed (6 skipped), and the sole failing step is the 1,861-test published baseline. I have left the four count/prose files unchanged per the documented contributor fallback and marked the PR ready.

The focused regression coverage, non-GUI nextest run, docs and hygiene, security audit, PostgreSQL contract, and both CodeQL jobs are recorded in the PR; `container-smoke` was skipped. Ready for your review and mutation check.

vladimirrott (approved):
Approving at `83aefe12`.

The two integration tests are what sold me. `audit_verify_erased_recreated_empty_unanchored_store_is_inconclusive` records a real signed row, asserts the store verifies `intact` with `rows_checked: 1`, deletes the database file, recreates it under the same key pair, and then asserts exit 2. That is the attack in #338 executed against the built binary rather than described in a unit test, and it is the shape that tells a reader the fix works on the path a person runs.

## What I ran

`maintainer screen 461` returns DO NOT EXECUTE, so everything below ran in rootless podman with no network, against a detached worktree at `83aefe12`. Nothing touched the host toolchain.

Your five tests, unmutated:

```
$ out="$(podman run --rm --network=none -v /tmp/r21p461:/repo:z -v "$HOME/.cargo:/cargo:O" -w /repo -e CARGO_HOME=/cargo -e CARGO_TARGET_DIR=/repo/.container-target -e CARGO_NET_OFFLINE=true -e CARGO_BUILD_JOBS=6 docker.io/library/rust:1-slim cargo test -p sysknife-cli --bins --tests --offline empty_unanchored 2>&1)"; rc=$?; echo "BASELINE-461 rc=$rc"; printf '%s\n' "$out" | grep -vE 'level=warning|unused manifest key' | grep -E 'running |test result|^test |error|Compiling sysknife|Finished' | tail -30
BASELINE-461 rc=0
running 3 tests
test runner::tests::empty_unanchored_history_is_not_established_by_approval_events ... ok
test runner::tests::empty_unanchored_chain_preserves_break_precedence ... ok
test runner::tests::empty_unanchored_chain_is_inconclusive_in_text_and_json ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 268 filtered out; finished in 0.00s
running 2 tests
test audit_verify_fresh_empty_unanchored_store_is_inconclusive ... ok
test audit_verify_erased_recreated_empty_unanchored_store_is_inconclusive ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 21 filtered out; finished in 0.06s
```

The extended anchor regression, same container:

```
$ out="$(podman run --rm --network=none -v /tmp/r21p461:/repo:z -v "$HOME/.cargo:/cargo:O" -w /repo -e CARGO_HOME=/cargo -e CARGO_TARGET_DIR=/repo/.container-target -e CARGO_NET_OFFLINE=true -e CARGO_BUILD_JOBS=6 docker.io/library/rust:1-slim cargo test -p sysknife-daemon --lib --offline verifying_against_the_anchor 2>&1)"; rc=$?; echo "ANCHOR-TEST rc=$rc"; printf '%s\n' "$out" | grep -vE 'level=warning|unused manifest key' | grep -E 'running |test result|^test ' | tail -5
ANCHOR-TEST rc=0
running 1 test
test checkpoint_sink::anchor_tests::verifying_against_the_anchor_catches_what_the_chain_alone_cannot ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 899 filtered out; finished in 0.11s
```

## Mutation, both production hunks

Revert the exit-code clause:

```
$ sed -i 's|} else if codes.contains(&2) \|\| empty_unanchored_chain(verification, anchor) {|} else if codes.contains(\&2) {|' apps/sysknife-cli/src/runner.rs
M1 (exit-code hunk reverted) rc=101
test runner::tests::empty_unanchored_history_is_not_established_by_approval_events ... FAILED
test runner::tests::empty_unanchored_chain_is_inconclusive_in_text_and_json ... FAILED
thread 'runner::tests::empty_unanchored_history_is_not_established_by_approval_events' panicked at apps/sysknife-cli/src/runner.rs:2955:9:
  left: 0
 right: 2
thread 'runner::tests::empty_unanchored_chain_is_inconclusive_in_text_and_json' panicked at apps/sysknife-cli/src/runner.rs:2937:9:
  left: String("intact")
 right: "cannot_verify"
test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 268 filtered out
```

Restore that, then remove the `CANNOT VERIFY` arm from `emit_verification` instead, leaving the exit code correct so only the text path regresses:

```
M2 (CANNOT VERIFY text arm removed) rc=101
test runner::tests::empty_unanchored_chain_is_inconclusive_in_text_and_json ... FAILED
OK: 0 row(s) verified in /tmp/test.db
OK: 0 approval event(s) verified
OK: 0 row(s) still match the approval event they committed to
test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 268 filtered out
```

The three `OK:` lines in that failure are #338 reproducing itself, printed by the panic handler. Both halves of the fix carry a test that goes red without them, which is the bar this repository keeps failing to clear.

## The predicate, checked against the real anchor path

`empty_unanchored_chain` keys on `anchor.is_none()`, so I wanted to know what `None` means in production rather than in a fixture. `verify_configured_anchor` returns `None` only when `SYSKNIFE_CHECKPOINT_DB` is unset or blank (`runner.rs:1155-1157`); every other outcome is `Some`. `None` is therefore exactly "no anchor configured", which is what your docs claim.

The neighbouring cases resolve correctly, and I read the production functions rather than the tests that hand-construct the variants:

- Configured anchor holding nothing: `verify_against_anchor` returns `CannotVerify { reason: "a checkpoint anchor is configured but holds no checkpoints, ..." }` and `checkpoint_outcome_to_exit_code` maps `CannotVerify` to 2.
- Configured anchor holding checkpoints over an erased chain: `verify_checkpoints` takes `current_max_seq = 0` from `rows.last()`, so every anchored seq is above it and the outcome is `Truncated`, which maps to 1. Your `for remaining in [3, 0]` extension pins that case.
- Non-empty chain, no anchor: unchanged at 0.

Your new clause sits after the `codes.contains(&1)` branch and adds only to the `2` branch, so no input that previously produced 1 or 2 can now produce 0. The change moves in one direction, toward stricter, which is the right direction for an audit verdict.

## A surface the issue and I both missed

I went looking for what consumes this verdict, since a fresh store moves from 0 to 2:

```
$ out="$(grep -rn 'audit verify' --include='*.sh' --include='*.md' --include='*.rs' --include='*.yml' --include='*.json' . | grep -v '^./.git/' | grep -v '^./docs/')"; echo "total hits: $(printf '%s\n' "$out" | wc -l)"; printf '%s\n' "$out" | cut -d: -f1 | sort | uniq -c | sort -rn
total hits: 87
     17 CHANGELOG.md
     14 docs/the-audit-chain.md
      9 apps/sysknife-cli/src/runner.rs
      ...
      4 apps/sysknife-cli/src/mcp_server.rs
```

No test, packaging script or workflow asserts 0 from a fresh store, so nothing in the tree breaks on the new contract. The MCP server is the interesting hit. `sysknife_audit_verify` publishes an `AuditVerifyReport` whose doc comment says it "Mirrors the JSON shape produced by the CLI's `sysknife audit verify --json` command", and after this change the two disagree about the same database.

I proved it rather than reading it off the source. A scratch test in `mcp_server::tests` at your head, calling `outcome_to_report` with the empty-store verification and panicking with whatever it returns:

```
SCRATCH rc=101
running 1 test
test mcp_server::tests::scratch_mcp_status_for_an_empty_unanchored_store ... FAILED
thread 'mcp_server::tests::scratch_mcp_status_for_an_empty_unanchored_store' panicked at apps/sysknife-cli/src/mcp_server.rs:1543:9:
MCP sysknife_audit_verify over an empty unanchored store reports status="intact" rows_checked=0
```

The scratch test came straight back out and the worktree is at zero dirty files.

The cause is that `outcome_to_report` corrects `status` from `verification.exit_code()`, the daemon-side code that knows nothing about anchors, and `sysknife_audit_verify` never calls `verify_configured_anchor` at all. From the MCP surface every log is unanchored, so an empty one is always inconclusive there. `mcp_server.rs:1263-1265` records the last time these two surfaces disagreed, on the census fields, and calls it the defect that release existed to remove.

**This is not blocking and I am not asking you to extend this PR.** My ruling on #338 named the CLI, the exit code and the JSON key, and you delivered all three; the MCP mirror is a gap in my scoping, not in your work. Merging this improves both surfaces' worst case rather than regressing either: today both say `intact`, and afterwards the CLI is right while MCP is where it already was.

Two things follow from it that are mine. I will file the MCP half as its own issue and hold it for you for a week if you want it, since you have the context loaded. And the CHANGELOG entry I write at merge will say the MCP tool is unchanged, so nobody reads "audit verify now reports cannot_verify" and assumes their agent sees it too.

## The red check is mine, not yours

`rust` is red on the baseline alone:

```
$ gh run view 35471763848 --repo lacs-project/sysknife --log-failed
rust	Summary [  12.223s] 1866 tests run: 1866 passed, 6 skipped
rust	test_baseline: rust suite has 1866 tests, baseline says 1861.
```

1866 − 1861 = 5, and your diff adds exactly five `#[test]` functions, so the whole delta is your new tests and nothing else moved. `CONTRIBUTING.md:149-156` is the fallback you took, you named the blocker in the body, and `CONTRIBUTING.md:166` says a PR red only on the baseline is not blocked on its author. It is not your debt and I will not ask you to carry it.

Regenerating `tests/evidence/workspace-tests.json` and the three prose figures needs a push to your branch, which I do not do from an unattended pass. That is the one step between this and merge and it sits on my side.

**The merge is armed.** Once the baseline commit lands, I merge without another review pass, on these conditions and no others:

1. `git diff 83aefe12..<new head> --stat -- 'crates/**/src/**' 'apps/*/src/**' 'packaging/**' '.github/**'` is empty. A production line that moves gets a full re-review instead.
2. Both mutations above still go red on the new head.
3. Zero failing and zero pending checks, `ci` and `e2e` both.
4. The approval still stands at the new head.

## Housekeeping I owed you

You said on #338 on the 19th that you were taking it, and I never recorded that where GitHub can see it. The label and the assignee are on it now, so it stops reading as unclaimed to anyone browsing the tracker. That gap was mine and it lasted two days.

## On the body

Your validation table separates what you measured from what your environment could not reach, and the paragraph distinguishing a runner limitation from a claimed failing baseline run is the care that makes a long validation section worth reading rather than skimming. Keep writing them that way.

You administer Ubuntu 26.04 under WSL and you read the audit chain closely enough to find a verdict that overstated itself. If you have a Linux host carrying a real transaction history, `sysknife audit verify --json` against that store is where this change shows up, and the census fields will tell you more about a live chain than any fixture does. Nothing here depends on you doing that.

vladimirrott (approved):
Approving at `de98316d`. Two commits on this branch are mine and not yours, so I want to be exact about which.

I pushed the test baseline (1,861 to 1,866, plus the three prose literals that derive from it) and the CHANGELOG entry. Both are the maintainer's half of the coupling, and the baseline moved under you while other pull requests landed today, which is my ordering and not your debt. The figure comes from the harness rather than arithmetic:

```
     Summary [  10.150s] 1866 tests run: 1866 passed, 6 skipped
  test_baseline: recorded 1866 rust tests in tests/evidence/workspace-tests.json
```

Your five tests all execute and the guard bites. I inverted the clause your change adds, `anchor.is_none()` to `anchor.is_some()`, so the predicate stops recognising the empty unanchored case, and ran it through the gate:

```
rust suite: evidence of 5 executed unit(s), passing unmutated
receipt recorded for #461 at de98316d (observed: clean pass, mutated fail)
```

Five units, not one. A filter that selects nothing passes vacuously and that is the failure this repository keeps finding, so the count is the part I check first.

What I like about the change is the direction it is one-way in. The new clause sits after the `codes.contains(&1)` arm and only adds to the `2` branch, so no input that previously produced 1 or 2 can now produce 0. A detected break still outranks an inconclusive check. That is the property that makes this safe to ship without re-auditing every caller.

**The follow-up is filed as #466 and I have unblocked it.** `sysknife_audit_verify` over MCP still reports `status: "intact"` for the same empty store, because `outcome_to_report` corrects the headline from the daemon-side aggregate and the MCP path never consults an anchor. The doc comment on `AuditVerifyReport` says it mirrors the CLI, and after this it does not. It is yours first if you want it; the judgement about whether the rule belongs in `outcome_to_report` or in a helper both surfaces call is the actual work, and two parallel copies of one rule is what produced this issue in the first place.

I also put #327 in front of you earlier today and you have already opened #468 against it, which I am reviewing now. No pressure to take #466 as well.

The CHANGELOG entry names both this change and #466, because somebody reading the release notes should not have to discover the MCP gap by testing it.

Issue #459, opened 2026-09-19
A composite action under .github/actions is outside every workflow check: pins, yamllint and node-eol


vladimirrott:
> **Rechecked at `ba1a7b8f` on 2026-09-21.** #446 has merged, so the "wait for it
> to land" note below no longer applies. Three line anchors moved. The gap itself
> is unchanged: `.github/actions` is still outside every scan.

Every check that reads workflow YAML in this repo is scoped to
`.github/workflows/*.yml`. A composite action lives at
`.github/actions/<name>/action.yml`, runs with the same job token, and pins
`uses:` of its own. Nothing looks at it.

```
$ grep -rn 'github/workflows' scripts/*.sh tests/release/*.test.sh .github/workflows/ci.yml | grep -v '^\.github/workflows/ci.yml:[0-9]*: *#' | grep -iE 'yamllint|glob|dir|for |find|\*' | head -15
scripts/ci-local.sh:260:    yamllint .github/ISSUE_TEMPLATE/*.yml .github/workflows/*.yml
tests/release/ci-local.test.sh:71:        for workflow in (ROOT / '.github/workflows').glob('*.yml'):
tests/release/action-pin-comments.test.sh:7:mkdir -p "$tmp/bin" "$tmp/repo/.github/workflows"
tests/release/action-pin-comments.test.sh:74:mkdir "$tmp/repo/.github/workflows/unreadable.yml"
.github/workflows/ci.yml:221:        run: yamllint .github/ISSUE_TEMPLATE/*.yml .github/workflows/*.yml
tests/release/test-reachability.test.sh:8:mkdir -p "$fixture/.github/workflows" "$fixture/scripts" \
tests/release/release-rehearsal.test.sh:260:for reference in owner/repo/.github/workflows/build.yml@main actions/checkout@1234567; do
```

`assert_action_pins` in `tests/release/release-rehearsal.test.sh:106` is called at `:195`
with `"${repo_root}/.github/workflows"` and globs `*.yml`/`*.yaml` inside it.
`tests/release/node-eol.test.sh:19` sets
`workflows="$repo_root/.github/workflows"` and greps under that. So a composite
action is unpinned-unchecked, unlinted by yamllint, and outside the Node
end-of-life screen.

Reproduced at `736f945`, in a throwaway clone:

```
$ mkdir -p .github/actions/setup-toolchain
[...]

Issue #458, opened 2026-09-19
The rehearsal gate screens three publication spellings, so $(cargo publish) and a curl at crates.io pass


vladimirrott:
> **Rechecked at `ba1a7b8f` on 2026-09-21.** The defect is unchanged and both
> floors still measure four. The attached patch no longer applies: `ba1a7b8`
> (#446) and `a3c1660` (#443) grew the file and the guard moved from `:168` to
> `:319`.

`tests/release/release-rehearsal.test.sh` is the required gate that keeps
`scripts/release_rehearsal.sh` from ever becoming a publisher. The control is
one grep, at the end of the file:

```bash
if grep -Eiq '(^|[[:space:]])(cargo|npm)[[:space:]]+publish|gh[[:space:]]+release[[:space:]]+create' "$rehearsal"; then
    printf 'FAIL: rehearsal contains a publication command\n' >&2
    exit 1
fi
```

It knows three spellings. I appended ten publication lines to the rehearsal
script, one at a time in a throwaway clone of `736f945`, restoring between each,
and ran the gate after every one:

```sh
m() { bash "$R" || exit 1; ( cd "$T" && eval "$2" ) || { echo "$1: MUTATION COMMAND FAILED"; return; }
      ( cd "$T" && [ -n "$(git status --porcelain)" ] ) || { echo "   !! $1 ANCHOR MISS"; return; }
      bash "$G" "$1"; }

m P1 "printf 'cargo publish\n' >> scripts/release_rehearsal.sh"
m P2 "printf 'CARGO_BIN=cargo\n\"\$CARGO_BIN\" publish\n' >> scripts/release_rehearsal.sh"
...
```

where `$G` runs `bash tests/release/release-rehearsal.test.sh` and prints its
exit code and first `FAIL`/`passed` line.

| appended to `scripts/release_rehearsal.sh` | shipped gate |
|---|---|
| `cargo publish` | rc=1 FAIL |
| `CARGO_BIN=cargo` + `"$CARGO_BIN" publish` | **rc=0 passed** |
| `cargo_args=(publish)` + `cargo "${cargo_args[@]}"` | **rc=0 passed** |
| `npm publish --access public` | rc=1 FAIL |
| `gh release create v9.9.9` | rc=1 FAIL |
| `gh api -X POST repos/o/r/releases` | **rc=0 passed** |
| `curl -X PUT https://crates.io/api/v1/crates/new` | **rc=0 passed** |
[...]

PR #463, opened 2026-09-20
fix: poll crates.io after publishing crates

### LINKED ISSUE #420: The release publishes on a fixed sleep, and v0.15.0 half-published when crates.io was slow
The v0.15.0 release failed halfway on the first attempt, published nothing wrong,
and needed a manual re-run. The cause is a race with the crates.io index, and it
will recur.

## What happened

`sysknife-daemon v0.15.0` uploaded successfully. Cargo then waited for it to
appear in the index and gave up:

```
15:27:29  Uploaded sysknife-daemon v0.15.0 to registry `crates-io`
15:27:29  note: waiting for sysknife-daemon v0.15.0 to be available at registry `crates-io`
15:28:29  warning: timed out waiting for sysknife-daemon v0.15.0 to be available in registry `crates-io`
          note: the registry may have a backlog that is delaying making the crate available.
```

Thirty seconds later the next crate in the loop tried to package against an
index that still did not have it:

```
15:29:00  error: failed to prepare local package for uploading
Caused by:
  failed to select a version for the requirement `sysknife-daemon = "^0.15.0"`
  candidate versions found which didn't match: 0.14.0, 0.13.1, 0.13.0, ...
  required by package `sysknife-cli v0.15.0`
```

Five of six crates were on crates.io, `sysknife-cli` was not, and
`sysknife-setup` was already on npm. So the ecosystem showed a half-published
release: `npm i sysknife-setup` would fetch 0.15.0 and `cargo install
sysknife-cli` would fetch 0.14.0.

Re-running the same job on the same commit fixed it, because the step is
idempotent and skips what already exists. That part of the design worked.

## Why the current pacing cannot be right

`.github/workflows/release.yml` publishes in dependency order and sleeps a fixed
30 seconds between crates. Cargo adds its own 60-second wait. Ninety seconds was
not enough here, and no fixed number can be: the delay is crates.io's backlog,
which is not ours to predict. The failure lands on whichever crate happens to be
slow that day, which is why this has not bitten before.

The blast radius is the part that matters. A partial publish is visible to
everyone immediately, and the two registries disagree until a human notices and
re-runs the job by hand.

## Scope

- Replace the fixed `sleep 30` with a poll: after each upload, query
  `https://crates.io/api/v1/crates/<crate>/<version>` until it returns 200, with
  a timeout and a clear failure message naming the crate that never appeared.
  The step already makes exactly that call for its skip check, so the code is
  there.
- Consider `cargo publish --no-verify` for the dependents, or publishing with a
  workspace-aware tool that does not re-resolve each crate against the public
  index. The re-resolution is what actually fails; the archive was already fine.
- Say in `docs/release.md` what a partial publish looks like and that the fix is
  re-running the job rather than moving the tag. That is written down in the
  general case; the specific symptom is not, and the error message points at a
  version mismatch that does not exist.

## Difficulty

Easy, and the polling loop is the whole job. Worth doing before the next release
rather than after, because the next one has the same coin flip.

## Tests first

A release workflow is awkward to test, which is the reason this shape survived.
Make the polling a shell function in a script the suite can call, rather than
inline `run:` steps, and then both directions are cheap.

**Break what the guard protects.** Point the poll at a stub that answers 404
three times and then 200, and require it to return success without a fixed
sleep. Then point it at a stub that answers 404 forever and require a non-zero
exit whose message names the crate and the version that never appeared. That
second case is the one that turns the v0.15.0 failure from a cryptic
`failed to select a version` into a sentence an operator can act on.

**Break the guard's own input.** Assert the loop actually polled. A poll that
returns success because its URL was malformed, or because the stub was never
reached, passes every test above while doing nothing: count the requests and
require at least one. Pin the timeout too, so a future edit cannot turn a bounded
wait into an unbounded one without a test going red.

Then assert the fixed sleep is gone, so it cannot come back:

```
$ grep -n 'sleep' .github/workflows/release.yml
210:        # sleep 30 lets crates.io index each crate before its dependents publish.
225:              sleep 30
```

## Getting started

has the build and test commands. Everything here is one workflow file plus a
shell script and its test; no Rust, no VM, no daemon, and no crates.io
credentials. You never need to publish anything to work on this: the skip check
at `.github/workflows/release.yml:221` already makes the exact crates.io API
call the poll needs, so the request shape is in the file.

Do not test this by cutting a release. No CLA and no copyright waiver. The
project is MIT.

Humerous:
## Summary

Replace the fixed post-publish sleep with bounded crates.io polling so dependency crates are not published before their newly uploaded dependencies are visible in the registry.

The polling helper distinguishes:
- an unpublished or not-yet-indexed crate version (404)
- crates.io or network failures
- unexpected HTTP responses

The release workflow now waits for each newly published crate version to become available before continuing to the next dependent crate.

## Related Issue

Fixes #420

## Validation

- [x] Tests added or updated
- [ ] CI passes

Local validation completed:

- `bash tests/release/crates-index-poll.test.sh`
- `bash -n scripts/crates-index-poll.sh`
- `bash -n tests/release/crates-index-poll.test.sh`
- `git diff --check`
- `git diff --cached --check`

ShellCheck was not run locally because installing it was blocked by the unaccepted Xcode licence. A broader local release-test loop was attempted but the local session crashed, so it is not reported as passing.

## Notes for Reviewers

The polling timeout defaults to 300 seconds with a 15-second interval. Tests cover delayed indexing, perpetual 404 timeout, network failure, request counting, the default timeout bound, and removal of the fixed `sleep 30`.

This PR keeps the change focused on the polling path described in #420.


vladimirrott (changes requested):
Thanks for this, and for the honesty in the body about ShellCheck and the crashed local loop. Two blocking items below, both about the test rather than the fix. The polling helper itself is right, and I want to say what I verified before I say what is missing.

Head reviewed: `b4b15427`. `maintainer screen 463` returns DO NOT EXECUTE, so everything ran in rootless podman with no network against a detached worktree.

## What holds up

Your test passes as written:

```
$ out="$(podman run --rm --network=none -v /tmp/r21p463:/repo:ro -w /repo docker.io/library/bash:5 bash tests/release/crates-index-poll.test.sh 2>&1)"; rc=$?; echo "rc=$rc"; printf '%s\n' "$out" | grep -v level=warning | tail -2
rc=0
sysknife-daemon 0.15.0 is available on crates.io
crates index polling contract passed.
```

ShellCheck, the run you could not do locally, is clean over the whole sweep the `scripts-lint` job runs:

```
$ cd /tmp/r21p463 && out="$(find tests/e2e tests/release scripts assets/demo -type f -name '*.sh' -print0 | xargs -0 shellcheck --severity=warning 2>&1)"; rc=$?; echo "shellcheck rc=$rc"; printf '%s\n' "$out" | head -40
shellcheck rc=0
```

That is ShellCheck 0.10.0 here against 0.11.0 in CI, so treat it as strong rather than conclusive.

The `status=$?` in the `else` branch is the line I expected to find wrong and it is correct. Bash keeps the condition's status there, which I confirmed rather than assumed:

```
$ cat /tmp/semtest.sh
#!/usr/bin/env bash
set -e
f() { return "${1:-2}"; }
for v in 0 1 2; do
  if f "$v"; then
    echo "v=$v: branch=then"
  else
    status=$?
    echo "v=$v: branch=else status=$status"
  fi
done
$ bash /tmp/semtest.sh; echo "rc=$?"
v=0: branch=then
v=1: branch=else status=1
v=2: branch=else status=2
rc=0
```

So a 404 from the skip check reaches `((status != 1))` as 1 and falls through to publish, and a network failure reaches it as 2 and exits. That last part is a real improvement the PR body undersells: the old `curl --fail` collapsed "crates.io is unreachable" and "this version does not exist" into the same non-zero, so a blip during the skip check sent the job into a publish it had no basis for. Yours names the failure and stops.

## Blocking 1: the test reaches no CI job

`tests/release/*.test.sh` scripts run in CI by name, one step each in the `docs-and-hygiene` job. Yours is not in that list, and on your branch it is the only release test in that position:

```
$ cd /tmp/r21p463 && for f in tests/release/*.test.sh; do b=$(basename "$f"); if ! grep -rqF "$b" .github/workflows/; then echo "NOT IN WORKFLOWS: $b"; fi; done
NOT IN WORKFLOWS: crates-index-poll.test.sh
```

The same loop on `main` prints nothing, so this would be the first one. `scripts/ci-local.sh` discovers every release test by glob and would run it on a developer's machine, and `tests/release/ci-local.test.sh` asserts CI ⊆ local but not local ⊆ CI, so nothing goes red today. #377 is open and adds exactly that missing direction; whichever of the two lands second will turn this red, so it is worth fixing now either way.

Add a step next to the others in `.github/workflows/ci.yml`:

```yaml
      - name: Check the release waits for crates.io to index each crate
        run: bash tests/release/crates-index-poll.test.sh
```

## Blocking 2: the test stays green with the fix removed

This is the one I care about. Your test asserts the helper's behaviour and asserts that `sleep 30` is gone, and nothing asserts the workflow calls the helper. Delete the call and everything still passes:

```
$ cd /tmp/r21p463 && cp /tmp/rel463.orig .github/workflows/release.yml && sed -i '/wait_for_crate_version "\$crate" "\$VERSION"/d' .github/workflows/release.yml
$ out="$(podman run --rm --network=none -v /tmp/r21p463:/repo:ro -w /repo docker.io/library/bash:5 bash tests/release/crates-index-poll.test.sh 2>&1)"; rc=$?; echo "MUTATION-2 (workflow never calls the poller) rc=$rc"; printf '%s\n' "$out" | grep -v 'level=warning' | tail -4
MUTATION-2 (workflow never calls the poller) rc=0
sysknife-daemon 0.15.0 is available on crates.io
crates index polling contract passed.
```

That tree publishes six crates back to back with no wait of any kind, which is worse than the `sleep 30` you replaced, and the contract test calls it a pass. The `sleep 30` assertion is half an anchor: it catches the old mechanism coming back and misses the new one going away.

The restoring assertion is short. Something in the shape of:

```bash
grep -Fq 'wait_for_crate_version "$crate" "$VERSION"' "$workflow" || {
    printf 'FAIL: release workflow does not wait for the index after publishing\n' >&2
    exit 1
}
grep -Fq 'source scripts/crates-index-poll.sh' "$workflow" || {
    printf 'FAIL: release workflow does not load the polling helper\n' >&2
    exit 1
}
```

Please run the same mutation yourself once it is in and paste the red output, the way you pasted the green.

For completeness, the two mutations that do bite, so you know which parts are already anchored. Putting `sleep 30` back:

```
$ sed -i 's|              wait_for_crate_version "$crate" "$VERSION"|              sleep 30|' .github/workflows/release.yml
$ out="$(podman run --rm --network=none -v /tmp/r21p463:/repo:ro -w /repo docker.io/library/bash:5 bash tests/release/crates-index-poll.test.sh 2>&1)"; rc=$?; echo "MUTATION-1 rc=$rc"; printf '%s\n' "$out" | grep -v 'level=warning' | tail -4
MUTATION-1 rc=1
sysknife-daemon 0.15.0 is available on crates.io
FAIL: release workflow still contains fixed sleep 30
```

And neutering the timeout guard inside the helper:

```
$ sed -i 's|^        if ((elapsed >= timeout)); then|        if false; then|' scripts/crates-index-poll.sh
$ out="$(timeout 120 podman run --rm --network=none -v /tmp/r21p463:/repo:ro -w /repo docker.io/library/bash:5 bash tests/release/crates-index-poll.test.sh 2>&1)"; rc=$?; echo "MUTATION-3 (timeout guard removed) rc=$rc"
MUTATION-3 (timeout guard removed) rc=124
```

## Two optional notes

`rc=124` above is my `timeout 120` firing, not your test failing. With `CRATES_IO_SLEEP=true` the perpetual-404 case spins with nothing to stop it, so a broken timeout arithmetic hangs the job until GitHub's six-hour limit instead of going red in two seconds. Bounding the loop by iteration count as well as elapsed seconds, or wrapping the perpetual-404 case in `timeout`, turns that into a fast failure. Optional.

`tests/release/crates-index-poll.test.sh` has no newline at end of file. No gate objects (`tracked-eol.test.sh` checks CRLF and mixed endings only), so this is cosmetic.

## Your checks, and why they are sitting there

The board shows no checks because four runs are parked at `action_required`:

```
$ gh api repos/lacs-project/sysknife/actions/runs --jq '.workflow_runs[] | select(.head_sha|startswith("b4b1542")) | "\(.id)\t\(.name)\t\(.status)\t\(.conclusion)"'
35508302296	ci	completed	action_required
35508302271	e2e	completed	action_required
35508302270	CodeQL	completed	action_required
35508302269	secret-scan	completed	action_required
```

I approve first-time contributors' runs as a matter of course, and I am not approving these, because this PR edits `.github/workflows/release.yml`. Approving a fork run that touches workflow files is the one case where that click can hand a fork the repository's token, so a human reads the diff first. Nothing about that is a judgement on you or on the change; it is the rule for the file, not for the author. I will read it and release the runs.

## Scope

You kept this to the polling path and left the `--no-verify` half of #420 alone, which is the right call. That second bullet is a separate decision about how we publish, and it should not ride along on a fix that is already testable on its own.

Once the two blocking items are in, re-request review and I will run the same mutations against the new head.

One more thing. You are comfortable in release plumbing and you wrote a curl stub with a request counter rather than trusting a sleep, which is the instinct this repository is built around. If you run Ubuntu anywhere, SysKnife's read-only side is worth ten minutes: `sysknife plan "why is my disk filling up"` shows you the proposed actions without executing anything, and the approval step is where the whole design lives. Take it or leave it, and it has nothing to do with this PR landing.

vladimirrott (approved):
Both blocking items are closed, and you closed the harder one the way I hoped. The two `grep -Fq` lines against `release.yml` are what turn this from a description of the helper into a contract about the release. A test that only exercises `wait_for_crate_version` in isolation cannot notice the workflow dropping the call, and that was the whole finding.

Re-pinned at `c5cba691`. `main` moved under you (nine commits since your first push), so everything below ran on the three-way merge of `35f4970e` and your head, which resolves clean at `f9ecf196`.

## The mutation that used to stay green

Last time I deleted the poller call from the workflow and your test still printed `crates index polling contract passed`. Same deletion, this head:

```
$ cp /tmp/rel463.orig /tmp/r22mut/.github/workflows/release.yml && sed -i '/wait_for_crate_version "\$crate" "\$VERSION"/d' /tmp/r22mut/.github/workflows/release.yml && out="$(podman run --rm --network=none -v /tmp/r22mut:/repo:ro -w /repo localhost/sk-yamlgate:1 bash tests/release/crates-index-poll.test.sh 2>&1)"; rc=$?; echo "MUTATION-1 (workflow never calls the poller) rc=$rc"; printf '%s\n' "$out" | grep -v 'level=warning'
MUTATION-1 (workflow never calls the poller) rc=1
sysknife-daemon 0.15.0 is available on crates.io
FAIL: release workflow does not wait for the index after publishing
```

Four more, all on the merged tree:

| What I broke | rc | What the test said |
|---|---|---|
| deleted `source scripts/crates-index-poll.sh` from the workflow | 1 | `FAIL: release workflow does not load the polling helper` |
| restored `release.yml` to `main`'s version, so the whole production hunk is gone | 1 | `FAIL: release workflow does not load the polling helper` |
| helper returns 0 on HTTP 404 | 1 | `FAIL: expected 4 poll requests, got 1` |
| deleted `release.yml` outright | 1 | `grep: .github/workflows/release.yml: No such file or directory` then `FAIL: release workflow does not load the polling helper` |

The last one matters more than it looks. A guard that reads a file has to fail when the file is missing rather than inspect nothing and pass, and yours does, because `grep -Fq` returns 2 and your `|| { ... exit 1; }` does not care which non-zero it got.

## The CI step, and why it is doing real work now

#377 landed on `main` after your last push and added `scripts/check_test_reachability.sh`, which refuses any `tests/release/*.test.sh` that no gate invokes. Your `ci.yml` step is what keeps that green:

```
$ out="$(podman run --rm --network=none -v /tmp/r22m463:/repo:ro -w /repo localhost/sk-yamlgate:1 bash scripts/check_test_reachability.sh 2>&1)"; rc=$?; echo "UNMUTATED reachability rc=$rc"; printf '%s\n' "$out" | grep -v 'level=warning'
UNMUTATED reachability rc=0
test-reachability: every release and E2E test is invoked by a gate.
```

```
$ cp /tmp/ci463.orig /tmp/r22mut/.github/workflows/ci.yml && sed -i '/bash tests\/release\/crates-index-poll.test.sh/d;/Check the release waits for crates.io to index each crate/d' /tmp/r22mut/.github/workflows/ci.yml && out="$(podman run --rm --network=none -v /tmp/r22mut:/repo:ro -w /repo localhost/sk-yamlgate:1 bash scripts/check_test_reachability.sh 2>&1)"; rc=$?; echo "MUTATION-3 (new test reaches no CI job) rc=$rc"; printf '%s\n' "$out" | grep -v 'level=warning'
MUTATION-3 (new test reaches no CI job) rc=1
test-reachability: test is not invoked by a gate: tests/release/crates-index-poll.test.sh
```

You wrote the step before that gate existed on `main`, which is the ordering I would rather have than the other one.

## The rest of the gate set

ShellCheck 0.11.0, the same version CI installs, using the command from `e2e.yml` and the repo's own file list:

```
$ out="$(podman run --rm --network=none -v /tmp/r22m463:/repo:ro -v /tmp/sc463.sh:/sc.sh:ro -v /tmp/shellcheck-static:/usr/local/bin/shellcheck:ro localhost/sk-yamlgate:1 bash /sc.sh 2>&1)"; rc=$?; echo "SHELLCHECK rc=$rc"; printf '%s\n' "$out" | grep -v 'level=warning' | head -40
SHELLCHECK rc=0
files=210
```

You said in the PR body that you could not install ShellCheck locally. Nothing in your two new scripts trips it, so that gap cost you nothing.

`yamllint 1.38.0` on the workflows: rc=0, warnings only on three issue templates that were already there. `tests/release/ci-local.test.sh`: 5 tests, OK. `tracked-eol.test.sh`: `ok: tracked Git index contains no CRLF or mixed line endings`, which also picks up the missing trailing newline you fixed. That last one I could not run at all last time; the failure was my container mount, not your file, and I said the wrong thing by leaving it ambiguous.

## One optional note, unchanged, not holding anything up

Break the timeout check itself and the test hangs instead of going red:

```
$ sed -i 's|if ((elapsed >= timeout)); then|if false; then|' /tmp/r22mut3/scripts/crates-index-poll.sh
MUTATION-7 (timeout check disabled) rc=124
```

124 is my own `timeout 30` firing. On a runner that is a job spinning until GitHub's six-hour limit rather than a two-second failure. A `timeout` around the poll in the test, or a hard iteration cap in `wait_for_crate_version`, would close it. I am not asking for it here, and I will not hold the merge on it.

## Two things that are mine, not yours

Your four workflow runs at `c5cba691` are sitting at `action_required`:

```
$ gh api 'repos/lacs-project/sysknife/actions/runs?per_page=100&status=action_required' --jq '.workflow_runs[] | "\(.id)\t\(.name)\t\(.head_branch)\t\(.head_sha[0:8])\t\(.created_at)"'
35569989964	secret-scan	fix/crates-index-poll	c5cba691	2026-09-21T06:47:27Z
35569989966	CodeQL	fix/crates-index-poll	c5cba691	2026-09-21T06:47:27Z
35569989962	ci	fix/crates-index-poll	c5cba691	2026-09-21T06:47:27Z
35569990048	e2e	fix/crates-index-poll	c5cba691	2026-09-21T06:47:27Z
```

First-time contributors' runs are gated here, and this PR edits `.github/workflows/`, which is the case where clicking approve hands a fork the workflow token. I read the workflow diff by hand instead. The same rule means the merge button on this one is not mine either: a workflow change gets a second pair of eyes before it lands, whatever the board says. Neither of those is a judgement about your work and neither is a delay you caused.

Approving at `c5cba691`. The proof above is recorded so nobody has to re-run it, and I will write the CHANGELOG entry at merge; this one goes under Fixed, since the failure it prevents is a release that publishes `sysknife-daemon` before crates.io can see `sysknife-brain`.

## If you want another

#458 is the closest match to this PR: `tests/release/release-rehearsal.test.sh` screens three spellings of the publish command, so `$(cargo publish)` and a couple of other forms walk straight past it. Same file family, same shape of fix, and you have now read that workflow more carefully than anyone. #459 is smaller if you would rather have a quick one: the composite action under `.github/actions` sits outside every workflow check, including the pin verifier. I will hold #458 for you for a week; say the word either way and I will assign it.

And since you are running releases somewhere: SysKnife's read-only side (`sysknife ask "what services are listening"`, `sysknife audit verify`) needs nothing privileged and no daemon, so it is a five-minute thing to point at a box you already own. I would like to know what it gets wrong on your machines. Easy to ignore if it is not your kind of tool.

vladimirrott:
Correction to the last paragraph of that review, and the miss is mine.

@QinXi-ai opened #469 against #458 twelve minutes before I posted, and #471 against #459 five minutes before. Both issues are theirs, assigned and labelled now. I read the claim state earlier in the evening and wrote that closing paragraph from the earlier read instead of a fresh one, so I pointed you at two things somebody was already working on.

Nothing about your pull request changes. It is approved at `c5cba691`, the proof is recorded, and it is waiting on a human to read the workflow diff before it lands.

I owe you a replacement, and I would rather spend a day finding one nobody is holding than name a third in a hurry. I will come to you inside the week with something in the release-gate area and will not point anyone else at it first. If you would rather choose, [open issues with `help wanted` and without `claimed`](https://github.com/lacs-project/sysknife/issues?q=is%3Aopen+label%3A%22help+wanted%22+-label%3Aclaimed) are the free ones; name any of them and I will assign it to you.

